PR #180, opened 2015-01-12
make trends the default data view

jebeck:
What it says on the tin. @jh-bate review?

For Trello card https://trello.com/c/kt8691fr

### DIFF
--- a/app/pages/patientdata/patientdata.js
+++ b/app/pages/patientdata/patientdata.js
@@ -76,7 +76,7 @@ var PatientData = React.createClass({
         },
         timePrefs: this.props.timePrefs
       },
-      chartType: 'daily',
+      chartType: 'modal',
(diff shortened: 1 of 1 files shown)


app/pages/patientdata/patientdata.js
```diff
@@ -76,7 +76,7 @@ var PatientData = React.createClass({
         },
         timePrefs: this.props.timePrefs
       },
-      chartType: 'daily',
+      chartType: 'modal',
```
jh-bate: is it worth being pedantic and calling this something like defaultChart: 'modal' ??

jh-bate: ha ha - ok just viewed the whole file and spotted its set in the getInitialState

jh-bate:
LGMT

PR #182, opened 2015-01-12
Abbreviates dates in the patient data charts and brings arrows in closer.

ursooperduper:
This PR abbreviates the dates shown in daily, weekly, and trend views of patient data. 
@jebeck @skrugman 

To tighten the position of the arrows, I removed a large block element that surrounded the dates in each view. When you test this, you'll notice that this causes the arrows to bounce slightly as the width of the date string changes. In my testing, I found that you can still click the arrows repeatedly to move forward or backward in time without losing focus. 

If the slight jump/bounce of the arrows is undesirable, I can create CSS rules for each view that will reintroduce the block around the date, just a bit tighter than before when there was one block for all views. 

Let me know how you'd like me to proceed!

Resolves: https://trello.com/c/RELPBpRy

I agree to the terms of Tidepool Project’s Volunteer/Contributor License Agreement v1.0
as it exists at http://tidepool-org.github.io/TidepoolVCLA.pdf on Monday, January 12, 2015.
cc/ @kentquirk

### DIFF
--- a/app/components/chart/daily.js
+++ b/app/components/chart/daily.js
@@ -117,7 +117,7 @@ var Daily = React.createClass({
     else {
       timezone = timePrefs.timezoneName || 'UTC';
     }
-    return sundial.formatInTimezone(datetime, timezone, 'ddd, MMM Do');
+    return sundial.formatInTimezone(datetime, timezone, 'ddd, MMM D, YYYY');
   },
   // handlers
   handleClickModal: function(e) {
--- a/app/components/chart/daily.js
+++ b/app/components/chart/daily.js
@@ -1,16 +1,16 @@
 /** @jsx React.DOM */
-/* 
+/*
  * == BSD2 LICENSE ==
  * Copyright (c) 2014, Tidepool Project
- * 
+ *
  * This program is free software; you can redistribute it and/or modify it under
  * the terms of the associated License, which is identical to the BSD 2-Clause
  * License as published by the Open Source Initiative at opensource.org.
- * 
+ *
  * This program is distributed in the hope that it will be useful, but WITHOUT
  * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
  * FOR A PARTICULAR PURPOSE. See the License for more details.
- * 
+ *
  * You should have received a copy of the License along with this program; if
  * not, you can obtain one from Tidepool Project at tidepool.org.
  * == BSD2 LICENSE ==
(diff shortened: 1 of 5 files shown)


app/components/chart/daily.js
```diff
@@ -117,7 +117,7 @@ var Daily = React.createClass({
     else {
       timezone = timePrefs.timezoneName || 'UTC';
     }
-    return sundial.formatInTimezone(datetime, timezone, 'dddd, MMMM Do');
+    return sundial.formatInTimezone(datetime, timezone, 'ddd, MMM D, YYYY');
```
jebeck: Do we want to keep ordinal markers after the dates? e.g., 14th instead of 14?

This is a question as much for @skrugman and @brandonarbiter 

If yes, everywhere or just daily view?

app/pages/patientdata/patientdata.less
```diff
@@ -33,16 +33,21 @@
 
 .patient-data-subnav-text {
   display: inline-block;
-  padding: @patient-data-subnav-vertical-padding @patient-data-subnav-horizontal-padding;
+  padding: @patient-data-subnav-vertical-padding 0 @patient-data-subnav-horizontal-padding 0;
```
jebeck: There's a funny thing going one when the app is in the loading data state (title bar gets really vertically skinny), and I think this change might be the root?
screenshot 2015-01-13 10 19 44

To get a good long look at it you can use "mock" mode and and the URL http://localhost:3000/?api.patientdata.get.delay=5000#/patients/11/data (auth.skip in mock mode seems to be broken though...)

skrugman: Lets get rid of the th after the #

*Sara Krugman *
Lead designer

*Tidepool *An open source, not-for-profit effort to build an open data
platform and better applications that reduce the burden of Type 1 Diabetes
and accelerate the commercialization of closed-loop systems.

*Phone :* 617 549 1483
*Email : *sara@tidepool.org
*Web : *Tidepool.org <http://tidepool.org/>

ursooperduper: I definitely introduced that one! It's fixed in my latest commit. There was inconsistent padding on linked header items versus plain text. I've correct that and cleaned up a bit!

app/pages/patientdata/patientdata.less
```diff
@@ -27,27 +27,33 @@
   color: #fff;
   background-color: @gray-dark;
 
-  padding-left: 0;
-  padding-right: 0;
+  // padding-left: 0;
+  // padding-right: 0;
```
jebeck: Can you delete commented out lines?

jebeck:
Just waiting on deletion of commented-out code and another update merge from master; otherwise looks good.

ursooperduper:
@jebeck 
Removed all the commented out code. Sorry about that! The branch is merged and the build passes.

jebeck:
Awesome, thanks!

PR #183, opened 2015-01-12
Removes the gray bar and makes the refresh link a button.

ursooperduper:
@jebeck @skrugman 

The gray bar has been removed from footer in the patient data chart. 

The Refresh link is now a button (using the 'teritary' btn style from the core styles). Let me know if you'd prefer something different. The primary button style seemed too distracting and the secondary was too light to be noticeable. 

The links on the right side of the footer are now dark gray and an underline is displayed on hover. 

Resolves: https://trello.com/c/EJL3rmil

### DIFF
--- a/app/components/chart/footer.js
+++ b/app/components/chart/footer.js
@@ -1,16 +1,16 @@
 /** @jsx React.DOM */
-/* 
+/*
  * == BSD2 LICENSE ==
  * Copyright (c) 2014, Tidepool Project
- * 
+ *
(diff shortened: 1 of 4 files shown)


ursooperduper:
@jebeck @skrugman 

I updated the button style to match Sara's design. In doing so, I added it as a core button style, just so we don't button styling littered all over our .less files. 

There was no design for hover, so I just lightened the button when it's hovered over.

jebeck:
This looks good. Merging.

PR #184, opened 2015-01-12
set lines off, group, range & avg. on by default for trends

jebeck:
Set the desired default options for trends view.

@jh-bate another quick look-see?


jh-bate:
LGTM

PR #187, opened 2015-01-14
ignore error on logout and destroy the session

jh-bate:
this relates an error seen when a session times out, then the user clicks "log out", the app hangs. It should re-direct to the login page.

@jebeck if you could have a quick look?

### DIFF
--- a/app/core/api.js
+++ b/app/core/api.js
@@ -114,7 +114,7 @@ api.user.logout = function(cb) {
   if (!api.user.isAuthenticated()) {
     api.log('not authenticated but still destroySession');
     tidepool.destroySession();
-    return;
+    return cb();
   }
 
   tidepool.logout(function(err) {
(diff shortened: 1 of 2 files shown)


app/core/api.js
```diff
@@ -112,14 +112,17 @@ api.user.logout = function(cb) {
   api.log('POST /user/logout');
 
   if (!api.user.isAuthenticated()) {
+    api.log('not authenticated but still destroySession');
+    tidepool.destroySession();
     return;
```
kentquirk: You still need to call cb here -- you can't just return.

jebeck:
Other than @kentquirk's comment, looks good to me.

jebeck:
Oh but also plz update from master @jh-bate

app/core/api.js
jh-bate: hmm i don't think it was even there before? just adding

PR #188, opened 2015-01-14
Light gray outline added to the navbar dropdown menu.

ursooperduper:
@skrugman @jebeck 

This PR removes the shadow from the dropdown menu and adds a light gray border around it. The change took a bit more time than I thought since you have to do some CSS magic to get an outline around that triangle in the menu. I've tested my change in Chrome, Safari, and Firefox to make sure it's working cross-browser and it looks good.

Resolves: https://trello.com/c/Y5eXAWyQ

### DIFF
--- a/app/components/navbar/navbar.less
+++ b/app/components/navbar/navbar.less
@@ -321,13 +321,13 @@
 }
 
 .Navbar-menuDropdown {
+  border: 1px solid @gray-light;
   cursor: default;
   position: absolute;
   background: white;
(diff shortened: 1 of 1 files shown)


jebeck:
This looks good to me, merging.

Since you mentioned other browsers @ursooperduper, I had a look in Firefox and Safari to confirm, and I think I found a bug with the icon alignment in Firefox, but I filed that separately on Trello.

PR #189, opened 2015-01-14
Link cleanup for the patientcard component

ursooperduper:
@skrugman @jebeck 

This PR updates the patient card component in two places: 
1) On the patients page, the bold hover effect has been removed from the View|Share|Upload links to make navigation less jumpy. 
2) On the patient data page, the patient name and icon (in the patient card component in the navbar) now has a light blue green hover effect with an underline for the patient name for consistency with other text links in the site. 

Per @skrugman, in the future, we'll look for better affordances for indicating hover than the outline, but this will be good for accessibility for the time being. 

Resolves: https://trello.com/c/W6OCIaCY

### DIFF
--- a/app/components/patientcard/patientcard.less
+++ b/app/components/patientcard/patientcard.less
@@ -76,6 +76,7 @@
     margin-top: 5px;
   }
 }
+
 .patientcard-fullname {
   padding-top: 5px;
   padding-left: 5px;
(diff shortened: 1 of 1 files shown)


jebeck:
Looks good, merging this as well.

(P.S. I know I'm a broken record on this one, but in case there's consensus that this would be a good time to work on this: the patientcard is one of the key places that does not meet our goal of WCAG AA accessibility requirements for text contrast.)

PR #192, opened 2015-01-15
Only show "Remove People" link when there's an account displayed that the user does not own.

ursooperduper:
@jh-bate 

This PR fixes functionality that determines whether the 'Remove people' link should be shown. The original code (editControls, editablePersonExists) was incorrectly always showing the link because it was looking for the existence of admin perms, which are always there - because the logged in user has admins perms for their own account. 

To get this working, I added an .isRemoveable function to core/personutils.js that, when passed a person, checks to ensure that the perms aren't empty and their they're not root and not admin. If that is true, then it's a person the user has view/upload/etc rights for. When that is the case, the remove button should be shown onscreen. 

Resolves: PR: https://trello.com/c/1YFnGntX

### DIFF
--- a/app/components/peoplelist/peoplelist.js
+++ b/app/components/peoplelist/peoplelist.js
@@ -71,7 +71,7 @@ var PeopleList = React.createClass({
       'people-list-single': this.props.people.length === 1
     });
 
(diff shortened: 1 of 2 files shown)


jh-bate:
LGTM

PR #164, opened 2014-11-09
Skip dashboard and redirect to data in certain situations

nicolashery:
This is a feature extracted from https://github.com/tidepool-org/blip/pull/162, and moved to its own PR so we can work on it a bit more...

PR #193, opened 2015-01-16
Determine whether to render the Patients or skip and go directly to Patient Data

ursooperduper:
This change picks up on a stale PR (https://github.com/tidepool-org/blip/pull/164) which had some logic problems. 

The approach is pretty simple, whenever we call redirectToDefaultRoute, we call showPatients, which now sets a state called canSkipPatients as true.

In the renderPatients function, when we see that canSkipPatients flag, we do some checks to make sure data has loaded, then figure out how many people the user can view data for - then direct them to either the patients page or patient data page. After stepping carefully through all the cases (if you're a patient, caregiver, etc.) and the permissions you have for various other users, the logic I arrived at it pretty simple: 

Providing the user has no pending invites, if there is only patient (including the user themselves) the user can see data for, go directly to the Patient Data page, otherwise, go to the Patients page.

I've walked through this with @kentquirk, but @jebeck or @jh-bate, can you review? Thank you!

Resolves: https://trello.com/c/bmAQYqbu

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -427,10 +427,11 @@ var AppComponent = React.createClass({
   },
 
   redirectToDefaultRoute: function() {
-    this.showPatients();
+    this.showPatients(true);
   },
 
-  showPatients: function() {
+  showPatients: function(canSkipPatients) {
+    this.setState({canSkipPatients: canSkipPatients});
     this.renderPage = this.renderPatients;
     this.setState({page: 'patients'});
     this.fetchInvites();
@@ -439,8 +440,9 @@ var AppComponent = React.createClass({
   },
 
   renderPatients: function() {
+    var patients;
     /* jshint ignore:start */
-    return <Patients
+    patients = <Patients
         user={this.state.user}
         fetchingUser={this.state.fetchingUser}
         patients={this.state.patients}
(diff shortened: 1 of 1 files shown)


app/app.js
```diff
@@ -427,10 +427,11 @@ var AppComponent = React.createClass({
   },
 
   redirectToDefaultRoute: function() {
-    this.showPatients();
+    this.showPatients(true);
   },
 
-  showPatients: function() {
+  showPatients: function(canSkipPatients) {
```
jh-bate: hmm just sounds weird  - showPatients ( canSkipPatients )

jh-bate:
I get what it is trying to do but I don't think that code reads quite right yet. I think its hard as everything has the word patient in it :)

app/app.js
```diff
@@ -456,6 +458,49 @@ var AppComponent = React.createClass({
         onDismissInvitation={this.handleDismissInvitation}
         onRemovePatient={this.handleRemovePatient}/>;
     /* jshint ignore:end */
+
+    // Determine whether to skip the Patients page & go directly to Patient data.
+    // If there is only one patient you can see data for, go to the patient's data.
+    // Otherwise, display the Patients page.
+    if (this.state.canSkipPatients) {
```
jh-bate: so is this canSkipPatientsPage ?? or should it be called  showPatientData??

ursooperduper:
@jh-bate 

Good call. The naming was a bit awkward. I've changed it to showPatientData as you suggested and think the readability has improved. Let me know what you think.

jh-bate:
Thanks - feels better.

PR #194, opened 2015-01-19
Jhbate/fix logout

jh-bate:
@jebeck if you could quickly look at this.

I have now just called the logout and then continue with the success process as we aren't waiting for feedback from the server anyway.


jebeck:
OK, merging. I'll attempt my fancier fix later if I have time, and we'll see if it works.

PR #195, opened 2015-01-19
Jhbate/signup flow

jh-bate:
@jebeck I have some questions / things I am looking for feedback on

Addition of SignupVerification page that is used to tell the user that the authentication email has been sent or that they haven't yet authenticated if they try and login before accepting the signup confirmation.

Questions:

- I used the footer email link to show the 'help me with my signup' link see https://github.com/tidepool-org/blip/compare/jhbate/signup-flow?expand=1#diff-2de9eda1655d499e934e7465e87ca0bdL334 maybe I should have kept it in the new page? Thoughts?

- I call the starting of the signup flow in the existing signup api call. If the signup email being sent fails I don't fail the whole process - thought being I shouldn't as they can then email later and ask for help (or use a resend link) and I would rather do that then kill their signup. Thoughts /  comments? https://github.com/tidepool-org/blip/compare/jhbate/signup-flow?expand=1#diff-a9e125b91e9e2c58c685083cbfa74a49R97

- I am getting a jshint error that doesn't quite make sense and though you might know the answer to

```
.... blip/app/app.js
  line 432  col 7  Missing 'new' prefix when invoking a constructor.
```  

- NOTE: I am referencing platform client from a branch as I have yet deployed all the dependencies, I will fix this up on the merge


app/pages/signupverification/signupverification.js
```diff
@@ -0,0 +1,63 @@
...
+          <div className="SignupVerification-instructions">
+            <p>{'We just sent you an email. To verify we have the right email address, please click the link in the email to activate your account.'}</p>
+          </div>
+        </div>
+      );
+    }
+    else {
+      content = (
+        <div className="SignupVerification-intro">
+          <div className="SignupVerification-title">{'Hey your not verified yet.'}</div>
```
jebeck: Copy-editing: Hey, you're not verified yet.

app/pages/signupverification/signupverification.js
```diff
@@ -0,0 +1,63 @@
...
+            <div className="SignupVerification-instructions">
+              <p>{'Check your email and follow the link there. (We need to confirm that you are really you).'}</p>
```
jebeck: Copy-editing: final period inside the parens: (We need to confirm that you are really you.)

jebeck:
Copy-editing questions (possibly more for @brandonarbiter and @HowardLook):
- Do we use `email` (as @jh-bate has it) or `e-mail`? (My inner copy-editor prefers the latter.)
- Ditto `signup` vs. `sign-up`
- Is "Help, I cannot complete signup" the link text we want for sign-up confirmation help?

@jh-bate as I said before, I'd prefer a separate help link and keep the feedback link the same on all pages. The help link should also be the same color as the other text on the sign-up page (for text contrast accessibility - color is #727375), and maybe give it some generous margin below to keep it distinct from the feedback link? (Actually it'd be super if the feedback link was basically persistent height across all pages.)

jebeck:
@jh-bate did you have a plan for devs getting around the lack of conf e-mail sent when running locally? Without one, we'll never be able to get into new local accounts, right?

jebeck:
I think your take on not failing the whole sign-up process makes sense, but my question is whether the resend link is not included in the initial 1.0 version of this? Did we agree to manually respond to support e-mails? The problem I have with that is how the e-mail link doesn't work for people who have gmail as there default client. Don't know how many non-nerds know how to set that up, but it strikes me as a user experience problem, at least potentially.

CC @brandonarbiter @HowardLook

app/app.js
```diff
@@ -50,6 +50,7 @@ var PatientNew = require('./pages/patientnew');
 var PatientData = require('./pages/patientdata');
 var RequestPasswordReset = require('./pages/passwordreset/request');
 var ConfirmPasswordReset = require('./pages/passwordreset/confirm');
+var SignupVerification = require('./pages/signupverification');
```
jebeck: Part of me doesn't like the name `SignupVerification` and all the variations thereof. I'd prefer `EmailVerification` (since that's actually what we're verifying). But, my feelings aren't super strong on this. Change is part of you agrees, or don't, if you don't :)

jh-bate:
thanks @jebeck will do those changes now.

- dev tool for new account - yeah I can sort something for a local host environment

- the resend link - I have already coded this on the api so it would probably be best just to use it. Works in the same way as a pw-reset and it may have been my bad that I didn't wire it up anyway.

app/pages/login/login.js
```diff
@@ -189,6 +190,10 @@ var Login = React.createClass({
     submit(formValues, function(err) {
       if (err) {
         var message = 'An error occured while logging in.';
+        if(err.status === 403){
+          self.props.onSubmitNotAuthorized();
+          return;
+        }
```
jebeck: This whole bit is a bit confusing to read. Since `onSubmitNotAuthorized` redirects to the sign-up verification page, the error message here will never get rendered. (a) Is that the intended behavior? (b) If yes, can you at least move the line defining the message below the `if` block where `onSubmitNotAuthorized` is called?

jebeck:
OK, I'm done commenting now ;) Let me know if you have questions @jh-bate

jh-bate:
nah thats good - thanks a lot all makes sense to me

app/app.js
jh-bate: yeah agreed - i started with email then changed to signup but it does read better and match the intent so I will change this.

app/app.js
```diff
@@ -368,30 +380,58 @@ var AppComponent = React.createClass({
 
   showLogin: function() {
     this.renderPage = this.renderLogin;
+    //always check
+    this.finialiseSignup();
```
jebeck: Can you change these to American spelling - `finalizeSignup`? It's what we use elsewhere (e.g., 'authorize')

jebeck:
Gave it a quick look-over again @jh-bate, but didn't test locally or anything. LGTM.

jh-bate:
ta - will just update spelling

PR #197, opened 2015-01-20
Removes 'create later' text from welcome flow.

ursooperduper:
@jebeck 

This changes just removes a line of text from the welcome flow that didn't make sense anymore given recent layout changes.

Resolves: https://trello.com/c/vLKsJmWe


jebeck:
Merging.

PR #198, opened 2015-01-21
Fixes webpack configuration for favicon.ico

ursooperduper:
@kentquirk @jebeck 

Updates the webpack config to use the file loader which allows you to specify a name for a file in a build so that it's not hashed like most other assets.  

Resolves: https://trello.com/c/iQwJAoxZ


kentquirk:
I've built this and run it locally and it works for me on my local machine.

PR #199, opened 2015-01-22
Fixes long name layout issues.

ursooperduper:
@jebeck 

By making the navbar icons larger in an earlier change the other week, the spacing allotted for username length was reduced by several pixels causing wrapping of the menu section. 

To fix this, I removed extra padding created by the ul element automatically, which reduced the amount of space left for layout in the div. I also reduced the allotted space for a username by 10px. I've tested this change with extremely long and short names in Safari, Chrome, and Firefox to ensure our previous changes to align arrows and increase sizes were not effected. Everything checks out. 

Resolves:  https://trello.com/c/y9MIvhtH


jebeck:
Looks good, thanks! Merging.

PR #200, opened 2015-01-22
Rename Notification; it seems React now has one

kentquirk:
Sarah's jshint mysteriously broke -- it seems React has now added an object called Notification, which conflicts with ours. So I've renamed it TidepoolNotification.

I didn't see this problem until I deleted my node_modules and reinstalled them. Then jshint started failing.


kentquirk:
@jebeck @jh-bate I'm going to merge this as it has no semantic effect and it's blocking Sarah.

PR #201, opened 2015-01-25
Sk/trends iterations

ursooperduper:
@jebeck 

This PR makes a number of updates to the patient data trends view. 

* Add checkboxes to the all week/all weekend selectors. 
* Adds checkboxes (instead of text links) to range and avg, group, and line selectors. 
* Add a checkbox (instead of text link) to the values selector on the weekly view of patient data. 
* Moves the "days in view" text below week selectors. 
* Uses button (instead of text links) for the 1 week, 2 week, 4 week selectors. 

The changes match the latest design revisions posted on Pixelapse, with one slight change. I put the checkboxes for the weekday/weekend selectors to the left of the data they affect (instead of having the weekday selector to the left and the weekend selector to the right). I think this approach is clearer. Let me know what you think. A screenshot of the changes is included below: 

trends-iteration

This resolves the tasks from the card I have a good handle on from: https://trello.com/c/lOa72Cqa

Note: There are two outstanding tasks for which I need more clarification before I could approach: 
1. Tooltip for "Range and Average" (styling leveraging range box). <-- I'm not sure what this task means. 
2. Alternate time navigation / data paging. (I don't know what the design is for the alternate paging). 

I felt enough changes had been made to the charts that a pull request for the completed tasks was worth it. 

Tested on: Chrome, Firefox, Safari

### DIFF
--- a/app/components/chart/footer.js
+++ b/app/components/chart/footer.js
@@ -46,8 +46,7 @@ var TidelineFooter = React.createClass({
       <label htmlFor="valuesCheckbox">
         <input type="checkbox" name="valuesCheckbox" id="valuesCheckbox"
           checked={this.props.showingValues}
-          onChange={this.props.onClickValues}
-          onKeyDown={this.handleValuesKeyDown} /> Values
+          onChange={this.props.onClickValues} /> Values
       </label>
       );
     /* jshint ignore:end */
@@ -58,22 +57,19 @@ var TidelineFooter = React.createClass({
         <label htmlFor="linesCheckbox">
           <input type="checkbox" name="linesCheckbox" id="linesCheckbox"
             checked={this.props.showingLines}
-            onChange={this.props.onClickLines}
-            onKeyDown={this.handleLinesKeyDown} /> Lines
+            onChange={this.props.onClickLines} /> Lines
         </label>
 
         <label htmlFor="groupCheckbox">
           <input type="checkbox" name="groupCheckbox" id="groupCheckbox"
             checked={this.props.grouped}
-            onChange={this.props.onClickGroup}
-            onKeyDown={this.handleGroupKeyDown} /> Group
+            onChange={this.props.onClickGroup} /> Group
         </label>
 
         <label htmlFor="overlayCheckbox">
           <input type="checkbox" name="overlayCheckbox" id="overlayCheckbox"
             checked={this.props.boxOverlay}
-            onChange={this.props.onClickBoxOverlay}
-            onKeyDown={this.handleOverlayKeyDown} /> Range &amp; Average
+            onChange={this.props.onClickBoxOverlay} /> Range &amp; Average
         </label>
       </div>
       );
@@ -100,20 +96,6 @@ var TidelineFooter = React.createClass({
       </div>
       );
     /* jshint ignore:end */
-  },
-
-  // The following handlers let the user select a checkbox by hitting enter
-  handleValuesKeyDown: function() {
-    return event.keyCode !== 13 || this.props.onClickValues();
-  },
-  handleLinesKeyDown: function() {
-    return event.keyCode !== 13 || this.props.onClickLines();
-  },
-  handleGroupKeyDown: function() {
-    return event.keyCode !== 13 || this.props.onClickGroup();
-  },
-  handleOverlayKeyDown: function() {
-    return event.keyCode !== 13 || this.props.onClickBoxOverlay();
   }
 });
 
--- a/app/components/chart/footer.js
+++ b/app/components/chart/footer.js
@@ -43,21 +43,23 @@ var TidelineFooter = React.createClass({
 
     /* jshint ignore:start */
     var showValues = (
-      <label htmlFor="valuesCheckbox">
-        <input type="checkbox" name="valuesCheckbox" id="valuesCheckbox"
-          checked={this.props.showingValues}
-          onChange={this.props.onClickValues} /> Values
-      </label>
+      <div className="footer-right-options">
+        <label htmlFor="valuesCheckbox">
+          <input type="checkbox" name="valuesCheckbox" id="valuesCheckbox"
+            checked={this.props.showingValues}
+            onChange={this.props.onClickValues} /> Values
+        </label>
+      </div>
       );
     /* jshint ignore:end */
 
     /* jshint ignore:start */
     var modalOpts = (
-      <div>
-        <label htmlFor="linesCheckbox">
-          <input type="checkbox" name="linesCheckbox" id="linesCheckbox"
-            checked={this.props.showingLines}
-            onChange={this.props.onClickLines} /> Lines
+      <div className="footer-right-options">
+        <label htmlFor="overlayCheckbox">
+          <input type="checkbox" name="overlayCheckbox" id="overlayCheckbox"
+            checked={this.props.boxOverlay}
+            onChange={this.props.onClickBoxOverlay} /> Range &amp; Average
         </label>
 
         <label htmlFor="groupCheckbox">
@@ -66,10 +68,10 @@ var TidelineFooter = React.createClass({
             onChange={this.props.onClickGroup} /> Group
         </label>
 
-        <label htmlFor="overlayCheckbox">
-          <input type="checkbox" name="overlayCheckbox" id="overlayCheckbox"
-            checked={this.props.boxOverlay}
-            onChange={this.props.onClickBoxOverlay} /> Range &amp; Average
+        <label htmlFor="linesCheckbox">
+          <input type="checkbox" name="linesCheckbox" id="linesCheckbox"
+            checked={this.props.showingLines}
+            onChange={this.props.onClickLines} /> Lines
         </label>
       </div>
       );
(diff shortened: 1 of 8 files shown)


app/components/chart/footer.js
```diff
@@ -41,82 +41,40 @@ var TidelineFooter = React.createClass({
...
-    var overlayLinkText = getOverlayLinkText(this.props);
-
     /* jshint ignore:start */
     var showValues = (
-      <a href="" onClick={this.props.onClickValues}>{valuesLinkText}</a>
+      <label htmlFor="valuesCheckbox" ref="label">
+        <input type="checkbox" name="valuesCheckbox" id="valuesCheckbox"
+          checked={this.props.showingValues}
+          onChange={this.props.onClickValues}
+          ref="control" /> Values
```
jebeck: The 'label' and 'control' refs aren't being used that I can see, and React refs have to be unique, I believe (and the same ones appear on all the label/input pairs). Maybe just omit them?

app/components/chart/modalsubnav.js
```diff
@@ -109,20 +114,20 @@ var ModalSubNav = React.createClass({
...
   renderDomainLink: function(domain) {
     var domainLinkClass = cx({
-      'active': domain === this.props.activeDomain,
-      'modalDomain': true
+      'btn btn-chart-control' : true,
+      'active': domain === this.props.activeDomain
     });
     /* jshint ignore:start */
     return (
-      <a href="" className={domainLinkClass} key={domain} onClick={this.props.domainClickHandlers[domain]}>{domain}</a>
```
jebeck: Can you put the `key={domain}` back? Without it you get a React warning in the console (you may need React dev tools installed to see these): "Each child in an array should have a unique "key" prop."

app/components/chart/modalsubnav.js
```diff
@@ -134,7 +139,7 @@ var ModalSubNav = React.createClass({
     }
     /* jshint ignore:start */
     return (
-      <div className="dayGroupsContainer">
+      <div className="daysGroupContainer">
```
jebeck: Heehee, good catch on my naming inconsistency ;)

jebeck:
Some comments inline as well. And @skrugman should have a look at the styling.

Can we try to add enter handling for the checkboxes? If you tab through the page, you can do everything link-based by pressing enter, but it doesn't work for the checkboxes.

I also wonder if there's any way to get a pointer cursor on the checkboxes themselves (app-wide)? I did a quick Google, and it seems this is a point of controversy anyway (one camp says pointer should be reserved for links only; other camp says everything clickable should have pointer, which is generally the approach we seem to be taking).

CC @brandonarbiter 

Thanks!

app/components/chart/footer.js
ursooperduper: Good call. They do seem to be in conflict with React convention. I'll remove them.

app/components/chart/modalsubnav.js
ursooperduper: Will do! Sorry about that.

app/components/chart/modalsubnav.js
ursooperduper: ;)

ursooperduper:
Regarding using pointer cursors for checkbox elements, while the W3C test suite even specifies that pointers are used to indicate text links, I stand on the "make actionable elements stand out" camp. I've added a style to always make checkbox affordances use the pointer cursor.

skrugman:
@ursooperduper and @jebeck a few notes for styling here...

- Align the checkbox's next to the day pickers to the bottom of the day squares. 
- And is it possible to style them? The blue checked state is totally out of our color rules and palette (but i also understand why its like that) just wondering if we can change it (hopefully)?
- Give more space in-between each checkbox/label (i.e. move the Range and Average over 10-15px to the left, and then also Group over the same amount. (might be easier to do in browser together)
- One week/Two week/Four week/ box height should be the same height as the days of the week boxes
- The month labels on the scroll bar got days added to them and I dont think they should be there, i.e. i think they should just say months @jebeck ?

jebeck:
Re: the month labels on the scroll bar @skrugman that only happens where there isn't a large amount of data. This was a change @brandonarbiter and I made together a while back. You can end up with only a single month label sometimes (and it won't be in the center) otherwise.

skrugman:
Can we make a rule that only* if there is less than one month then we show
dates?

*Sara Krugman *
Lead designer

*Tidepool *An open source, not-for-profit effort to build an open data
platform and better applications that reduce the burden of Type 1 Diabetes
and accelerate the commercialization of closed-loop systems.

*Phone :* 617 549 1483
*Email : *sara@tidepool.org
*Web : *Tidepool.org <http://tidepool.org/>

jebeck:
@skrugman and I resolved the question of the scrollbar date tick labels; no change necessary now.

ursooperduper:
@skrugman I made some of the style changes you requested: 
* Ensuring the days of the week button boxes are the same height as the week button boxes. 
* Put more space between each label/checkbox group in the footer of the chart. 
* Ensure the days of the week group checkboxes line up vertically to the bottom of the day button boxes. 

The only change I didn't make was to change the styling of the checkboxes. By default, CSS support for checkbox styling is pretty limited. But there are some tricks you can do where you completely hide the default checkboxes and replace them with CSS drawn elements or even graphics. 

If you can give me a design - even just for color changes, I can see what we can do with this. We should do some good accessibility testing to make sure that the replacement checkboxes still work as well. We should also make this styling change globally for any checkbox element in Blip for consistency. 

So if you want to go ahead with styling those checkboxes, let me know what colors/design you'd like to use and I'll get to work.

jebeck:
My (unsolicited) opinion on the checkboxes issue is that it's better to stick with the native browser checkboxes. The hiding-and-replacing strategy forces us into the long-term technical burden of keeping them accessible, whereas using the native ones means we automatically benefit from any bugfixes or improvements the browser vendors make regarding their accessibility. My 2¢

And re: the enter-handling I asked for on the checkboxes. No need! I looked this up (I thought it was odd there wasn't native support for keyboard-based toggling of them), and it turns out of course there *is* native support - it's Spacebar, not Enter. I learned something new :)

CC @ursooperduper @skrugman

ursooperduper:
@skrugman I tend to agree with @jebeck. While it's possible to change checkbox styling with these CSS tricks, it will add technical debt that we'll have to deal with in the future. Toolkits like Bootstrap are great when more advanced styling is desired, simple because they'd invested the time in testing and it's seen a lot of use. But from an accessibility standpoint, this one is probably best left as is for now.

@jebeck For the enter-handling, I thought you wanted that in addition to the space  bar (which works as you discovered). I should have pointed that out earlier to save you the hunt. :) I'll remove the enter handling code I added since it's not abstracted to the point where it would just work for any new checkbox that's added. Right now, it would require custom handlers wherever you wanted enter to work.

skrugman:
@jebeck @ursooperduper for checkbox's - changing color to a dark grey or any grey from the bright blue would make a huge difference. The info design of Blip uses color only to indicate meaning and this bright blue messes that up. I understand that styling system standards is hard/adds code to maintain so perhaps there is another solution. There are standard box's like that in other places (gmail settings, and kick starter) that are grey that would be better design wise - is that a good solution from your end? 
screen shot 2015-01-30 at 9 00 52 am
screen shot 2015-01-30 at 9 02 15 am

jebeck:
@ursooperduper sorry about the enter handling thing, didn't notice you'd already done it! That was a brain fart on my part not to look it up first...

ursooperduper:
@skrugman @jebeck If you take a look at the styling of checkboxes in different browsers, you'll see that they appearance can vary. Sites like gmail rely on user-agent styles to render form controls. When you do this, you ensure that a user will recognize a checkbox on Google, the same way they will on Kickstarter or in any other app. If I call up Gmail or Kickstarter on my machine, I see the blue checkboxes, not the ones in the image in the above comment. The computers and OS we use and the browsers we use all affect how these form controls are rendered. So unfortunately, we don't have much choice other than hiding the default checkbox and redrawing it to change this color. 

That said, I can implement a custom checkbox using grey colors, but it requires more CSS, more testing, and can be an impediment to accessibility. Also, on older browsers, this styling won't work, so we'll need to do additional testing to make sure the styles degrade gracefully.

I'll add a new commit with some updated styling once I've had a chance to do some thorough testing.

jebeck:
Do you want to create a new branch and PR for that @ursooperduper? We can probably get this in soon without checkbox styling (although I want to make sure the finishing touches for https://trello.com/c/Pub5ZwkN get onto master before this PR, so that we don't have to hold up a deploy to prod based on testing this PR).

ursooperduper:
@jebeck That makes sense. I'll do the checkbox styling in a fresh branch and shiny new PR.

jebeck:
@ursooperduper was just doing a final check on this, and now I've noticed that the tab-order is out-of-sequence on the range & ave, group, and lines checkboxes because of the use of float. Can you fix?

Also an update from master wouldn't go amiss... :)

Thanks!

ursooperduper:
@jebeck I'll fix the tab order and update this branch. The way the CSS was set up for the footer is a bit funky. I had hoped not to have to rework it, but for the sake of accessibility, it's probably the right thing to do. I'll take care of this today.

jebeck:
Thanks @ursooperduper ! I don't know if you're as big a fan of flexbox as I am, but that's probably how I'd arrange the links instead of floating and get tab order for "free"... We've also been switching to flexbox in a lot of places, so it would be a future-proof kind of change. But feel free to do whatever you think is best.

ursooperduper:
Flexbox fans unite! I'm liking it more and more and you're right, from a future-proofing is the way to go. I'll clean it up that way.

ursooperduper:
@jebeck I cleaned up the CSS styling in the footer, removing the multiple grid-related styles, and use flexbox instead. It should be much easier to add and modify items in the footer moving forward.

jebeck:
Looks good. Yay flexbox \o/

PR #202, opened 2015-01-28
Adds ability to route an external request to different app location based on whether the user is logged in or out.

ursooperduper:
@jh-bate @kentquirk 

This PR is the first of a two parter to fix a bug that occurs when a user clicks the "reset password" link from the Chrome Uploader. Today, when a user clicks that 'reset password' link, they are taken from the uploader to Blip and if they're logged in, you are redirected to the /patients by default. 

To resolve this bug and make future external requests a bit easier to manage, I made some changes that allow us to support URLs that will route to different locations based on whether a user is logged in or out. 

For the Chrome Uploader, reset password link bug, the solution will be to change the reset password link to a new URL (request-password-from-uploader) rather than the default 'request-password-reset'. When the user clicks that link in the Uploader, if they are logged in to Blip, they will be sent to their Profile page where they can update their password. If the user is not logged into Blip, they will be sent to the Request Password Reset page. 

To test, change your path to something like localhost:3000/#/request-password-from-uploader from both logged in and out scenarios. 

Resolves the first task in: https://trello.com/c/ilkdLJKh
Once this PR is approved and all clear, I'll make the corresponding link change in the Uploader.

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -76,6 +76,7 @@ var app = {
   router: router
 };
 
+// List of routes and associated handlers which display/render the routes
 var routes = {
   '/': 'redirectToDefaultRoute',
   '/login': 'showLogin',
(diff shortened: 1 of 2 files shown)


jh-bate:
LGTM

PR #203, opened 2015-01-30
reset state on logout so that login

jh-bate:
@ursooperduper this resolves issue seen in https://trello.com/c/Pub5ZwkN/26-ui-debt-if-there-is-one-account-user-should-be-taken-directly-to-data-not-care-team


ursooperduper:
Ah, perfect! Thanks, @jh-bate

kentquirk:
LGTM.

PR #205, opened 2015-02-03
Sk/update terms

ursooperduper:
@jebeck 

Quick text updates to the terms overlay for Blip. Copy is on the Trello card linked below.

Resolves: https://trello.com/c/ljrwYEP8


jebeck:
@HowardLook is the non-standard capitalization in "I Agree to these Terms" really intended? Just wanted to double-check. Otherwise, LGTM.

HowardLook:
"I agree to these terms" is fine, too, but also OK to push to prod as is
and fix later.

H

jebeck:
Since I asked for an additional fix on #201, @ursooperduper should be able to change the capitalization before we pull in another batch of blip stuff to be tested for prod.

ursooperduper:
@jebeck @HowardLook I will make that capitalization change now. I took the capitalization as it is now from the approved updated copy, so I assumed it had been reviewed and was supposed to be that way.

PR #206, opened 2015-02-05
Jhbate/metrics

jh-bate:
@jebeck this add's some of the missing metrics calls that @brandonarbiter requested. I have tested localhost - there should be no real surprises.

### DIFF
--- a/app/components/patientcard/patientcard.js
+++ b/app/components/patientcard/patientcard.js
@@ -85,9 +85,19 @@ var PatientCard = React.createClass({
       'patientcard-actions--highlight': (!this.props.isNavbar && this.state.highlight === 'view') || this.props.currentPage && this.props.currentPage.match(/(data)$/i)
     });
 
+    var self = this;
+    var handleClick = function(e) {
+      if (e) {
+        e.preventDefault();
+        e.stopPropagation();
+      }
+      self.props.trackMetric('Clicked VDF View Data');
+      self.props.onClick(e);
+    };
+
     return (
       /* jshint ignore:start */
-      <a className={classes} href={this.props.href} onClick={this.props.onClick}>View</a>
+      <a className={classes} href={this.props.href} onClick={handleClick}>View</a>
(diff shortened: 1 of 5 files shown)


app/components/patientcard/patientcard.js
```diff
@@ -85,9 +85,19 @@ var PatientCard = React.createClass({
       'patientcard-actions--highlight': (!this.props.isNavbar && this.state.highlight === 'view') || this.props.currentPage && this.props.currentPage.match(/(data)$/i)
     });
 
+    var self = this;
+    var handleClick = function(e) {
+      if (e) {
+        e.preventDefault();
+        e.stopPropagation();
+      }
```
jebeck: Why are these here? They prevent following the link...?

app/components/patientcard/patientcard.js
```diff
@@ -172,10 +182,19 @@ var PatientCard = React.createClass({
       'patientcard-actions--highlight': (!this.props.isNavbar && this.state.highlight === 'share')  || this.props.currentPage && this.props.currentPage.match(/(share)$/i)
     });
 
+    var self = this;
+    var handleClick = function(e) {
+      if (e) {
+        e.preventDefault();
+        e.stopPropagation();
+      }
```
jebeck: Ditto.

jh-bate:
@jebeck I have updated the PR

jebeck:
Fixed, but some of the metrics calls appear to be repeated. Is that a problem?

From the logs:
Api              | GET /metrics/Clicked%20VDF%20View%20Data
api.js:580 Api              | GET /metrics/Clicked%20Own%20Care%20Team
api.js:580 Api              | GET /metrics/Clicked%20Own%20Care%20Team

You get the second one twice every time you click View from the VDF page.

jh-bate:
I will have a look. It won't cause an issue other than noise in metrics

jebeck:
OK, then LGTM.

PR #209, opened 2015-02-16
move the patients route setting into own function

jh-bate:
@jebeck how does this look? unless I am missing something I think i can also do away with the state.showPatientData also

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -495,80 +495,72 @@ var AppComponent = React.createClass({
       /* jshint ignore:end */
     );
   },
+  setPatientsRoute : function(){
+    // Determine whether to skip the Patients page & go directly to Patient data.
+    // If I have invites to review then display the Patients page
+    //
+    // If there is only one patient you can see data for, go to the patient's data.
+    // Otherwise, display the Patients page.
+    if (_.size(this.state.invites) === 0) {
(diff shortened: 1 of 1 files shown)


app/app.js
```diff
@@ -495,80 +495,72 @@ var AppComponent = React.createClass({
...
+      patients: function(cb) {
+        self.fetchInvites(cb);
       }
-
+    },
+    function(err, results) {
+      self.setPatientsRoute();
+      self.renderPage = self.renderPatients;
+      self.setState({page: 'patients'});
+      trackMetric('Viewed Care Team List');
```
jebeck: Isn't this metric not necessarily correct since the route can vary?

jebeck:
Yes, I agree about getting rid of `state.showPatientData`. I don't think there's much value in having a boolean in the state that is determined exclusively by *other* parts of the state.

app/app.js
jh-bate: very true

jebeck:
I just had another look at this because I was looking at this part of the code to figure out my remaining issue with the name + data mismatch. This doesn't help, and I actually think we don't want to pull it in because it simplifies a bit too much - no access to the care team page at all anymore for people who only have access to their own data. Closing for cleanliness sake...

PR #210, opened 2015-02-17
Jhbate/handle no data store

jh-bate:
@jebeck so this should resolve the issue seen with the beta testers where no storage has been setup. If you could take a look at the proposed solution. Feel free to tell me its a bad idea :)

Note ready to merge until
- wording cleaned up @brandonarbiter 
- state of overlay decided. i.e. should I close it as soon as you click through to fix storage or once the storage is fixed. Also should it be dismissible or should the only option be to click through and fix your storage @brandonarbiter


kentquirk:
I just tweaked the wording a bit on a few of the items.

app/app.js
```diff
@@ -1384,6 +1384,30 @@ var AppComponent = React.createClass({
...
+        <p>{message}</p>
+        {link}
+      </div>
+    );
+
+    this.setState({
+      notification: {
+        type: 'alert',
+        body: body,
+        isDismissable: true
```
jebeck: I defnitely think it should remain dismissable. Could we also auto-dismiss it once your set-up is successful? It's not a great experience to still have the error there after you finish.

jh-bate: Yeah I am just going to close dialog once you click through to the link.

jebeck:
Generally this approach seems like a good idea to me (just a nit about the alert's dismissability), and it works locally. Probably need to remember to do an update merge from master before it comes in, especially since there is still language to be reviewed. Otherwise looks merge-able to me.

jh-bate:
updates addressed, now merging. Possible tweaks to wording after second review but close enough for now

PR #211, opened 2015-02-17
Jhbate/errors show utc

jh-bate:
@jebeck a bit of an error message cleanup and also displaying utc datetime to user

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -1116,12 +1116,12 @@ var AppComponent = React.createClass({
 
     app.api.patient.get(patientId, function(err, patient) {
       if (err) {
-        self.setState({fetchingPatient: false});
-
-        // Patient with id not found, cary on
+        // Patient with id not found, carry on
         if (err.status === 404) {
           app.log('Patient not found with id '+patientId);
-          return;
+          var setupMsg = (patientId === self.state.user.userid) ? usrMessages.ERR_YOUR_ACCOUNT_NOT_CONFIGURED : usrMessages.ERR_ACCOUNT_NOT_CONFIGURED;
+          var dataStoreLink = (<a href="#/patients/new">{usrMessages.YOUR_ACCOUNT_DATA_SETUP}</a>);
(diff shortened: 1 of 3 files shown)


app/app.js
```diff
@@ -1363,13 +1376,10 @@ var AppComponent = React.createClass({
 
         body = (
           <div>
-            <p>
-              {usrMessages.ERR_GENERIC}
-              <a href="/">refresh your browser</a>
```
jebeck: So removing the instruction to try refreshing completely?

jh-bate: I believe so. We should only ask users to do something that we know should work. We tell them to refresh when prob 90% of the time they will get the exact same thing - the app in a broken state

app/onerror.js
```diff
@@ -54,10 +56,9 @@ module.exports = function myErrorHandler(errorMessage, fileUrl, lineNumber, coln
 
     html = [
     '<div style="background: #fefefe;border: gray solid 1px;margin-left: -200px;position: fixed;left: 50%;top: 20%;z-index: 11;width: 390px;padding: 20px 25px;padding-top:30px">',
-      '<p>' + chosenMessage + '</p>',
-      '<p>We were unable to log this error to our server so could you please send us a note at <a style="text-decoration: underline;" href="mailto:support@tidepool.org">support@tidepool.org</a> and we\'ll try to see what broke?</p>',
```
jebeck: So the thing is - it'd actually be helpful for us to prompt for a screenshot almost all the time, y/y? Because having the UTC timestamp will help us pin to the logs so much easier.

jebeck:
A few comments more around strategy than code. Code looks reasonable, but haven't run locally - I'll leave it to you all to test thoroughly. So basically if you agree with my concerns, address them and merge, otherwise just merge :)

app/onerror.js
jh-bate: Yeah good idea

jh-bate:
Merging - will address idea of asking for screenshot when talking with @brandonarbiter.

PR #213, opened 2015-03-06
Jhbate/fix date val

jh-bate:
This PR removes all instances of moment from blip and ensure's we use sundial for date formatting and display.

@jebeck if you could have a look that would be great, and yeah I tested I could have a birthdate on the 31 of March :)

@jebeck FYI - i have updated the version of sundial but haven't yet gone through all usages of it to make sure they are updated.


app/pages/patient/patientinfo.js
```diff
@@ -289,7 +288,15 @@ var PatientInfo = React.createClass({
       return;
     }
 
-    return datetimeUtils.yearsOldText(birthday);
+    var yrsAgo = new Date().getFullYear()-new Date(birthday).getFullYear();
```
jh-bate: @jebeck just putting this option out there for getting the years

jebeck: Yeah, this isn't nuanced enough (nor is the similar code for diagnosis date).

Dates (not just years) need to be involved. If my birthday was March 9th instead of March 6th, this would still tell me today (March 8th) that I'm 31 years old, when I'd still only be 30...

app/pages/patient/patientinfo.js
```diff
@@ -374,12 +391,12 @@ var PatientInfo = React.createClass({
     }
 
     var birthday = formValues.birthday;
-    if (!(birthday && datetimeUtils.isValidDate(birthday))) {
+    if (!(birthday && sundial.isValidDate(birthday))) {
```
jebeck: This doesn't work quite as intended either.

It allows you to save a completely invalid date without exposing any error:
screenshot 2015-03-08 14 06 48

But then silently translates it into a valid date:
screenshot 2015-03-08 14 06 55

I don't know if there's anything we can do about this - I think it's how moment works :/

jh-bate: will double check but it was from memory a straight swap from the isValidDate login in blip to the sundial version

app/pages/patientnew/patientnew.js
```diff
@@ -331,14 +334,14 @@ var PatientNew = React.createClass({
     if (!formValues.birthday) {
       validationErrors.birthday = IS_REQUIRED;
     }
-    else if (!datetimeUtils.isValidDate(formValues.birthday)) {
+    else if (!sundial.isValidDate(formValues.birthday)) {
```
jebeck: Same validation issue. I think this is the deeper bug that Kent was looking at - moment will reject dates like 2014-02-40 (no month has 40 days), but it will not reject 2014-02-30 (even though Feb. never has 30 days), and it might be because it is trying to be "flexible" about whether months are 0-indexed or not, I don't know. I do think we need to figure out a way to invalidate cases like this. It may mean adding a more robust validation function (which we may not want to use everywhere, but only for form validation, because it will probably be more expensive).

app/pages/patient/patientinfo.js
jh-bate: added function into sundial see https://github.com/tidepool-org/sundial/pull/16

app/pages/patient/patientinfo.js
jh-bate: so even according to the moment docs there parsing and subsequent date validation is very loose and hence our use of the isValid can yield a result that isn't strictly valid. 

W can also use a strict parsing which requires that the format and input match exactly and as in this case we have both we can use that. Details http://momentjs.com/docs/#/parsing/string-format/

jh-bate: FYI added this solution to https://github.com/tidepool-org/sundial/pull/16

jh-bate: with the updates in sundial 

screen shot 2015-03-09 at 1 19 29 pm

now fails validation

jebeck:
(Moving this comment, didn't mean to comment on the commit.)

None of the validation works on the initial sign-up, but it does work editing later from the patient info page. Must be a bug in the flow? I just successfully saved a new account with a birthday of March 32nd (which moment translated into April 1st).

jebeck:
Can we fix the capitalization on this:
screenshot 2015-03-10 10 37 21

There's also still a bug in sign-up. When I sign up a new user with diagnosis date 2014-03-31 (I logged this to the console to check the raw date string was right), it displays on patientinfo as 2014-03-30:
screenshot 2015-03-10 10 52 03

jh-bate:
from my logging 

bday raw  2000-02-28
bday formated  02/28/2000
bday raw  2014-03-31
bday formated  03/31/2014

jebeck:
Trying creating a new user with diagnosis date 2014-03-31 then view it on the patientinfo page. That's where the problem is.

jebeck:
LGTM. Merge at will.

jh-bate:
updated to use new mask translation and no longer saving time info for these dates.

Tested with a bday of 31-Mar-1951 and dday of 28-Feb-2001. Toggled around those dates (i.e. setting dday to the valid date of 29-Feb-2008 which was a leap year) and validation looks good.

PR #217, opened 2015-03-17
Jhbate/t and c

jh-bate:
NOTE: these updates are in conjunction with PR https://github.com/tidepool-org/platform-client/pull/48, I will also need to make an adjustment to shoreline so that we can return the terms for a user also. @HowardLook is getting the actual T&C's that will be used and published on our site verified by the lawyers. Until then this is NOT ready to be merged.

@jebeck @kentquirk please have a look at the code it's and make any comments about this. Thanks


app/app.js
```diff
@@ -969,19 +950,22 @@ var AppComponent = React.createClass({
...
     trackMetric('Signup Verified');
   },
-
   handleAcceptedTerms: function() {
-    this.setState({
-      showingAcceptTerms: false
+    var self = this;
+    var acceptedDate = sundial.utcDateString();
+    app.api.user.acceptTerms({ terms: acceptedDate },function(err) {
+      if(_.isEmpty(err)){
```
jebeck: This was a little hard to read, since we're usually doing `if (err) { handle error }` whereas this is the opposite. Was there a specific reason you wanted handling the error to be the default action here?

app/components/termsoverlay/termsoverlay.js
```diff
@@ -52,6 +49,28 @@ var TermsOverlay = React.createClass({
     );
   },
 
+  websiteTerms: function() {
+    return React.DOM.iframe({
+      className         : "terms-overlay-iframe",
+      src               : "http://tidepool.org/products/blip",
+      scrolling         : "no",
```
jebeck: Why no scrolling?

app/components/termsoverlay/termsoverlay.js
```diff
@@ -52,6 +49,28 @@ var TermsOverlay = React.createClass({
...
+      frameborder       : "0",
+      allowTransparency : "true"
+    });
+  },
+
+  terms: function() {
```
jebeck: Why are we keeping this around if we're not using it anymore?

jebeck:
My biggest question is about the lack of scrolling in the iframe.

Also, in terms of app flow, once you're on the terms page, you can't log out, can't do *anything* but accept the terms. Is that our intention? I ran into this locally because I was setting up a bunch of accounts, and I didn't finish on one of them, and now I don't even know which account it is. Probably not something our users are going to run into, but it feels a bit odd to me that you can't even log out without accepting the terms.

jh-bate:
@jebeck lack of scrolling was a mistake - fixed now 

@brandonarbiter thoughts on flow / lack of ability to log out? I think this is consistent with our current hard coded T&C so maybe and idea for the future

app/app.js
jh-bate: yeah no reason - made sense at the time but makes no sense on a second look :) Fixed

jebeck:
LGTM now. Merge at will when we figure out what the path forward is given the state of the current text.

jh-bate:
@jebeck if you could please take an initial look at this now - thanks.

 FYI - local host will require the latest version of shoreline running

jh-bate:
FYI - for an initial test build on devel I had to show the wrong T&C pages as they need to be hosted using https. Before this PR is merged I will correct the URL's to those that @HowardLook points me too

HowardLook:
Updated Terms of Use and Privacy Policy are now hosted here:
https://tidepool.org/terms-of-service/
https://tidepool.org/privacy-policy/

Still need to figure out how to make them use the legal CSS.

app/components/app/app.js
```diff
@@ -127,6 +127,15 @@ var AppComponent = React.createClass({
     };
   },
 
+  initAppUtils: function(injectedUtils){
```
jebeck: What's the purpose of this? If it's for test mocking, there are better ways... Do a `git grep rewire` to see how [rewire](https://github.com/jhnns/rewire) is being used for dependency injection in tests.

app/components/termsoverlay/termsoverlay.js
```diff
@@ -13,82 +13,217 @@
...
-
 var React = require('react');
+var LoginNav = require('../loginnav');
 
 var TermsOverlay = React.createClass({
   propTypes: {
     onSubmit: React.PropTypes.func,
-    trackMetric: React.PropTypes.func.isRequired
+    trackMetric: React.PropTypes.func.isRequired,
+    ages: React.PropTypes.object,
```
jebeck: If you're defaulting these props, might as well make them `isRequired` - will never fail (and I prefer/I think it's best practice to require props whenever possible). So related, can we also require `onSubmit` or no?

app/components/app/app.js
jh-bate: yeah tests fail every-time otherwise b/c of the chrome check. Will have a look at rewire

app/components/termsoverlay/termsoverlay.js
jh-bate: sounds good to me

app/components/termsoverlay/termsoverlay.js
```diff
@@ -13,82 +13,217 @@
...
+    messages: React.PropTypes.object
+  },
+  getDefaultProps: function(){
+    return {
+      ages: {
+        OF_AGE : { value: '>=18', label: ' I am 18 years old or older.'},
+        WITH_CONSENT : { value: '13-17', label: ' I am between 13 and 17 years old. You\'ll need to have a parent or guardian agree to the terms on the next screen.' },
```
jebeck: Dunno if this language if still up for debate, but the switch from 'I' as the subject, to 'You'll need to have...' is confusing. CC @brandonarbiter @HowardLook

app/components/termsoverlay/termsoverlay.js
```diff
@@ -13,82 +13,217 @@
...
+        <p className='terms-overlay-sorry-message'>{this.props.messages.SORRY_NOT_OF_AGE}</p>
+        {backBtn}
+      </div>
+    );
   },
+  render: function() {
+
+    var content = this.renderAgeConsentStep();
+
+    if(this.state.ageConfirmed && this.state.ageSelected !== this.props.ages.NOT_OF_AGE.value){
```
jebeck: Add parens around the second conjunct (after the `&&`) for logic clarity and break these onto two lines for code readability? Same with the next one.

app/components/termsoverlay/termsoverlay.js
```diff
@@ -13,82 +13,217 @@
...
+        <LoginNav hideLinks={true} trackMetric={this.props.trackMetric} />
+        <div className='terms-overlay-content terms-overlay-box'>
+          {content}
+        </div>
+      </div>
+    );
 
-  handleSubmit: function(e) {
+  },
+  websiteTerms: function() {
```
jebeck: The iframes take quite a bit to load - I'm thinking we might need a loading message, otherwise could be confusing/people might click away.

app/components/termsoverlay/termsoverlay.js
```diff
@@ -97,9 +232,9 @@ var TermsOverlay = React.createClass({
 
     var submit = this.props.onSubmit;
     if (submit) {
```
jebeck: Related to my question above about requiring the `onSubmit` prop...why would we ever expect it to be absent??

jebeck:
A few additional comments, not in line.

I think the (lack of) spacing in these buttons is an accessibility/usability issue:
screenshot 2015-11-03 16 03 23

CC @brandonarbiter @HowardLook ^^^

Please remove app/preprocessor.js (was for Jest tests).

Haven't looked at the tests yet, since I know there's still some work to be done there re: rewire. I believe the React 0.14 update also had implications for `React.withContext`, so it might be good to touch base with @GordyD about compatibility with the soon-to-come update in the tests.

app/components/termsoverlay/termsoverlay.js
jh-bate: i believe it was habit from other components

app/components/termsoverlay/termsoverlay.js
jh-bate: its the same as the original language proposed by @HowardLook can change if needed

app/components/termsoverlay/termsoverlay.js
jh-bate: yeah had same thought but will leave up to @HowardLook and @brandonarbiter to decide if its good enough

test/unit/components/termsoverlay.test.js
```diff
@@ -7,7 +7,7 @@ var expect = chai.expect;
 var TermsOverlay = require('../../../app/components/termsoverlay');
 
 describe('TermsOverlay', function () {
-  
+
   describe('render', function() {
     it('should console.warn when trackMetric not set', function () {
```
jebeck: Add tests like these for the other required props now that they are such?

test/unit/components/termsoverlay.test.js
```diff
@@ -29,4 +30,186 @@ describe('TermsOverlay', function () {
...
+        expect(termsElem.state.agreed).to.equal(false);
+        expect(termsElem.state.agreedOnBehalf).to.equal(false);
+
+        var checkboxes = TestUtils.scryRenderedDOMComponentsWithTag(termsElem,'input');
+
+        expect(checkboxes.length).to.equal(2);
+
+        var agreed = checkboxes[0];
+        var agreedOnBehalf = checkboxes[1];
+        //only check one
```
jebeck: Maybe a third test parallel to this one where you only check one of two boxes, but you check the other one?

test/unit/components/termsoverlay.test.js
jh-bate: yeah - just updating so it covers all required props and will fail if new required props are added but bit tested for

test/unit/components/termsoverlay.test.js
jh-bate: cool - done

jebeck:
LGTM - with the known caveats around the untestability of the top-level app state and routing.

PR #219, opened 2015-04-21
lodash _.first doesn't return an array

jebeck:
I believe this addresses both bugs recently found on staging: https://trello.com/c/2I11ddBi and https://trello.com/c/0zfGV7oG.

The issue is that `_.first` does not return an array, and we were trying to extract from it at index 0. Possibly/probably(?) this is a change in lodash and came in with the dependency updates.

Ping @kentquirk

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -1496,7 +1496,7 @@ app.start = function() {
   var self = this;
 
   this.init(function() {
-    self.component = React.renderComponent(
+    self.component = React.render(
       /* jshint ignore:start */
       <AppComponent />,
       /* jshint ignore:end */
--- a/app/components/messages/__tests__/messages-test.js
+++ b/app/components/messages/__tests__/messages-test.js
@@ -1,0 +1,51 @@
+/**
+ * Copyright (c) 2015, Tidepool Project
+ *
+ * This program is free software; you can redistribute it and/or modify it under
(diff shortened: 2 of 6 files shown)


jebeck:
Updated this, ping @kentquirk @jh-bate 

To run the regression test, you have to switch to node 0.10.x and either clear and reinstall the dependencies or do an `npm rebuild`.

kentquirk:
I don't know how this is going to work if travis is set up to run tests against v0.12 (which it is). Also, how useful is it, really, to specify that we're using v0.12 but our tests have to run against v0.10? It doesn't really give us confidence in either the positive or negative cases.

jh-bate:
Yeah I think we have realised its potentially of minimal use at this stage with the whole jest and lack of v0.12 support. I also invested sometime with the terms and conditions to do tests - I think at this stage we will add the tests for 'documentation' and then have to either switch test framework or see if jest updates in the near future. Either way its a bit crap :(

jebeck:
Yeah, I didn't realize that the situation was as problematic as it is. But hey, now we know. I'm going to sign up for updates to the Jest repo and we can just see how it plays out in the next month or so. If they (or someone else who's desperate for the update) update it to get it to work with 0.12, then we can try to expand the tests. If not, we'll start thinking about other solutions.

jh-bate:
LGTM - test framework will have to wait

PR #220, opened 2015-04-22
make sure edited note text is rendered right away

jebeck:
Nick found another regression - edited note text wasn't displaying until after closing and reopening the note. `git blame` showed yours truly to blame for the regression :/

Ping @jh-bate for a quick review.

No regression test at the moment, could add one since I've already got started on Jest stuff for this component. LMK what you'd prefer @jh-bate.

### DIFF
--- a/app/components/messages/__tests__/message-test.js
+++ b/app/components/messages/__tests__/message-test.js
@@ -1,0 +1,39 @@
+/**
+ * Copyright (c) 2015, Tidepool Project
+ *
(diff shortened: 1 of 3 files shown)


jh-bate:
Lets add the regression test. @jebeck as discussed lets have a bit of faith in jest and at worst adding a test will document an issue we have had in prod.

jh-bate:
test now added so LGTM

PR #223, opened 2015-05-20
bump platform-client version for addition of GUIDs

jebeck:
Tag @jh-bate, you're it!


jh-bate:
LGTM - did a quick sanity check locally

PR #224, opened 2015-05-21
Jhbate/oauth params

jh-bate:
@jebeck I am just going to update the version of platform client also and link that to this PR. 

The platform-client version referred to is via commit hash as it is a dependency on this change and also has a PR see https://github.com/tidepool-org/platform-client/pull/50

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -200,7 +200,7 @@ var AppComponent = React.createClass({
         self.fetchUser();
         self.setState({authenticated: true});
         trackMetric('Logged In with OAuth');
-        //go to the specified paitient if there is one
+        //go to the specified patient if there is one
         if(_.isEmpty(data.target)){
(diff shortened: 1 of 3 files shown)


app/app.js
```diff
@@ -191,6 +191,29 @@ var AppComponent = React.createClass({
...
+  doOauthLogin:function(accessToken){
+
+    var self = this;
+    app.api.user.oauthLogin(accessToken, function(err, data){
+      if(_.isEmpty(err)){
+        app.log('Logged in via OAuth');
+        self.fetchUser();
+        self.setState({authenticated: true});
+        trackMetric('Logged In with OAuth');
+        //go to the specified paitient if there is one
```
jebeck: Spelling if you wouldn't mind fixing - `patient`

jh-bate: no worries  - updated it :)

PR #225, opened 2015-05-27
default to the timezone from the most recent upload metadata for display

jebeck:
This isn't as sophisticated as I'd hoped. It's a chicken-and-egg problem to add the timezone to the data-viewing URL when we don't know what timezone to display in until *after* we've loaded the data and found the most recent metadata object with a timezone.

So:
- if a timezone is in the query params before the hash (e.g., `?timezone=US-Pacific#/`), that timezone takes precedence
- otherwise, blip defaults to the timezone from the most recent upload metadata for display, but this timezone will *not* be displayed in the URL (the only way to do that requires either a full page refresh and re-fetch of data (if in query params) or setting a new route and thus re-fetching data) :(

We could just take the setting of the timezone via query params away now, **but**:
- most people won't know it's there
- it's invaluable for testing - in particular misspelling a timezone in the query params will become our only way to turn *off* timezone-aware display if we need to

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -1236,7 +1236,7 @@ var AppComponent = React.createClass({
       return Date.parse(d.time);
     }).reverse()[0];
     var timePrefsForTideline;
-    if (!_.isEmpty(mostRecentUpload.timezone)) {
+    if (!_.isEmpty(mostRecentUpload) && !_.isEmpty(mostRecentUpload.timezone)) {
       try {
         sundial.checkTimezoneName(mostRecentUpload.timezone);
         timePrefsForTideline = {
--- a/app/app.js
+++ b/app/app.js
@@ -1235,14 +1235,24 @@ var AppComponent = React.createClass({
     var mostRecentUpload = _.sortBy(_.where(data, {type: 'upload'}), function(d) {
       return Date.parse(d.time);
     }).reverse()[0];
-    var timePrefsForTideline = {
-      timezoneAware: true,
-      timezoneName: mostRecentUpload.timezone
-    };
+    var timePrefsForTideline;
+    if (!_.isEmpty(mostRecentUpload.timezone)) {
+      try {
+        sundial.checkTimezoneName(mostRecentUpload.timezone);
+        timePrefsForTideline = {
+          timezoneAware: true,
+          timezoneName: mostRecentUpload.timezone
+        };
+      }
+      catch(err) {
+        app.log(err);
+        app.log('Upload metadata lacking a valid timezone!', mostRecentUpload);
+      }
+    }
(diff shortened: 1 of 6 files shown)


jh-bate:
>otherwise, blip defaults to the timezone from the most recent upload metadata for display, but this timezone will not be displayed in the URL ...

So will I know what timezone my data is displaying in, how would I find out? do I care?

app/app.js
```diff
@@ -1232,11 +1232,42 @@ var AppComponent = React.createClass({
...
+    if (!_.isEmpty(mostRecentUpload) && !_.isEmpty(mostRecentUpload.timezone)) {
+      try {
+        sundial.checkTimezoneName(mostRecentUpload.timezone);
+        timePrefsForTideline = {
+          timezoneAware: true,
+          timezoneName: mostRecentUpload.timezone
+        };
+      }
+      catch(err) {
+        app.log(err);
```
jh-bate: so if I don't have a timezone and there isn't one from the query params what is the expected behaviour?

app/components/chart/daily.js
```diff
@@ -33,6 +33,7 @@ var Daily = React.createClass({
   propTypes: {
     bgPrefs: React.PropTypes.object.isRequired,
     chartPrefs: React.PropTypes.object.isRequired,
+    timePrefs: React.PropTypes.object.isRequired,
```
jh-bate: Simple question - so timePrefs were part of chartPrefs but have been now broken out - what was this to achieve? Just curious more than anything

jh-bate:
Code LGTM - I have a few general questions though

app/app.js
jebeck: If you don't have a timezone from your data and there isn't one in query params, we default to the non-timezone aware delay (the current prod way of doing things). That's [here](https://github.com/tidepool-org/blip/blob/jebeck/tz-for-utc-bootstrap/app/app.js#L146) in the code. I did just remove the old hard-coded default timezoneName and change that to `null`.

app/components/chart/daily.js
jebeck: Basically because there are now a couple places in app.js where they might be modified - via the query params or after fetch of data (when we check most recent upload metadata for the timezone), so it was easier to break them out than to keep deep-cloning `chartPrefs` and updating them there.

jebeck:
It's true that currently you can't find out what timezone your data is currently displaying in (well, unless you think to look in the console - it's logged there). I don't think you care much. The differences between the timezone-naive and timezone-aware display are only really apparent around device time changes (DST or travel).

jh-bate:
LGTM - thanks for the details

PR #226, opened 2015-05-29
Add onClickRefresh property on Footer in settings.js to make Refresh …

GordyD:
…button fetch data

Simple fix, just add the missing property that should be passed into the Settings component so that clicking the Refresh button has the appropriate handler.


jebeck:
Excellent, merging!

PR #227, opened 2015-06-03
Jhbate/fix pwreset

jh-bate:
@jebeck so the key wasn't being passed through when trying to perform a pw-reset. As discussed this would be a good candidate for a regression test.

As part of the change I made a more general function getQueryParam, that will log if the asked for param doesn't exist.

The current fixes the issue but question is if we don't get a resetKey what do we think we should do? Thoughts ....

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -1449,43 +1449,49 @@ var AppComponent = React.createClass({
   },
 
   showRequestPasswordReset: function() {
-    this.renderPage = this.renderRequestPasswordReset;
-    this.setState({
-      page: 'request-password-reset'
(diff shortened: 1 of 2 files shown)


jebeck:
LGTM, and tested against devel.

My current hunch about why this just broke is because the updated version of React we're now using does *not* allow any old prop to be named `key` because that's a reserved prop name. @jh-bate as soon as you've scrubbed for other places where we did this, you're good to merge.

jh-bate:
@jebeck I  looked through and no other instances of prop's named `key` were found.

PR #230, opened 2015-06-04
store patient data in hash under userid to prevent name + data mismatch

jebeck:
@jh-bate I believe this fixes the issue, and I hope I found all the uses of patient data so I'm not creating new bugs by changing what the prop looks like. I'd definitely appreciate another pair of eyes taking a close look to trace through uses of `patientData` as a prop to make sure nothing's been left out.

My plan is to try to get this fix in ASAP. On a separate branch I am pulling in @GordyD's work so far and there I will try to start working on (a) a regression test and (b) the clean up on `app/components/peoplelist.js` that we discussed (i.e., getting rid of the attempt to mutate props).

P.S. A side benefit of this change is it paves the way quite nicely for caching data that's already been fetched so we don't refetch every time you navigate between people :)

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -922,9 +922,11 @@ var AppComponent = React.createClass({
     );
     /* jshint ignore:end */
   },
-  handleUpdatePatientData: function(data) {
+  handleUpdatePatientData: function(userid, data) {
+    var patientData = _.cloneDeep(this.state.patientData);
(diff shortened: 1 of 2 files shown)


jh-bate:
LGTM - as discussed we will include a regression test for the separately once the new testing updates are merged

PR #232, opened 2015-06-05
get rid of long browser hang

jebeck:
This fixes the issue I found on staging after deploying #230 yesterday. It turned out the `_.cloneDeep` I'd added was hanging the browser for *seconds* when there was actually something already in `this.state.patientData` (understandably, because it was deep-cloning all of the device data). The deep clone isn't actually necessary here, so I changed it to `_.clone`, and that works.

As a reminder, the issue on staging was this: log into an account, go to view data, navigate away (first data fetch still happening in background), navigate to view away (second data fetch starts), and then navigate away again, to the edit profile page. I noticed I couldn't keep editing my profile after a bit - app was frozen. In the console it was apparent this was happening after both the first and second data fetches (and tideline data preprocessing) concluded - turns out that's because it was stuck trying to deep clone all the device data before updating the app state. (It would finish eventually, so it turned out to not be a complete app freeze, just looked like it because it took *forever*.)

Ping @jh-bate for review.

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -119,9 +119,9 @@ function objectDifference(destination, source) {
   var result = {};
 
   _.forEach(source, function(sourceValue, key) {
-    var destinactionValue = destination[key];
-    if (!_.isEqual(sourceValue, destinactionValue)) {
-      result[key] = destinactionValue;
+    var destinationValue = destination[key];
+    if (!_.isEqual(sourceValue, destinationValue)) {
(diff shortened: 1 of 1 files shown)


jh-bate:
LGTM  - I guess the interaction is a little more complicated than we thought. Something else to consider with @GordyD while we clean up / rationalise / integrate better testing framework

PR #233, opened 2015-06-05
new testing strategy

jebeck:
**WIP: don't merge.**

Will be merging this into master as soon as we've deployed the urgent bug fix around data + name mismatch to prod.


app/pages/patient/patientinfo.js
```diff
@@ -399,6 +423,22 @@ var PatientInfo = React.createClass({
       return 'Diagnosis date needs to be a valid date';
     }
 
+    currentDateObj = currentDateObj || new Date();
+    var birthdayDateObj = sundial.parseFromFormat(birthday, FORM_DATE_FORMAT);
+    var diagnosisDateObj = sundial.parseFromFormat(diagnosisDate, FORM_DATE_FORMAT);
+
+    if (birthdayDateObj > currentDateObj) {
```
jebeck: Was just smoke-testing as I'm preparing to merge things in, and these comparisons aren't *quite* working as intended. I'm allowed to set my birthday or diagnosis date to 2015-06-12 even though it's still 2015-06-11 here in Pacific time. The reason for this is that > is comparing the raw integer values (i.e., ms since 1970-01-01 UTC) of the date objects involved, and even though it's still 2015-06-11 in Pacific time it's already 2015-06-12 in UTC. Without a timezone as a third arg to `sundial. parseFromFormat` (link: [sundial source](https://github.com/tidepool-org/sundial/blob/master/lib/datetimeWrapper.js#L329)), the `birthdayDateObj` and `diagnosisDateObj` are being created as UTC, so `2015-06-12T00:00:00.000Z` is not evaluating as greater than my *UTC* now of `2015-06-12T05:38:10.608Z`.

Since we don't *have* a current timezone to pass into `parseFromFormat` as the third arg (as you know @GordyD, these are difficult to infer!), the best solution *may* be to do a string comparison of ISO date strings, taking something like `birthdayDateObj.toISOString.slice(0,10)` and comparing it to an ISO-formatted date string constructed from the local `currentDateObj`. (Annoyingly, JavaScript date does not allow you to get the local date out easily as an ISO-formatted string, so as far as I can tell this would have to be done by something like `sundial.translateMask(currentDateObj.toDateString(), 'ddd MMM DD YYYY', 'YYYY-MM-DD')`.)

Hmm, thinking aloud here, but another way to do it might be to create the `currentDateObj` in a manner more similar to 
the creation of the Date objects in the tests, e.g.:
```
var now = new Date();
var currentDateObj = new Date(now.getFullYear(), now.getMonth(), now.getDate());
```
Yeah, I think I like option 2, since it guarantees creating a local midnight for the relevant date, although let me know your thoughts @GordyD.

app/pages/patient/patientinfo.js
```diff
@@ -298,9 +297,10 @@ var PatientInfo = React.createClass({
     } else {
       return 'Birthdate not known';
```
jebeck: I know you didn't introduce this @GordyD but I never knew it was there, and it makes me cringe. (It's somewhat rare but possible for infants to be diagnosed with type 1 diabetes, so this is a possible use case for a parent creating an account to track their child's data.) Let's change to 'Born this year'? Or maybe add an `else if` for `yrsAgo === 0` returning 'Born this year' and keeping the `else` for the truly unknown edge cases?

app/pages/patient/patientinfo.js
GordyD: Good catch!

Datetimes and timestamps are certainly interesting animals! There are a number of cases that could be problematic/work in ways that are unexpected in this area when you don't give the user explicit control over the timezone in which they are speaking. E.g.

The problem you identify.
Creating your profile in one timezone, when you were born in another (In the US, born in UK). 
Entering a the birthday of a new born in another timezone ahead of UTC.

I think what you suggest as option 2 is the best route to go. Even though the birth date and diagnosis dates are stored in UTC it restricts acceptable values based on the current date in the current time zone.

I'll update accordingly now.

app/pages/patient/patientinfo.js
GordyD: Will do.

GordyD: One thing I have just been triangulating around is the behaviour of dateDifference in sundial (but really the behaviour of diff in moment.js). 

If you take a look at the [documentation](http://momentjs.com/docs/#/displaying/difference/), there i is an edge case for using diff in this regard when looking at at difference between dates within 1 year. The point is best illustrated with a [jsFiddle](http://jsfiddle.net/ho70pxn4/2/) I have created.

Basically, when a diff value is between -1 and 0, using date1.diff(date2, 'years') in this range rounds the value up to 0. Whereas, when returning a floating point, by setting the optional third value to true, you can see that indeed the underlying value is negative.

In relation to this request, it can cause an oddity here where if a birthdate is in the future, but within 365 days of the current date, the text rendered would be 'Born this year'/'Diagnosed this year'

For the time being I'll leave this edge case in, perhaps in the future we can address this by allowing a  4th argument into dateDifference to allow floating point values to be returned.

app/pages/patient/patientinfo.js
jebeck: Awesome, thanks! In the future we will have a timezone to pass through to sundial because we will be offering a UI for deciding what timezone to view data in (which will be defaulted to the timezone most recently selected from the uploader for a device upload), so we will have a bit of a better solution to this, but for now option 2 sounds good - LMK when you've updated. I'm looking at #234 now and we may need to coordinate updates if that's ready to merge.

app/pages/patient/patientinfo.js
jebeck: Oh yeah, this is exactly the kind of method behavior that makes me not a huge fan of moment. :( The oddity is definitely acceptable for now, and in the future we can either adjust sundial in that way, or maybe not use moment's method for this, especially if this is the only place we're consuming that sundial method, which may be the case.

app/pages/patient/patientinfo.js
GordyD: This has been updated on my fork now :-)

jebeck: Ah, I forgot you're still on a fork. Thanks for the reminder! Will see what shenanigans I can do with PRs to update everything.

jebeck:
I brought in your additional commits via `git cherry-pick`. This is very close to ready, except there's a test broken now ;) (Expecting 'Birthdate not known' where it should be 'Born this year')

GordyD:
I can't find the broken test - all tests pass here :/

jebeck:
Wait really? I still have the test broken after the git pull:
```
FAILED TESTS:
  PatientInfo
    getAgeText
      ✖ should return unknown birthday if birthdate is in future
        PhantomJS 1.9.8 (Mac OS X)
      AssertionError: expected 'Born this year' to equal 'Birthdate not known'
          at /Volumes/Tidepool/blip/node_modules/chai/chai.js:875
          at assertEqual (/Volumes/Tidepool/blip/node_modules/chai/chai.js:1399)
          at /Volumes/Tidepool/blip/node_modules/chai/chai.js:4031
          at /Volumes/Tidepool/blip/tests.webpack.js:669

npm ERR! Test failed.  See above for more details.
```

GordyD:
Intriguing! Is this a timezone issue in the tests themselves. E.g because you are running them in Pacific time and I am in BST.

jebeck:
Still failing for me even after f7788c42d0b9c9f27de37549b20d9e7a8bb94b4a. Probably is a timezone issue, not sure. You might be able to replicate by switching your system time to a different timezone in your VM (that fools other browsers, at least, but not sure about PhantomJS).

jebeck:
Update: it just occurred to me to follow my own advice and see if I can replicate your success by changing my system time (especially as that may be easier for me to do in OS X than you in your VM). When I turn off OS X's auto-timezone detection and select London, all tests pass. So: this is definitely a timezone issue in the tests.

GordyD:
I just realised the text fixtures were using new Date too, so I have converted them to Date.UTC too. This 'should' ensure no timezone usage within the tested code AND the tests.

I've run the tests after adjusting the timezone using:

`sudo timedatectl set-timezone America/Los_Angeles`

test/unit/components/chart/daily.test.js
```diff
@@ -0,0 +1,105 @@
...
+
+var Daily = require('../../../../app/components/chart/daily');
+
+require('tideline/css/tideline.less');
+require('../../../../app/core/less/fonts.less');
+require('../../../../app/style.less');
+window.localStorage.debug = true;
+
+describe('Daily', function () {
+  describe('render', function() {
```
jebeck: Hmm, I imagine these weren't working because of troubles providing all the expected props? I'll play around with it a bit, but in general this might be just be exposure of the fact that the interface between blip and tideline needs work (and since I'm hoping we'll be working on a new tideline in the not-so-distant future, it's probably not worth putting effort into fixing now).

GordyD: Oops, I didn't actually mean to commit this. Must have snuck in. Shall I just remove it?

jebeck: Don't worry about it at the moment - I started playing around with it locally and have non-pushed work in flight right now that I may get a chance to get back to later today. I realize now the issue is actually not with the tideline interface but interestingly maybe with PhantomJS's abilities to simulate DOM elements? Although it also seems that if we can figure out how to mock the inner `DailyChart` React component, then we'll also be able to get around this...

GordyD: Yes, that would help greatly! As soon as we can mock, the testing world is our oyster!

jebeck:
Superceded by #242, so closing.

Issue #228, opened 2015-06-04
React warnings re: API and usage changes (required action for future React upgrades)


dotMR:
When running in 'Quick Start' mode there are currently React warnings in the dev console. To eliminate the warnings some code changes are required. These changes will need to be made to support upgrades to React >= 0.13 as well.

- React.PropTypes.component is now React.PropTypes.element
- React.PropTypes.renderable is now React.PropTypes.node
- 'transferPropsTo' is deprecated

```
Warning: React.PropTypes.component will be deprecated in a future version. Use React.PropTypes.element instead.
Warning: React.PropTypes.renderable will be deprecated in a future version. Use React.PropTypes.node instead.
Warning: transferPropsTo is deprecated. See http://fb.me/react-transferpropsto for more information.
Warning: Invalid prop `patientsComponent` supplied to `Invitation`, expected a ReactElement. Check the render method of `Patients`.
```

More info:
https://facebook.github.io/react/blog/2014/10/28/react-v0.12.html

jebeck:
Fixed in #234, thanks again!

PR #234, opened 2015-06-08
eliminate React warnings re: api deprecation and increase test coverage

dotMR:
Hi,
When I was checking out your app I noticed some warnings in the console and here is a patch to fix them. Do let me know if there is a better option than targeting this branch - I don't know the codebase well so I wanted to include test coverage to validate that my changes weren't breaking anything, and the best place to accomplish that at the moment was in this branch (as I wasn't able to run tests effectively in master).

##### Commit Message
- removed usage of 'transferPropsTo'
- changed React.PropTypes.renderable -> React.PropTypes.node
- removed unused property 'patientsComponent'
- added test coverage for the changes

Issue: #228

### DIFF
--- a/app/components/invitation/invitation.js
+++ b/app/components/invitation/invitation.js
@@ -22,7 +22,6 @@ var personUtils = require('../../core/personutils');
 var Invitation = React.createClass({
   propTypes: {
     invitation: React.PropTypes.object,
-    patientsComponent: React.PropTypes.component,
     onAcceptInvitation: React.PropTypes.func,
     onDismissInvitation: React.PropTypes.func,
     trackMetric: React.PropTypes.func.isRequired,
--- a/app/pages/patient/patient.js
+++ b/app/pages/patient/patient.js
@@ -161,15 +161,7 @@ var Patient = React.createClass({
 
   renderPatientTeam: function() {
     return (
-      <PatientTeam
-        user={this.props.user}
-        patient={this.props.patient}
-        pendingInvites={this.props.pendingInvites}
-        onChangeMemberPermissions={this.props.onChangeMemberPermissions}
-        onRemoveMember={this.props.onRemoveMember}
-        onInviteMember={this.props.onInviteMember}
-        onCancelInvite={this.props.onCancelInvite}
-        trackMetric={this.props.trackMetric}/>
+      <PatientTeam {...this.props} />
     );
   },
 });
--- a/app/pages/patient/patient.js
+++ b/app/pages/patient/patient.js
@@ -160,7 +160,17 @@ var Patient = React.createClass({
   },
 
   renderPatientTeam: function() {
-    return this.transferPropsTo(<PatientTeam />);
+    return (
+      <PatientTeam
+        user={this.props.user}
+        patient={this.props.patient}
+        pendingInvites={this.props.pendingInvites}
+        onChangeMemberPermissions={this.props.onChangeMemberPermissions}
+        onRemoveMember={this.props.onRemoveMember}
+        onInviteMember={this.props.onInviteMember}
+        onCancelInvite={this.props.onCancelInvite}
+        trackMetric={this.props.trackMetric}/>
+    );
   },
 });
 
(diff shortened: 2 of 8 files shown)


jebeck:
Before we can look at this, we need you to agree to our Volunteer/Contributer License Agreement, discussed [here](http://developer.tidepool.io/how-to-help/) and available [here](http://developer.tidepool.io/files/tidepool-vcla.pdf). Please follow the instructions from the latter to add the appropriate comment to this pull request if you agree to the terms in the agreement. Thanks!

dotMR:
Thanks. I did see the contributor guidelines and included my agreement disclaimer in the pull request body above, but I'm happy to move it to a separate comment if that is the official place for it.

dotMR:
I agree to the terms of Tidepool Project’s Volunteer/Contributor License Agreement v1.0
as it exists at http://tidepool-org.github.io/TidepoolVCLA.pdf on Jun 8, 2015.

jebeck:
Weirdly, I can't see the VCLA agreement in the body of the PR here on GitHub, only in my e-mail?? Did you edit it? In any case, thanks for the agreement by way of comment.

Will be reviewing this likely later this week. Currently have other priorities and need to wait for some recent changes on master to go through QA. After that will be trying to pull in new-testing branch hopefully along with this. Will let you know if I have any questions/comments.

Thanks again!

dotMR:
No black magic going on, I updated the PR body to remove the disclaimer and added as a comment instead :)

Thanks for the update - no rush on my end.

app/components/invitation/invitation.js
```diff
@@ -22,7 +22,6 @@ var personUtils = require('../../core/personutils');
 var Invitation = React.createClass({
   propTypes: {
     invitation: React.PropTypes.object,
-    patientsComponent: React.PropTypes.component,
```
jebeck: So I guess it looks like we're just not actually using this in the Invitation component. Can we remove where this prop is passed in too?

app/pages/patient/patient.js
```diff
@@ -160,7 +160,17 @@ var Patient = React.createClass({
   },
 
   renderPatientTeam: function() {
-    return this.transferPropsTo(<PatientTeam />);
```
jebeck: All other things being equal, I'd prefer the JSX equivalent of `transferPropsTo` for this, which seems to be the recommendation post-deprecation of `transferPropsTo`: [JSX Spread Attributes](https://facebook.github.io/react/docs/jsx-spread.html). Is there a reason you didn't think this was a good idea here @dotMR?

CC @GordyD because we've been talking about trying to do the JSX spread attributes in more places, potentially.

test/unit/components/messages/message.test.js
```diff
@@ -0,0 +1,61 @@
...
+var expect = chai.expect;
+
+var Message = require('../../../../app/components/messages/message');
+
+describe('Message', function () {
+  describe('getInitialState', function() {
+    it('should return an object with editing set to false', function() {
+      console.warn = sinon.spy();
+      // TODO: consider requiring Proptype: theNote (maybe define as a shape as well?)
+      //   -> Properties of theNote are accessed in componentDidMount()
```
jebeck: Feel free to address this TODO in this PR. Generally we do prefer `isRequired` on our props, but clearly there's some inconsistency across the codebase... Here I think we should be OK (i.e., without any other code changes) adding `isRequired` on `theNote` and `onSaveEdit`; possibly also on `imageSize`.

app/pages/patient/patient.js
dotMR: @jebeck I was actually on the fence. With that many props I see the argument that passing everything manually is tedious and verbose, but for me first looking at the code base I liked the explicitness of seeing exactly what was being passed. It also made it easier to try to find potential orphaned props. But I'm happy to change to spread if that is the desire going forward. They make quick mention of it in the docs here: [Manually Transferring Props](https://facebook.github.io/react/docs/transferring-props.html#manual-transfer)

app/components/invitation/invitation.js
dotMR: Yes, of course. I had intended to do that. Good catch!

test/unit/components/messages/message.test.js
dotMR: The way it is currently coded I don't believe imageSize or onSaveEdit are required, but I will make the update for theNote

jebeck:
Thanks again for this! Merging :)

PR #235, opened 2015-06-10
default to mg/dL for display of BG values (but allow change to mmol/L through query param)

jebeck:
Depends on [tideline 213](https://github.com/tidepool-org/tideline/pull/213). Once that's merged and tagged, will update tideline version in package.json here.

This makes the default BG display in mg/dL, with the option to switch to mmol/L by putting a query parameter `units=mmoll` before the hash (`#`).


jh-bate:
LGTM - couple of comments on the dependant tideline PR

PR #236, opened 2015-06-10
handle two crashing scenarios possible because of data + name mismatch fix

jebeck:
In testing the fix originally implemented in #230 @NSRiggall was getting some crashes occasionally when fetching data (reported [here](https://trello.com/c/HUPEouyS)). After talking with him and also experiencing similar myself on devel yesterday when tide-whisperer was timing out completely on data requests, I think the change in this pull request handles the issue by more gracefully recognizing when the server doesn't respond with data.

Still going to try to have a look in the logs to see if I can tell that this is the same issue as @NSRiggall was experiencing, but it may be difficult because of the weird stringification of blip errors that we haven't fixed yet in the logs...

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -908,6 +908,7 @@ var AppComponent = React.createClass({
         bgPrefs={this.state.bgPrefs}
         timePrefs={this.state.timePrefs}
         patientData={this.state.patientData}
+        fetchingPatient={this.state.fetchingPatient}
         fetchingPatientData={this.state.fetchingPatientData}
         isUserPatient={this.isSamePersonUserAndPatient()}
         queryParams={this.state.queryParams}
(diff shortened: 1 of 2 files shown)


jebeck:
Update: found the logs from the testing, and it looks like those were a related issue but not quite this one, so still working on this.

jebeck:
OK, so the two commits here handle two different possible crashing scenarios:
- first handles when the server doesn't succeed in finding data for a patient
- second (@NSRiggall's crash) handles when the request for the patient info comes back from the server *slower* than the patient data - this is pretty rare, but you can achieve it if you bounce around from patient to patient on the dashboard and get the app into a state where it has already loaded data for patient A from a previous request and when you navigate *back* to patient A it crashes trying to look up the data to render because  it doesn't have the profile info yet for patient A and so can't find the `userid` to use as the key to find the right data in the `patientData` object

jh-bate:
After discussing with @jebeck LGTM.

A strong candidate for some rethinking / rework based on the potential scenarios but this should work for now.

PR #237, opened 2015-06-11
Break down app

GordyD:
@jebeck @jh-bate  I've created the PR here into new-testing. I believe new-testing is a little behind origin/master and so some of the diffs are to do with that. 

Nevertheless you can see the two files:

app/app.js and app/components/app/app.js which show how the original one file has been split in 2!

Thoughts and feedback into approach here welcome.

You can also see the beginnings of testing of the app component.


jebeck:
@GordyD lemme know what you want to do with this. The diff is kind of impossible to read right now because it hasn't been updated with the recent changes on `new-testing`. Plus it's still on your old fork. So maybe close this and open a new one after creating a new branch on tidepool-org and locally merging from your old fork, then updating? IDK, that's probably how I would do it. Or cherry-pick commits if the relevant ones are small in numbers.

GordyD:
The huge difficulty with this branch is that because it is such a change to the structure of the app, and the contents of app.js keep changing on master, it is always tricky process handling merge conflicts which inevitably arise when app.js changes.

I'll move things over today to this repo and get things on track but I don't think we can focus on this until we get new-testing into master and this branch is then just based directly off master. What do you think?

jebeck:
Yeah, let's talk about this more tonight, but you're probably right, and if you haven't already moved things over and you want to wait until we're ready to focus on this, that could also make sense.

jebeck:
Superseded by and included in #242, so closing.

PR #240, opened 2015-06-26
timezone-aware bugfixes for notes

jebeck:
Fixes display and editing of timestamps on notes for new default to timezone-aware display.

Depends on [tideline #217](https://github.com/tidepool-org/tideline/pull/217).

PR #241, opened 2015-06-30
update build-related dependencies

jebeck:
Fun times. After I merged #240, I did a `git pull` on master to get the merged in update, then I cleaned out my node modules and did a fresh install just to do a final check before tagging. After the clean install, the `npm start` script was failing. The build (i.e., `npm run build`) was still working, but...not good to not be able to run locally.

We've been getting some autoprefixer warnings for a while, and I've been bitten before by the fact that even though *we* lock down our dependencies, if one of our deps doesn't lock down *its* dependencies, that can break things. (This happened to me once with Jest.)

So my theory is that's what happened. I went through and updated *all* of our webpack-related dependencies (with the help of [this tool](https://www.npmjs.com/package/npm-check-updates)), and now I can run both `npm start` and `npm run build` successfully locally.

@jh-bate if you could confirm on your local set up with a clean install of node modules and merge, that'd be great. Then I can tag & build the changes from #240 and get them back through devel and staging.

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,18 +14,18 @@
   },
   "dependencies": {
     "async": "0.9.0",
-    "autoprefixer-loader": "1.2.0",
+    "autoprefixer-loader": "2.0.0",
     "bows": "1.4.2",
     "connect": "3.3.5",
     "crossfilter": "1.3.11",
-    "css-loader": "0.9.1",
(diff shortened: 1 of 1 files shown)


jh-bate:
run locally and LGTM

PR #243, opened 2015-07-03
updates for bootstrapping to UTC

jebeck:
**WIP. Don't merge.**

Partly this is for your exposure to what's changing @GordyD (shouldn't be much).

This will also depend on a new tideline tag.


app/core/personutils.js
```diff
@@ -99,6 +99,9 @@ personUtils.isRemoveable = function(person) {
 personUtils.validateFormValues = function(formValues, isNameRequired, dateFormat, currentDateObj) {
   var validationErrors = {};
 
+  var INVALID_DATE_TEXT = 'Hmm, this date doesn’t look right';
```
jh-bate: not for now but you could expose these so they can be used in the tests also

jh-bate:
LGTM

app/core/personutils.js
jebeck: I think @GordyD already has tests around them in some fashion, so that might not be fully necessary, at least if the goal is to check what the text *is*.

PR #244, opened 2015-07-10
disable navigation buttons when in transition

GordyD:
@jebeck - I have created a helper method for rendering navigation buttons so that both the CSS class and the onClick function of the nav buttons are conditionally controlled by the `props.inTransition` value.


test/unit/components/chart/header.test.js
```diff
@@ -0,0 +1,295 @@
...
+      expect(props.onClickMostRecent.callCount).to.equal(0);
+    });
+
+    it('should trigger onClickNext when inTransition is false and next button is clicked', function () {
+      console.warn = sinon.stub();
+      var props = {
+        chartType: 'Awesome',
+        inTransition: false,
+        atMostRecent: false,
+        title: 'Most Awesome',
```
jh-bate: Nice!

test/unit/components/chart/header.test.js
```diff
@@ -0,0 +1,295 @@
...
+
+var React = require('react');
+var TestUtils = require('react/lib/ReactTestUtils');
+var _ = require('lodash');
+var expect = chai.expect;
+var rewire = require('rewire');
+var rewireModule = require('../../../utils/rewireModule');
+
+describe('Header', function () {
+  var Header = rewire('../../../../app/components/chart/header');
```
jh-bate: feels crazy but I guess thats what we have to do to access the header

app/components/chart/header.js
```diff
@@ -114,6 +122,33 @@ var TidelineHeader = React.createClass({
...
+   * @return {ReactElement}
+   */
+  renderNavButton: function(buttonClass, clickAction, icon) {
+    var nullAction = function(e) {
+      if (e) {
+        e.preventDefault();
+      }
+    };
+    if (this.props.inTransition) {
+      /* jshint ignore:start */
```
jh-bate: so with your other PR we can delete these can't we?

app/components/chart/header.js
```diff
@@ -39,28 +39,33 @@ var TidelineHeader = React.createClass({
...
     var weekLinkClass = cx({
+      'js-week': true,
       'patient-data-subnav-active': this.props.chartType === 'weekly',
       'patient-data-subnav-hidden': this.props.chartType === 'no-data'
     });
 
     var dateLinkClass = cx({
+      'js-date': true,
       'patient-data-subnav-text' : this.props.chartType === 'daily' || this.props.chartType === 'weekly' || this.props.chartType === 'modal',
       'patient-data-subnav-dates-daily': this.props.chartType === 'daily',
```
jh-bate: would it make sense to have `'daily'` as a const? Same with other strings for weekly etc

jh-bate:
@GordyD LGTM, a couple of small comments/ possible adjustments. I love the tests!

test/unit/components/chart/header.test.js
GordyD: Are you referring to rewire? It does seem quite heavy, but it really does help with isolation when running unit tests with this coding style.

app/components/chart/header.js
GordyD: Yea. Perhaps it makes sense to merge that branch last, after merging this and all others we can then remove all ignore flags. Thoughts?

jh-bate: yeah sounds like a plan

jebeck:
LGTM, merging.

PR #245, opened 2015-07-17
Update browser warning component + tests and remove unneccessary hand…

GordyD:
…lers and state from app


jh-bate:
code LGTM.

So the requirement was no longer to just warn - but force them to use chrome?

PR #246, opened 2015-07-24
Gordyd/jsxhint

GordyD:
Switch from jshint to jsxhint, so that we can remove the jshint ignore markers around use of jsx in our React components.

I have updated package.json so that Travis runs the new build step and also circle.yml so that it will mimic Travis if/when we move over.

One other point of note. I removed the call of `transferPropsTo` in message.js, as it is now deprecated. It  did not seem useful as it was not being used to transfer props to a child component. So instead of swapping it out with something like the props spread syntax in JSX I have just removed it altogether.


package.json
```diff
@@ -43,11 +42,7 @@
   "devDependencies": {
     "blip-mock-data": "tidepool-org/blip-mock-data#v0.3.7",
     "chai": "2.1.2",
-    "gulp": "3.8.11",
-    "gulp-jshint": "1.9.4",
-    "gulp-react": "3.0.1",
-    "jest-cli": "0.4.0",
-    "jshint-stylish": "1.0.1",
+    "jsxhint": "^0.15.1",
```
jh-bate: nit-pic - we are using strict versions everywhere so this should be  ``"jsxhint": "0.15.1"``

test/unit/components/messages/message.test.js
```diff
@@ -27,6 +27,7 @@ describe('Message', function () {
       expect(elem).to.be.ok;
 
       var initialState = elem.getInitialState();
+      console.log(initialState);
```
jh-bate: should this still be here?

jh-bate:
@GordyD LGTM - just a couple of notes above. What are your expectations on merging? I am keen to get some of these things moving and deployed

test/unit/components/messages/message.test.js
GordyD: Good spot!

GordyD:
I am unsure really. What is the usual process for managing merges? Jana has always handled the process from here on in so I am unfamiliar with how this is handled at Tidepool. :)

jebeck:
@GordyD if you could just de-duplicate and move the `...Chart` components (e.g., `DailyChart`) back to the bottom of their files (for all relevant things in `app/components/chart/`). And also lock down the versions on new dependencies (looks like the rewire-related things) since we don't use `^` as a rule.

circle.yml
```diff
@@ -1,5 +1,5 @@
 test:
   override:
-    - npm run jshint
+    - node_modules/jsxhint/cli.js 'app/**/*.js' 'mock/**/*.js' 'test/**/*.js' '*.js'
```
jebeck: Why did this have to change? The `npm run jshint` command still exists... Just curious.

GordyD: IIRC Circle CI was not running jsxhint when calling `npm run jshint` so I had to specify the command explicitly.

GordyD:
Going to close this as have opened a new pull request re-doing the messy merge. Here it is: https://github.com/tidepool-org/blip/pull/255

PR #247, opened 2015-07-24
Add css rule to allow selection of everything inside device settings …

GordyD:
…page


GordyD:
Add in ability to copy paste on login form and invitation screen (look into invitation screen)

jh-bate:
LGTM

PR #249, opened 2015-08-06
remove any references to imagesBaseUrl

GordyD:
remove all mentions of imageBaseUrl then check jshint and tests - all clear.


jh-bate:
@GordyD  LGTM. Happy to build and deploy whenever

PR #250, opened 2015-08-21
Add loading animation to blip

GordyD:
Loading screen has been updated to use new loading .gif.

Needs review from @jebeck 

One thing I am noticing is that the image is not pre-loaded, so sometimes when loading, the gif does not show as the race between loading the gif and loading the data happens and the data wins!

Is there a technique in React to preload this image? Or a way that has been done before in Blip?


jebeck:
LGTM.

PR #252, opened 2015-09-02
Gordyd/waitlist

GordyD:
WIP @jh-bate to complete back-end integration

See comments below.


app/components/app/app.js
```diff
@@ -422,11 +431,17 @@ var AppComponent = React.createClass({
   },
 
   renderSignup: function() {
+    var fakeCheckKey = function(key, cb) {
```
GordyD: This function for checking validity of an invite key will need replacing with something that checks a given key against an API service.

PR #255, opened 2015-09-03
Gordyd/jsxhint redux

GordyD:
New branch after quite nasty issues being experienced following merge from master.

Have jumped back to last known good commit and then re-merged and applied some further minor fixes.

Weirdly `jsxhint` picked up on usage of 'use strict' statement in some modules. So I removed them to get rid of the error. Doesn't seem to be used throughout our codebase so I thought it made sense to remove completely.

JSXhint and tests now pass. Versions in package.json are locked down.

De-duplication of Chart components is complete.


package.json
```diff
@@ -62,7 +57,7 @@
     "moment": "2.9.0",
     "phantomjs": "1.9.17",
     "react-tools": "0.13.2",
-    "rewire": "^2.3.4",
+    "rewire": "2.3.4",
     "rewire-webpack": "^1.0.0",
```
jebeck: Lock this version # down too plz?

package.json
```diff
@@ -71,15 +66,5 @@
...
-  },
-  "jest": {
-    "scriptPreprocessor": "app/preprocessor.js",
-    "unmockedModulePathPatterns": [
-      "node_modules/react",
-      "node_modules/sundial",
-      "node_modules/moment",
-      "node_modules/lodash",
-      "node_modules/blip-mock-data"
-    ]
```
jebeck: Do we still have the Jest tests sitting somewhere? If so, is now the time to remove them? (I do want them duplicated eventually in our non-Jest setup, so if we still want them around for reference, I guess that's cool. They should be in __tests__ dirs in messages and, well, the other ones are probably on #217.)

PR #256, opened 2015-09-15
Jhbate waitlist

jh-bate:
@jebeck @GordyD Addition of configuration for invite_key. 

Can be run localhost

``export INVITE_KEY='invite-me' ``

and then in the bowser

``http://localhost:3000/#/signup?inviteKey=invite-me``

and for localhost dev experience I have made it so if no config value is exported when blip is built then it wont check for one. This way we won't disrupt the ease of use when running localhost


jebeck:
LGTM, merging into Gordon's branch, then will merge that into master, tag & build for devel.

PR #259, opened 2015-09-29
Jhbate fix signup

jh-bate:
@jebeck If you could take an initial look at this please. FYI I want to do a full test on devel so I can test with the invite process / emails also.

Details in linked [trello card] (https://trello.com/c/5bFvHhPQ/74-bug-waitlist-when-a-new-user-is-invited-to-join-blip-via-an-existing-user-they-should-bypass-the-waitlist)

### DIFF
--- a/app/pages/signup/signup.js
+++ b/app/pages/signup/signup.js
@@ -60,15 +60,26 @@ var Signup = React.createClass({
     ];
   },
 
-  componentWillMount: function() {
-    var that = this;
-    if ((this.props.inviteKey || this.props.inviteKey === '' ) && this.props.checkInviteKey) {
-      this.props.checkInviteKey(this.props.inviteKey , function(valid) {
-        that.setState({loading: false, showWaitList: !valid });
+  isWaitListed:function(){
+
+    var hasInviteKey = !_.isEmpty(this.props.inviteKey) || this.props.inviteKey === '';
+    var hasInviteEmail = !_.isEmpty(this.props.inviteEmail);
+
+    if ( hasInviteKey && hasInviteEmail ) {
+      //don't show waitlist if invited user to create account and join careteam
+      return false;
+    } else if (hasInviteKey && this.props.checkInviteKey) {
+      //do we have a valid waitlist key?
+      return this.props.checkInviteKey(this.props.inviteKey, function(valid) {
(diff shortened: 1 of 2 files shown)


app/pages/signup/signup.js
```diff
@@ -60,15 +60,25 @@ var Signup = React.createClass({
...
   },
 
-  componentWillMount: function() {
-    var that = this;
-    if ((this.props.inviteKey || this.props.inviteKey === '' ) && this.props.checkInviteKey) {
-      this.props.checkInviteKey(this.props.inviteKey , function(valid) {
-        that.setState({loading: false, showWaitList: !valid });
+  isWaitListed:function(){
+
+    var hasInviteKey = !_.isEmpty(this.props.inviteKey) || this.props.inviteKey === '';
```
jebeck: Why does an empty string as an inviteKey pass through?

jh-bate: if you don't set the key in the URL param then you get and empty string. Then the check in checkInviteKey() that is passed in will see if there is a config value for the expected key - which might not be set so then in that case and empty string is legit.

It was a workaround so you could still develop localhost without having to pass through keys all the time.

test/unit/pages/signup.test.js
```diff
@@ -38,46 +38,61 @@ describe('Signup', function () {
...
         trackMetric: sinon.stub(),
         inviteKey: 'foobar'
       };
       var elem = React.createElement(Signup, props);
       var render = TestUtils.renderIntoDocument(elem);
-      var refreshButton = TestUtils.findRenderedDOMComponentWithClass(render, 'signup-form');
+      var signupForm = TestUtils.findRenderedDOMComponentWithClass(render, 'signup-form');
+    });
+
+    it('should render signup-form when both key and email are set', function () {
```
jebeck: Just to check my understanding, I changed this test to have `checkInviteKey: function(x, cb) { return cb(false); }` and it passed because only the e-mail is actually required to render the sign-up. So, either add another test showing this case or maybe amend the language of this test to make it clear that when the e-mail is set, it doesn't matter if the key is?

jh-bate: sure thing

PR #261, opened 2015-10-09
fix for basics logout bug

jebeck:
Prevent crash when trying to logout when logout precedes attempt to update patientData.

Plz review @GordyD and let's discuss tomorrow - this is an illustratively nasty bug not irrelevant to all the things we've been discussing.


GordyD:
LGTM!

PR #264, opened 2015-11-23
Displays react component (in device settings page) so that users can authorize Tidepool to reteive physical activity data from devices/apps like Jawbone, Runkeeper, etc

nosh:
If SHOW_DEVICE_AUTH_UI is not set to true, then the component does not display


jebeck:
OK, I am going to be the change I want to see in the world and finally close this outdated PR so I don't constantly have to filter it out when looking at open PRs.

PR #266, opened 2015-11-28
Gordyd/move to react router

GordyD:


app/routes.js
```diff
@@ -0,0 +1,173 @@
...
+export const requireAuth = (api) => (nextState, replaceState) => {
+  if (!api.user.isAuthenticated()) {
+    replaceState(null, '/login');
+  }
+};
+
+/**
+ * This function redirects any requests that land on pages that should only be
+ * visible when no data storage is set up if the user has data storage set up
+ *
```
jebeck: So @GordyD here's the first of the two functions I added to deal with similar issues with routes and the sign-up flow. The first issue was an infinite loop when navigating to /patients/new (via the address bar). This is only intended to be accessible if the user hasn't "set up data storage" (== fill out profile info including diagnosis date). So I basically ended up creating this additional `onEnter` function `requireAuthAndNoPatient` to redirect to /patients if the user already *has* set up data storage.

The big downside of this approach, obviously, is that it includes redundant API calls. Not a *huge* deal, but feels a little messy and like it shouldn't be necessary. I'm hoping once we move on to redux we'll have access to the global app state from here within routes.js and won't have to make these API calls? (The logic may have to stay until we do some further refactoring, but that feels less unacceptable.)

app/routes.js
```diff
@@ -0,0 +1,173 @@
...
+ * This function redirects any requests that land on pages that should only be
+ * visible when the user hasn't yet verified their sign-up e-mail
+ * if the user already has completed the e-mail verification
+ *
+ * @param  {Object} nextState
+ * @param  {Function} replaceState
+ *
+ * @return {boolean|null} returns true if hash mapping happened
+ */
+export const requireNotVerified = (api, cb) => (nextState, replaceState) => {
```
jebeck: @GordyD and here's the second one, because we weren't getting the proper flow through /email-verification

(This also required a backend change to return the `emailVerified` boolean with the user metadata.)

app/routes.js
```diff
@@ -0,0 +1,173 @@
...
+      }
+      return cb(err);
+    }
+    if (user.emailVerified === true) {
+      replaceState(null, '/patients');
+      return cb();
+    }
+    // we log the user out so that requireNoAuth will work properly
+    // when they try to log in
+    api.user.logout(() => {
```
jebeck: This is basically only relevant when:
- working with the entire stack locally using a `+skip` e-mail address to avoid ever doing verification
- and thus clicking Login directly from the /email-verification page

PR #270, opened 2016-01-13
Add test script to conform to test mechanism for other node projects

darinkrauss:
@jebeck @jh-bate Add simple test script to run npm test, similar to user-api-client testing (which also uses mocha to run tests)

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1,0 +1,3 @@
+#! /bin/bash -eu
+
+npm test


jh-bate:
LGTM

jebeck:
I don't understand why this is necessary? All of our client-side apps have the same standards for defining _all_ of our scripts in the package.json. Aside from `test` and `build`, we have about have a dozen other things defined and our CI configs use the `npm test` command.

PR #271, opened 2016-01-21
router replacement bugfixes

jebeck:
Two bugfixes here:

- JavaScript error was happening when trying to change password from /profile (turned out to be an issue in a small change made to the API wrapper)
- fix the fact that fonts wouldn't load when navigating directly to a nested URL (like /patients/:id/data) or refreshing on such

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,5 +1,6 @@
 var path = require('path');
 var webpack = require('webpack');
+var ExtractTextPlugin = require('extract-text-webpack-plugin');
 
 var entry = (process.env.MOCK === 'true') ? './app/main.mock.js' : './app/main.js';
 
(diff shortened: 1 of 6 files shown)


webpack.config.js
```diff
@@ -29,7 +30,7 @@ module.exports = {
       {test: /\.js$/, exclude: /(node_modules)/, loader: 'babel-loader'},
       // need this condition when Tideline loaded from Github branch
       {test: /node_modules\/tideline\/.*\.js$/, exclude: /tideline\/node_modules/, loader: 'babel-loader'},
-      {test: /\.less$/, loader: 'style-loader!css-loader!autoprefixer-loader!less-loader'},
+      {test: /\.less$/, loader: ExtractTextPlugin.extract('style-loader', 'css-loader!autoprefixer-loader!less-loader')},
```
GordyD: Nice!

webpack.config.js
```diff
@@ -43,7 +44,8 @@ module.exports = {
   },
   // tideline DEV env variable only needs to be true in tideline local dev
   plugins: [
-    defineEnvPlugin
+    defineEnvPlugin,
+    new ExtractTextPlugin('styles.css')
```
GordyD: Only a tiny nitpick with the name here. Why plural?

GordyD:
Other than css file name this LGTM and works on local when testing refreshing nested routes e.g. `http://localhost:3000/patients/4535a63c64/data`

Also password change whilst logged in on settings page works and gives user feedback. Able to log in with new password afterward.

Nice one!

PR #273, opened 2016-01-25
Tos

krystophv:
An effort to move the Terms of Use to it's own top level route.


app/components/app/app.js
```diff
@@ -319,21 +318,6 @@ export default class AppComponent extends React.Component {
       );
     }
 
-    if (!this.state.fetchingUser){
-      return this.renderTermsOverlay();
-    }
-
-    return null;
```
jebeck: Minor style nitpick: we usually keep an explicit `return null` when we only return stuff to render on certain `if` conditions. (Or at least that's the style I prefer - it's possible that we're not very consistent in blip... :/)

app/core/fetcher.js
```diff
@@ -33,7 +33,11 @@ export default class Fetcher {
       });
 
       //will show terms if not yet accepted
-      comp.renderOverlay = comp.renderTermsOverlay;
+      /*
+      if(!user.termsAccepted && comp.props.location.pathname !== '/terms'){
+        comp.props.history.pushState({originalPath: comp.props.location.pathname}, '/terms');
+      }
+      */
```
jebeck: Minor again, but delete your commented-out code?

app/pages/terms/index.js
```diff
@@ -13,4 +13,4 @@
  * not, you can obtain one from Tidepool Project at tidepool.org.
  */
 
-module.exports = require('./termsoverlay');
+module.exports = require('./terms');
```
jebeck: Awesome that you thought to remove all the "overlay" references!

jebeck:
I found one bug: if you're logged in (with Remember Me) and you _have_ already accepted the TOS, you'll get the age form again - this is actually due to a JS error getting thrown in the <Terms> `componentWillUpdate` - this would be the highest priority fix, IMO.

app/pages/terms/terms.js
```diff
@@ -100,18 +107,19 @@ var TermsOverlay = React.createClass({
...
-        </form>
+        {this.props.authenticated && !this.props.termsAccepted ?
+          <form className='terms-form'>
+            {agreeConfirmation}
+            {backBtn}
+            <button
+              className='terms-button terms-button-submit'
+              onClick={this.handleTermsAndPrivacySubmit}
+              disabled={continueBtnDisabled}>Continue</button>
+          </form> : '' }
```
jebeck: Style comment here - we like to create a var to store contents like this rather than have the ternary operator buried in the JSX where it's a bit hard to read.

app/pages/terms/terms.less
```diff
@@ -13,52 +13,52 @@
  * not, you can obtain one from Tidepool Project at tidepool.org.
  */
 
-@terms-overlay-zindex:        @zindex-overlay-background;
-@terms-overlay-text-zindex:   @zindex-overlay;
-@terms-overlay-box-width:     800px;
-@terms-overlay-age-box-width: 500px;
+@terms-zindex:        @zindex-overlay-background;
```
jebeck: We shouldn't need these z-indices anymore since it's not an overlay anymore, right?

app/routes.js
```diff
@@ -12,21 +12,32 @@ import PatientData from './pages/patientdata';
...
-export const requireAuth = (api) => (nextState, replaceState) => {
+export const requireAuth = (api) => (nextState, replaceState, cb) => {
   if (!api.user.isAuthenticated()) {
     replaceState(null, '/login');
+    return cb();
+  }
+  else {
+    api.user.get(function(err, user) {
+      if(!user.termsAccepted){
+        replaceState({originalPath: nextState.location.pathname}, '/terms');
```
jebeck: Nice touch with including the `originalPath`!

app/routes.js
```diff
@@ -167,6 +182,7 @@ export const getRoutes = (appContext) => {
     <Route path='/' component={AppComponent} {...props}>
       <IndexRoute components={{login:Login}} onEnter={onIndexRouteEnter(api)} />
       <Route path='login' components={{login:Login}} onEnter={requireNoAuth(api)} />
+      <Route path='terms' components={{terms:Terms}} />
```
jebeck: This struck me as odd at first, /terms being our only route w/o an `onEnter`, but after thinking it through, I guess it really is.

test/unit/pages/terms.test.js
```diff
@@ -61,9 +63,31 @@ describe('TermsOverlay', function () {
...
+      expect(termsElems.length).to.not.equal(0);
+    });
+    it('should NOT render age confirmation when user has acccepted terms and is logged in', () => {
+      var acceptDate = new Date().toISOString();
+      var props = { authenticated: true, termsAccepted: acceptDate, fetchingUser: false }
+      var termsElem = React.createElement(Terms, props);
+      var elem = TestUtils.renderIntoDocument(termsElem);
+
+      var termsElems = TestUtils.scryRenderedDOMComponentsWithClass(elem, 'terms-age-form');
+      expect(termsElems.length).to.equal(0);
```
jebeck: What about checking that the `terms-form` is also not present in this case?

test/unit/pages/terms.test.js
```diff
@@ -7,18 +7,20 @@
 var React = require('react');
 var TestUtils = require('react-addons-test-utils');
 var expect = chai.expect;
+var api = require('../../../app/core/api');
+var mock = require('../../../mock');
```
jebeck: You probably just added this because you thought you were gonna need it, but doesn't look like you're using it. Remove? (Minor, obviously.)

test/unit/routes.test.js
```diff
@@ -17,10 +17,21 @@ var expect = chai.expect;
 describe('routes', () => {
 
   describe('requireAuth', () => {
-    it('should update route to /login if user is not authenticated', () => {
+    it('should update route to /login if user is not authenticated', (done) => {
       let api = {
         user : {
-          isAuthenticated: sinon.stub().returns(false)
+          isAuthenticated: sinon.stub().returns(false),
+          get: (cb) => {
```
jebeck: You shouldn't need to mock the `get` in this test case, right?

test/unit/routes.test.js
```diff
@@ -44,9 +69,41 @@ describe('routes', () => {
...
+          isAuthenticated: sinon.stub().returns(true),
+          get: (cb) => {
+            cb(
+              null,
+              {
+                userid: 'a1b2c3',
+                profile: {
+                  patient: {}
+                },
+                termsAccepted: false
```
jebeck: Before terms are accepted the field comes back as `''` (empty string) - should be updated in these test fixtures ideally.

test/unit/routes.test.js
```diff
@@ -207,6 +265,34 @@ describe('routes', () => {
...
+      let api = {
+        user: {
+          //isAuthenticated: sinon.stub().returns(true),
+          get: (cb) => {
+            cb(
+              null,
+              {
+                userid: 'a1b2c3',
+                emailVerified: true,
+                termsAccepted: null
```
jebeck: Same comment as above re: `''`.

jebeck:
Just looked at your fixes, all LGTM! Not merging quite yet, I need to discuss with @GordyD about how this fits in with his ongoing work to migrate blip to redux for state management (the next step after our move to React Router). Some of this may conflict with parts of his work, and that's my poor planning to have not thought of it before (oops), but I think we'll find a way to merge and keep the changes we can - probably into his branch.

I'm still noodling on the integration tests myself, will let you know if I figure out what's going on there.

And thanks! 😎

jebeck:
#286 replaces this, so closing.

PR #275, opened 2016-01-27
Fix tidepool uploader URL

cozmo:
When first signing up for blip you're most likely presented with the following screen - 

Sadly the giant "Upload Data" link doesn't work. It takes you to http://tidepool.org/uploader?token=REDACTED. If you remove the token from this URL (doesn't appear to be needed anymore anyways) the above url will redirect to the correct destination - http://tidepool.org/products/tidepool-uploader/. If you don't you're stuck on the blog's 404 page which is not where you want to be. 

Three possible fixes
1. Update the redirector to support the old style links even if they have query params. This should probably happen but after a quick look at your github I can't find the code that powers the marketing website. 
2. Don't include the token and allow the redirector to do it's job - This seems like a hack
3. Correct the original url. That is what this PR does, although it appears as though you may need to update some env vars which aren't stored in Github if they're overwriting these values. 

Let me know if there's anything I can add to this PR, and also let me know if I just missed the marketing website code, I'd happily update that as well.

### DIFF
--- a/config.app.js
+++ b/config.app.js
@@ -41,7 +41,7 @@ module.exports = {
   VERSION: pkg.version,
   MOCK: booleanFromText(__MOCK__, false),
   MOCK_PARAMS: __MOCK_PARAMS__ || '',
-  UPLOAD_API: __UPLOAD_API__ || 'https://tidepool.org/uploader',
+  UPLOAD_API: __UPLOAD_API__ || 'https://tidepool.org/products/tidepool-uploader',
   API_HOST: __API_HOST__ || 'https://dev-api.tidepool.org',
   SHOW_ACCEPT_TERMS: booleanFromText(__SHOW_ACCEPT_TERMS__, true),
   PASSWORD_MIN_LENGTH: integerFromText(__PASSWORD_MIN_LENGTH__, 8),


darinkrauss:
@cosmo Thanks for looking into this issue. 

As it turns out, the root problem was an incorrectly configured redirect on the Tidepool website. It was intended to redirect from http://tidepool.org/uploader to http://tidepool.org/products/tidepool-uploader regardless of the URL query parameters, but did not. I updated the website configuration and now this should be working again. Check it out and let me know if you run into any difficulties.

The reason why we don't want to update the URL in Blip is because the Tidepool Uploader Chrome extension registers two url_handlers using the http://tidepool.org/uploader URL when installed. That way, if the uploader is installed, navigating to the URL (such as what Blip does) will automatically launch the uploader. However, if the uploader is NOT installed, navigating to the URL will go to the Tidepool website and be redirected to the correct page with instructions on how to install the Tidepool uploader. Hopefully that makes sense.

With that said, I'm going to close this PR. If you feel it is in error, please reopen the PR and ping me and I'll look into it again.  Thanks!

cozmo:
@darinkrauss Makes a lot of sense - I didn't realize the 2 upload url trick as I haven't dug much into the chrome extension code yet (that's next). Thanks for the details. 

Just to confirm - is the website code not on github? If it is and I just missed it could you direct me to it? Thanks!

darinkrauss:
@cozmo The key bit for the url_handlers in the uploader is https://github.com/tidepool-org/chrome-uploader/blob/master/manifest.json#L15.

Right now the new website repo is private, but we are looking into making it public. 

However, the website is pretty straightforward. It is a WordPress site and uses the [Quick Page/Post Redirect Plugin](https://wordpress.org/plugins/quick-pagepost-redirect-plugin/) to manage redirects.  The main problem that I fixed was that the redirects were configured with absolute URLs rather than relative, as the plugin documentation recommends.  Once I made that switch the query parameters were correctly ignored and everything worked correctly.

jebeck:
And I just wanted to add since you mentioned looking into the uploader @cozmo, both blip here and the uploader are in the midst of large refactorings - migrating to [redux](http://redux.js.org/) for state management. That means contributions may be difficult to work in if they conflict with changes in flight. If you find something you'd like to work on, it would nice if you could open an issue to give us a heads up and start discussion. Thanks!

cozmo:
Thanks @jebeck - Not planning to make any big changes, more just trying to learn the general lay of the land (and fix any bugs like this I find). I'll make sure I don't step on any toes :)

PR #276, opened 2016-02-09
fix grammar in new authorization error message

jebeck:
I managed to produce this error from the uploader (by clicking Go to Blip for a user that the user I had logged into blip w/Remember Me did not have access to). I noticed a grammatical mistake in the error message ('your' instead of 'you're') and fixed it here.

CC @darinkrauss @jh-bate


darinkrauss:
:+1:

PR #279, opened 2016-02-18
Move processing of patient data inside PatientData component

GordyD:


app/core/utils.js
```diff
@@ -181,12 +181,12 @@ utils.getInviteKey = function(location) {
...
 }
 
-utils.processPatientData = (data, queryParams, timePrefs, bgPrefs) => {
+utils.processPatientData = (comp, data, queryParams, timePrefs, bgPrefs) => {
   if (!(data && data.length >= 0)) {
     return null;
   }
 
-  var mostRecentUpload = _.sortBy(_.where(data, {type: 'upload'}), (d) => Date.parse(d.time) ).reverse()[0];
+  var mostRecentUpload = _.sortBy(_.filter(data, {type: 'upload'}), (d) => Date.parse(d.time) ).reverse()[0];
```
GordyD: Have swapped usage of _.where for _.filter as requested

app/components/app/app.js
```diff
@@ -367,7 +367,7 @@ let mapDispatchToProps = dispatch => bindActionCreators({
 
 let mergeProps = (stateProps, dispatchProps, ownProps) => {
   var api = ownProps.routes[0].api;
-  return _.merge({}, ownProps, stateProps, dispatchProps, {
+  return Object.assign({}, ownProps, stateProps, dispatchProps, {
```
GordyD: All uses of _.merge in mergeProps() functions used when connecting redux state to components has been removed.

app/pages/patientdata/patientdata.js
```diff
@@ -504,13 +485,20 @@ export let PatientData = React.createClass({
 
     var refresh = this.props.onRefresh;
     if (refresh) {
-      this.setState({title: this.DEFAULT_TITLE});
-      refresh();
+      this.props.clearPatientData(this.props.routeParams.id);
```
GordyD: Clearing patient data here, allows refresh to result in processing post data being passed in as props as the conditions of componentWillReceiveProps will pass.

app/pages/patientdata/patientdata.js
```diff
@@ -526,6 +514,62 @@ export let PatientData = React.createClass({
...
+    
+    var nextPatientData = _.get(nextProps, ['patientDataMap', userId], null);
+
+    if (!currentPatientData && nextPatientData) {
+      this.doProcessing(nextProps);
+    }
+  },
+
+  doProcessing: function(nextProps) {
+    // I realised the one source of truth we can rely on here to ensure we show the correct patient
```
GordyD: Important comment here regarding a single source of truth for the correct patient id that should be in view. We can use the route. Would love feedback here.

GordyD:
@jebeck Worth checking out when you have a chance tomorrow, some more subtle but significant changes made that I have commented on for you to look at.

app/pages/patientdata/patientdata.js
```diff
@@ -14,6 +14,8 @@
  * not, you can obtain one from Tidepool Project at tidepool.org.
  */
 
+var count= 0;
```
jebeck: This was just for logging, right? Want to take it out now?

GordyD: Good spot!

app/components/app/app.js
jebeck: :+1:

app/core/utils.js
```diff
@@ -209,10 +209,9 @@ utils.processPatientData = (data, queryParams, timePrefs, bgPrefs) => {
   }
   // but otherwise we use the timezone from the most recent upload metadata obj
   else {
-    // @todo: need to figure out a way to refactor this bit of shenanigans
-    // comp.setState({
-    //   timePrefs: timePrefsForTideline
-    // });
```
jebeck: Since you removed the TODO, no longer thinking about refactoring? I don't think I mind if you don't - since this preprocessing step is one of the things on the chopping block the hackery is not _too_ bad.

app/pages/patientdata/patientdata.js
```diff
@@ -124,7 +114,7 @@ export let PatientData = React.createClass({
   },
 
   renderPatientData: function() {
-    if (this.props.fetchingPatient || this.props.fetchingPatientData) {
+    if (this.props.fetchingPatient || this.props.fetchingPatientData || this.state.processingData) {
```
jebeck: Ah, so I see you won the bet on this. The GIF still freezes when you hit the data processing step. I guess the only thing that could change that is putting the data processing in a Web Worker, which is on the roadmap at least for investigation for future-tideline.

app/pages/patientdata/patientdata.js
jebeck: Hmm, just tried and I've got the GIF looping indefinitely. Will try to read through and see what I can see.

app/pages/patientdata/patientdata.js
```diff
@@ -504,13 +485,20 @@ export let PatientData = React.createClass({
...
+      this.setState({ 
+        title: this.DEFAULT_TITLE, 
+        processingData: true,
+        processedPatientData: null 
+      });
+      refresh(this.props.routeParams.id);
     }
   },
 
   updateBasicsData: function(userid, data) {
```
jebeck: This doesn't need userid anymore, so remove?

app/pages/patientdata/patientdata.js
```diff
@@ -567,19 +612,18 @@ let mapStateToProps = state => ({
...
   closeMessageThread: actions.sync.closeMessageThread,
 }, dispatch);
 
 let mergeProps = (stateProps, dispatchProps, ownProps) => {
   var api = ownProps.routes[0].api;
-  return _.merge({}, ownProps, stateProps, dispatchProps, {
+  return Object.assign({}, ownProps, stateProps, dispatchProps, {
     fetchers: getFetchers(dispatchProps, ownProps, api),
     uploadUrl: api.getUploadUrl(),
     onRefresh: dispatchProps.fetchPatientData.bind(null, api),
```
jebeck: Maybe this is behind the refresh not working? `onRefresh` is still calling the old `fetchPatientData` with the embedded data processing if I'm reading this right. Don't you want to map it to the async action creator `fetchPatientData` from the `fetchers`?

app/pages/patientdata/patientdata.js
jebeck: I think I quite like this idea. Will want to ruminate on it and discuss more as more of the state reorg happens - basically will want to make sure that the route is then the single source of truth for 'patient in view' full stop (app-wide), but I think that makes a lot of sense and puts the router in control in a way that feels quite right to me, actually.

app/pages/patientdata/patientdata.js
```diff
@@ -504,13 +485,20 @@ export let PatientData = React.createClass({
 
     var refresh = this.props.onRefresh;
     if (refresh) {
-      this.setState({title: this.DEFAULT_TITLE});
-      refresh();
+      this.props.clearPatientData(this.props.routeParams.id);
+      this.setState({ 
+        title: this.DEFAULT_TITLE, 
+        processingData: true,
+        processedPatientData: null
```
jebeck: Should set the other things we are deriving after processing to `null` too - `bgPrefs` &c.

app/core/utils.js
GordyD: Yea I def want to refactor this. Could tackle shortly. Do not think it is too difficult to tidy up.

app/pages/patientdata/patientdata.js
GordyD: Ok

app/pages/patientdata/patientdata.js
```diff
@@ -526,6 +512,61 @@ export let PatientData = React.createClass({
       datetimeLocation: datetime
     });
   },
+
+  componentWillMount: function() {
+    this.doFetching(this.props);
+    this.doProcessing(this.props);
```
jebeck: Still want to take the `doProcessing` out of here and leave it only in `componentWillReceiveProps` right? It is getting called 8-9 times now for each patient view.

app/pages/patientdata/patientdata.js
```diff
@@ -526,6 +512,61 @@ export let PatientData = React.createClass({
...
+    
+    if (!currentPatientData && nextPatientData) {
+      this.doProcessing(nextProps);
+    }
+  },
+
+  doProcessing: function(nextProps) {
+    // I realised the one source of truth we can rely on here to ensure we show the correct patient
+    // is to refer to the userId specified in the route! What do you think @jana?
+    var userId = _.get(this.props, 'routeParams.id', null);
```
jebeck: Now that you've mapped this is `currentPatientInViewId` use that? And then the next bit can also be reduced to something like `_.get(nextProps, ['patientDataMap', this.props.currentPatientInViewId || ''], null)`...

app/pages/patientdata/patientdata.js
GordyD: this has been taken care of now :)

app/pages/patientdata/patientdata.js
GordyD: I recall yesterday we took it out and there was a specific case for why it needed to be present on componentWillMount.

GordyD: Just cannot recall exactly why!

app/pages/patientdata/patientdata.js
GordyD: ok!

app/pages/patientdata/patientdata.js
jebeck: I think it was for the refresh, but you solved that in another way, so I think we can delete? If it doesn't break on delete, that is.

jebeck:
Just the comment about deleting the call to `doProcessing` from `componentWillMount` if it's really not doing anything (which I think it isn't). Otherwise, looks good to merge.

PR #280, opened 2016-02-20
Redux on enter refactor

GordyD:


app/routes.js
```diff
@@ -25,8 +25,10 @@ import personUtils from './core/personutils';
  *
  * @return {boolean|null} returns true if hash mapping happened
  */
-export const requireAuth = (api) => (nextState, replace) => {
-  if (!api.user.isAuthenticated()) {
+export const requireAuth = (api, store) => (nextState, replace) => {
+  let { blip: state } = store.getState();
+
+  if (!state.isLoggedIn && !api.user.isAuthenticated()) {
```
jebeck: Since `api.user.isAuthenticated` is a sync function, querying the state in this particular `onEnter` hook is not exactly necessary, right? I _think_ I'd prefer to take it out then, for simplicity, but I can probably be persuaded otherwise if I understand your reasoning.

app/routes.js
```diff
@@ -40,19 +42,26 @@ export const requireAuth = (api) => (nextState, replace) => {
...
     replace('/login');
     return cb();
-  }
-  else {
-    api.user.get(function(err, user) {
-      if (personUtils.isPatient(user)) {
+  } else {
+    if (state.loggedInUser) {
+      if (personUtils.isPatient(state.loggedInUser)) {
```
jebeck: Is `state.loggedInUser` currently a full user object? In order for this to work, I think it must be, but eventually in the normalized state we'll want `loggedInUser` to be just a `userid`. So I imagine in the future this will have to be changed to something like `if (personUtils.isPatient(state.allUsers[state.loggedInUser])`...

app/routes.js
```diff
@@ -65,8 +74,10 @@ export const requireAuthAndNoPatient = (api) => (nextState, replace, cb) => {
  *
  * @return {boolean|null} returns true if hash mapping happened
  */
-export const requireNoAuth = (api) => (nextState, replace) => {
-  if (api.user.isAuthenticated()) {
+export const requireNoAuth = (api, store) => (nextState, replace) => {
+  let { blip: state } = store.getState();
+
+  if (state.isLoggedIn || api.user.isAuthenticated()) {
```
jebeck: Same comment as I made in `requireAuth`. Also it feels dangerous here to use `state.isLoggedIn || api.user.isAuthenticated()`...surely we want the `api.user.isAuthenticated()` to be the source of truth, and not allow some potential weird state mismatch to make it look as if you're logged in?

app/routes.js
```diff
@@ -81,18 +92,27 @@ export const requireNoAuth = (api) => (nextState, replace) => {
...
-export const requireNotVerified = (api) => (nextState, replace, cb) => {
-  api.user.get(function(err, user) {
-    if (err) {
-      // we expect a 401 Unauthorized when navigating to /email-verification
-      // when not logged in (e.g., in a new tab after initial sign-up)
-      if (err.status === 401) {
+export const requireNotVerified = (api, store) => (nextState, replace, cb) => {
+  let { blip: state } = store.getState();
+  if (state.loggedInUser) { 
+    checkIfVerified(state.loggedInUser);
```
jebeck: Same comment re: eventual normalized state and `loggedInUser` presumably shrinking to just a `userid`.

app/routes.js
```diff
@@ -81,18 +92,27 @@ export const requireNoAuth = (api) => (nextState, replace) => {
...
       }
-      throw new Error('Error getting user at /email-verification');
-      return cb();
-    }
-    if (user.emailVerified === true) {
+      checkIfVerified(user, cb);
+    });
+  }
+
+  function checkIfVerified(userToCheck) {
```
jebeck: Can you move this function to the top of `requireNotVerified` for better readability/consistency in fn declarations?

app/routes.js
```diff
@@ -115,8 +135,10 @@ export const requireNotVerified = (api) => (nextState, replace, cb) => {
  *
  * @return {boolean|null} returns true if hash mapping happened
  */
-export const onUploaderPasswordReset = (api) => (nextState, replace) => {
-  if (api.user.isAuthenticated()) {
+export const onUploaderPasswordReset = (api, store) => (nextState, replace) => {
+  let { blip: state } = store.getState();
+
+  if (state.isLoggedIn || api.user.isAuthenticated()) {
```
jebeck: Ditto comment on `requireNoAuth`.

app/routes.js
GordyD: Sounds good. I'll remove it in this and any functions that have similar behaviour that you have commented on.

app/routes.js
GordyD: Yea, full user currently. In future we'll have to refactor all uses of `loggedInUser` so I think it is ok to leave this as is when we merge this branch, and then deal with normalisation task as a separate thing.

jebeck: Agreed.

PR #281, opened 2016-03-01
Gordyd/redux

GordyD:
WIP - do not consider for merge just yet.


jebeck:
I think I'm going to just add comments as comments and not try to comment in-line on the diff, which is making GitHub topple over...

First comment: in the state tree we have `timePrefs` and `bgPrefs` at the top level without mapping them via `userId`. That's fine for now, but in the near future when those are user-configurable it _may_ need to change (depends on if the logged-in user's prefs is used for viewing all data, not just their own). So just raising a point about near-future work. Do you agree @GordyD that we should keep it as-is for now?

jebeck:
There seems to be a typo in the constants - `CLEAR_PATIENT__IN_VIEW` (with two underscores).

jebeck:
In the `allUsersMap` reducer it looks like `FETCH_PATIENTS_SUCCESS` is being handled twice?

jebeck:
A couple more things from looking at reducers:

- maybe consider adding handling of FETCH_PATIENT_DATA_FAILURE in the patientDataMap and patientNotesMap reducers as an extra measure of security - it should clear the map. Just thinking about the driving force behind this refactor with the patient data mismatch.
- get rid of the redundant loggedInUser reducer (replaced by loggedInUserId, no?) - although there are action types being handled by the former that aren’t being handled by the latter?

jebeck:
In the misc reducers, I think you should replace all handling of `LOGOUT_SUCCESS` with `LOGOUT_REQUEST` - we don't wait for the API to respond when logging out, we destroy the user's session right away and the state should reflect that.

jebeck:
Question: why is the `sentEmailVerification` reducer responding to `SIGNUP_SUCCESS` (which we are talking about reacting differently to in the reducers re: the `signupConfirmed` reducer and the "welcome" UI on /patients?

jebeck:
`REMOVE_PATIENT_SUCCESS` is only be handled in `membershipPermissionsInOtherCareTeams` but not in `memberInOtherCareTeams`...shouldn't it be handled in both?

jebeck:
Ditto but opposite for `ACCEPT_RECEIVED_INVITE_SUCCESS`...is only handled in `memberInOtherCareTeams` but should be handled in `membershipPermissionsInOtherCareTeams` too, no?

jebeck:
Are the reducers that handled `UPDATE_USER_REQUEST` doing so for optimistic updating?

jebeck:
Is the tideline state persistence broken for The Basics? I ask because it looks like the `UPDATE_PATIENT_DATA_`* action types don’t appear in any reducers...

jebeck:
Reviewed extensively in the code by auditing tests for falsifiability (and often updating/fixing them) plus fixing a lot of remaining bugs. Merging \o/

CC @GordyD

PR #282, opened 2016-03-04
redux bugfix timePrefs

jebeck:
Fixes the issues with setting of timePrefs and removes both bgPrefs and timePrefs from global state - probably you'll want to double-check that I got everything in those deletions @GordyD


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;DUTPf8bw&#x2F;27-blip-document-and-or-fix-regressions-from-redux-work"> [blip] document and&#x2F;or fix regressions from redux work</a>

jebeck:
@GordyD has verbally approved this merge (but is having trouble on GitHub right now), merging.

PR #284, opened 2016-03-04
test mapStateToProps for app.js

jebeck:
What you had in mind @GordyD?

I think for each smart component may have to decide how many batches of tests to do. Here it made sense to do one batch on the initial (not logged in) state and one batch on the post-logged-in state since the mappings are a bit different.


GordyD:
LGTM!

PR #285, opened 2016-03-11
finsh mapStateToProps tests

jebeck:
Also a small bugfix. @GordyD maybe just give this a quick glance and merge? I rebased on top of the new gordyd/redux.


GordyD:
Minor nitpick, there is no test for pages/app.js. But other than that looks good to me.

jebeck:
Oh fudge, I had a thing in my stash for app.js but I assumed it was stale because I'd already done it. Let's see if I can recover it from my REFLOG...

jebeck:
Will be adding back the app.js mapStateToProps tests and doing ones for terms page after merge of #286, so merging this now.

PR #289, opened 2016-03-16
Remove waterfall logic for welcome screen and reinstate redirect when…

GordyD:
… logging in with one patient

The state logic for showing the welcome screen to users was confusing, and only triggered when a user had just signed up (confirmed signup), this has now been simplified to just be triggered any time they log in if they have no data and no access to patients and no pending invites.

The reinstated redirect for accounts with access to exactly one patient has been put back in to improve the UX

### DIFF
--- a/app/pages/patients/patients.js
+++ b/app/pages/patients/patients.js
@@ -36,6 +36,7 @@ export let Patients = React.createClass({
     user: React.PropTypes.object,
     patients: React.PropTypes.array,
     invites: React.PropTypes.array,
+    loading: React.PropTypes.bool,
     showingWelcomeMessage: React.PropTypes.bool,
     onHideWelcomeSetup: React.PropTypes.func,
     trackMetric: React.PropTypes.func.isRequired,
--- a/app/pages/patients/patients.js
+++ b/app/pages/patients/patients.js
@@ -15,7 +15,7 @@
  */
 
 import React from 'react';
-import { Link } from 'react-router';
+import { browserHistory, Link } from 'react-router';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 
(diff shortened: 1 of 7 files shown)


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;NxVw0mMy&#x2F;21-blip-redux-devs-internal-bug-hunt-and-final-review"> [blip] [redux] devs-internal bug hunt and final review</a>

app/pages/patients/patients.js
```diff
@@ -34,13 +34,9 @@ import Invitation from '../../components/invitation';
 export let Patients = React.createClass({
   propTypes: {
     user: React.PropTypes.object,
-    fetchingUser: React.PropTypes.bool,
     patients: React.PropTypes.array,
-    fetchingPatients: React.PropTypes.bool,
     invites: React.PropTypes.array,
-    fetchingInvites: React.PropTypes.bool,
```
jebeck: Oh, now I see what the question about the super-destructuring was about! I like this. Though isn't `loading` now a prop? Add it to the propTypes please?

app/pages/patients/patients.js
```diff
@@ -53,7 +49,7 @@ export let Patients = React.createClass({
   render: function() {
     var welcomeTitle = this.renderWelcomeTitle();
 
-    if (this.isLoading() && !(this.props.showingWelcomeTitle && this.props.showingWelcomeSetup)) {
+    if (this.isLoading() && !(this.props.showingWelcomeMessage)) {
```
jebeck: I had added this second condition and now I think we want to remove it... We _do_ want the "loading" message to render while waiting if we're showing the welcome setup.

app/pages/patients/patients.js
```diff
@@ -98,7 +94,7 @@ export let Patients = React.createClass({
     var handleClickYes = function(e) {
       e.preventDefault();
       self.props.onHideWelcomeSetup();
```
jebeck: I don't think we want to trigger `onHideWelcomeSetup` on clicking yes, right? Because if I get distracted and close the tab, then come back I won't get the welcome setup. OTOH, if I complete the welcome setup, my patients will now be > 0 so it won't matter that I have `showingWelcomeMessage` true still.

GordyD: Not sure I follow this. I'd have thought you'd def want to clear the message here.

GordyD: If you get distracted and close the tab, nothing happens anyway -> you have not clicked yes.

GordyD: But yea ok, if you do click yes and then close tab (and have remember me clicked) you won't see the welcome message again, but I'd argue thats your fault - you acknowledged the message, and we do have the fall back message anyway that gives affordance on what to do without anything set up.

app/pages/patients/patients.js
GordyD: done

app/pages/patients/patients.js
GordyD: done

app/pages/patients/patients.js
jebeck: Getting Howard to tie-break here ;)

PR #293, opened 2016-03-17
Add in logout handler and deal with bug around welcome message when l…

GordyD:
…ogging out from patient screen


app/pages/patients/patients.js
```diff
@@ -310,8 +311,9 @@ export let Patients = React.createClass({
   },
 
   componentWillReceiveProps: function(nextProps) {
-    let { loading, patients, invites, location, showingWelcomeMessage } = nextProps;
-    if (!loading && location.query.justLoggedIn) {
+    let { loading, loggedInUserId, patients, invites, location, showingWelcomeMessage } = nextProps;
+    
+    if (!loading && loggedInUserId && location.query.justLoggedIn) {
```
jebeck: So you had to add `loggedInUserId` here to catch when you log out and we clear the state then try to redirect you, is that it?

PR #297, opened 2016-03-18
PropType testing for pages

GordyD:
After wrestling with an issue inside React that suppresses repeat missing propType warnings see https://github.com/facebook/react/issues/6293 Jana and I have concluded that we can test the propType validation effectively as long as we first test the rendering of the component with all props specified and then test for warnings when no props specified, second.


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;NxVw0mMy&#x2F;21-blip-redux-devs-internal-bug-hunt-and-final-review"> [blip] [redux] devs-internal bug hunt and final review</a>

PR #299, opened 2016-04-01
rev tideline version to v0.4.10 for active basal sched display

jebeck:
What it says on the tin. Sign-off @jh-bate?


jh-bate:
LGTM

PR #303, opened 2016-04-07
Account for patient being other person

GordyD:
When displaying a list of patients whose data you are able to view we need to ensure the list is sorted by the patient's name rather than the name associated with the profile that manages the patient. This commit updates the sorting rules and adds a new test condition to ensure that this property is used when the patient isOtherPerson flag is set to true.

### DIFF
--- a/app/components/peoplelist/peoplelist.js
+++ b/app/components/peoplelist/peoplelist.js
@@ -48,7 +48,8 @@ var PeopleList = React.createClass({
 
       // first sort by fullName
       var sortedPeople = _.sortBy(this.props.people, function(person) {
-        return person.profile.fullName;
+        var patient = person.profile.patient;
+        return (patient && patient.isOtherPerson && patient.fullName) ? person.profile.patient.fullName : person.profile.fullName;
       });
 
       // then pop the logged-in user to the top if has data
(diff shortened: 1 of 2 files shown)


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;6U8jVp6l&#x2F;10-blip-make-sorting-of-names-in-patient-selector-more-self-evident"> [blip] Make sorting of names in patient selector more self-evident</a>

app/components/peoplelist/peoplelist.js
```diff
@@ -48,7 +48,8 @@ var PeopleList = React.createClass({
 
       // first sort by fullName
       var sortedPeople = _.sortBy(this.props.people, function(person) {
-        return person.profile.fullName;
+        var patient = person.profile.patient;
+        return (patient && patient.isOtherPerson && patient.fullName) ? person.profile.patient.fullName : person.profile.fullName;
```
jebeck: Let's be a bit more careful about our property accessing here by using `_.get` perhaps. We're going to be updating to real child accounts soon in conjunction with some backend changes, so I think it'll be worth it...

jebeck:
LGTM

PR #304, opened 2016-04-07
Update redux-mock-store to new version

GordyD:
- Update API usage of `redux-mock-store` due to breaking changes
 - Fix buggy tests that were not failing due to being caught out by old interface
 - Address suppressed bug in showWelcomeMessage.test.js where state was declared twice


jebeck:
Spot checked tests to make sure they fail under circumstances where they should. :+1:

PR #305, opened 2016-04-08
Naming updates around care team memberhships

GordyD:
Following a discussion with Jana we felt it was a good idea to rename some action creators to improve the consistency of naming.

In this commit the naming change has been applied to:
 - actionCreators
 - constants
 - a reducer for membershipInOtherCareTeams
 - tests
 - all usages within Blip

I've also tested setting up a new account, removing members from target care team and removing membership in other care teams on local.


jebeck:
:+1:

PR #306, opened 2016-04-08
Add onEnter rule for ensuring user is in logged out state and write t…

GordyD:
…est to verify behaviour of function


app/routes.js
```diff
@@ -94,13 +93,21 @@ export const requireAuthAndNoPatient = (api, store) => (nextState, replace, cb)
 };
 
 /**
+ * This function ensures any logged in state is destroyed on entering a route
+ * 
+ * @param  {Object} nextState
+ * @param  {Function} replace
+ */
+export const ensureNoAuth = (api) => (nextState, replace) => {
+  api.user.logout();
```
jebeck: So we're relying on the fact that `api.user.logout` is synchronous here (but does take an optional cb). I believe elsewhere in this file we're passing the cb...shall we align one way or the other, just for reducing confusion?

test/unit/routes.test.js
```diff
@@ -385,6 +386,20 @@ describe('routes', () => {
   });
 
   describe('requireNoAuth', () => {
```
jebeck: Update the `describe` to the right method name plz

app/routes.js
GordyD: ok sure thing.

jebeck:
Looks like the ESLint monster is demanding tribute :P

jebeck:
LGTM, but we are waiting for initial release of blip w/redux to merge to master.

jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;LqIxxZk7&#x2F;26-blip-bug-when-user-a-is-logged-into-blip-with-remember-me-checked-and-user-b-clicks-reset-password-from-an-email-in-the-same-bro"> [blip] [bug] When User A is logged into Blip with remember me checked, and User B clicks &quot;reset password&quot; from an email in the same browser window (different tab), User B should not be redirected to User A&#39;s account.</a>

PR #309, opened 2016-04-11
remove a metric that will never actually get recorded

jebeck:
This metric will also be fire before login, but we can't push metrics until after login. This PR therefore removes the metric for greatest code readability.


krystophv:
LGTM :+1:

PR #316, opened 2016-06-10
Fix typos in errorMessages.js

PeterSampson:
I agree to the terms of Tidepool Project’s Volunteer/Contributor License Agreement v1.0 as it exists at http://tidepool-org.github.io/TidepoolVCLA.pdf on June 9, 2016.


jh-bate:
:shipit: - thanks @PeterSampson

PR #319, opened 2016-08-02
Post VCA Additions

krystophv:
Styling fixes on in-clinic created account verification and addition of a few metrics surrounding that process.

### DIFF
--- a/app/pages/verificationwithpassword/verificationwithpassword.js
+++ b/app/pages/verificationwithpassword/verificationwithpassword.js
@@ -65,6 +65,13 @@ export let VerificationWithPassword = React.createClass({
     }
   },
 
+  componentWillReceiveProps: function(nextProps) {
+    if (_.get(this.props, 'notification.message', null) === null &&
+        _.get(nextProps, 'notification.message') === errorMessages.ERR_BIRTHDAY_MISMATCH) {
+      this.props.trackMetric('VCA Home Verification - Birthday Mismatch')
+    }
+  },
+
   getInitialState: function() {
     var formValues = {};
 
@@ -88,9 +95,6 @@ export let VerificationWithPassword = React.createClass({
   },
 
   render: function() {
-    if (_.get(this.props, 'notification.message', '') === errorMessages.ERR_BIRTHDAY_MISMATCH) {
-      this.props.trackMetric('VCA Home Verification - Birthday Mismatch')
-    }
     return (
       <div className="VerificationWithPassword">
         <LoginNav
--- a/app/pages/verificationwithpassword/verificationwithpassword.js
+++ b/app/pages/verificationwithpassword/verificationwithpassword.js
@@ -85,7 +85,8 @@ export let VerificationWithPassword = React.createClass({
       <div className="VerificationWithPassword">
         <LoginNav
           page="VerificationWithPassword"
-          trackMetric={this.props.trackMetric} />
+          trackMetric={this.props.trackMetric}
+          hideLinks={true} />
         <LoginLogo />
         <div className="container-small-outer VerificationWithPassword-form">
           <div className="container-small-inner VerificationWithPassword-form-box">
--- a/app/pages/verificationwithpassword/verificationwithpassword.js
+++ b/app/pages/verificationwithpassword/verificationwithpassword.js
@@ -19,6 +19,7 @@ import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 
 import * as actions from '../../redux/actions';
+import * as errorMessages from '../../redux/constants/errorMessages';
 
 import _ from 'lodash';
 import config from '../../config';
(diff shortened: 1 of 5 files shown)


krystophv:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;VWfUa0A2&#x2F;1-ux-0-dev-3-tst-1-uploader-in-clinic-post-release-feature-cleanup"> &lt;ux:0 dev:3 tst:1&gt; [uploader] [in-clinic] Post-release feature cleanup</a>

krystophv:
ping @jebeck for PR review

app/pages/verificationwithpassword/verificationwithpassword.js
```diff
@@ -81,11 +88,15 @@ export let VerificationWithPassword = React.createClass({
   },
 
   render: function() {
+    if (_.get(this.props, 'notification.message', '') === errorMessages.ERR_BIRTHDAY_MISMATCH) {
+      this.props.trackMetric('VCA Home Verification - Birthday Mismatch')
+    }
```
jebeck: Side effects in a `render` method make me 😧. Definitely more appropriate in `componentWillReceiveProps` at worst, at best is there a way to fold it into the redux actions going on?

app/pages/verificationwithpassword/verificationwithpassword.js
```diff
@@ -58,6 +59,12 @@ export let VerificationWithPassword = React.createClass({
     this.setState({ loading: false });
   },
 
+  componentDidMount: function() {
+    if (this.props.trackMetric) {
+      this.props.trackMetric('VCA Home Verification - Screen Displayed');
```
jebeck: I suppose in blip we don't have the strings of the metrics separated out as constants? Can you double-check?

app/redux/utils/trackingMiddleware.js
```diff
@@ -25,7 +25,8 @@ const trackMetricMap = {
   SETUP_DATA_STORAGE_SUCCESS: 'Created Profile',
   UPDATE_PATIENT_SUCCESS: 'Updated Profile',
   UPDATE_USER_SUCCESS: 'Updated Account',
-  LOGOUT_REQUEST: 'Logged Out'
+  LOGOUT_REQUEST: 'Logged Out',
+  VERIFY_CUSTODIAL_SUCCESS: 'VCA Home Verification - Verified'
```
jebeck: Hmm, for testing these metrics (since that's on my mind), we'd need to add our metrics middleware to the config of the redux-mock-store and add the metrics to the expectations, yes? I think maybe we should try that.

jebeck:
A couple of TODOs (other than things mentioned in in-line comments):
- [ ] like the uploader PR, final in-browser design review with SK before merge; in particular, the centering of elements on the verification still look off to me... I tried to look into it and it's baffling me a bit, but look at the footer pieces (in particular the center 'Get Support' vs. form vs. the blip logo)
- [ ] unit tests for metrics calls added to verification with password...should be able to use the lifecycle methods and sinon stubs to do this pretty easily

app/pages/verificationwithpassword/verificationwithpassword.js
krystophv: Every call to `trackMetric` I could find passed in a string. Couldn't find any list of metrics constants.

app/pages/verificationwithpassword/verificationwithpassword.js
krystophv: I struggled with this a little. Yes, you're right that it should get out of render (my bad) and into one of the other lifecycle methods. This is a result of the server giving back an error, so we can't track it before the prop comes back. I think that ideally, it would happen where the `verifyCustodialFailure` is created in the `verifyCustodial` async action creator, but that would mean making a direct API call to the metrics tracking, which isn't done anywhere else in the codebase, so I was trying to avoid adding one. Do you think it would be best to add the tracking API call in the action creator?

jebeck: Hmm, tough one. Perhaps can we try doing it in `componentWillReceiveProps` and look carefully to see if it gets triggered more than once? I suppose we also might be able to do something like only call the metric if `this.props.notification.message === null && nextProps.notification.message === ERR_BIRTHDAY_MISMATCH` to avoid triggering except on initial prop change. Then we keep an eye on this type of situation - if an additional one or two things come along where making a metrics call directly within the action creators would be best, then I think we redo the metrics middleware to match the uploader's more closely so that the middleware looks for metrics in all actions and dispatches them as-is rather than dispatching a pre-configured metric for certain action types. Does that make sense?

krystophv: Yeah, that all sounds reasonable.

krystophv:
@jebeck - design review complete, unit tests added.

jebeck:
LGTM, 👍

PR #320, opened 2016-08-12
MVP CGM in trends view

jebeck:
Summary of changes here:
- add additional code for the footer, to add the toggle between CGM & CGM for trends view as well as only show the other options when on BGM view
- add the same back/forward/most recent buttons used in Daily view to header for navigation of trends view (and get rid of the mini scroll bar that was used instead before)
- get rid of the "inner" `<ModalChart />` component and replace it with the `<TrendsContainer />` in @tidepool/viz
- added the `<SMBGTrends />` component as a temporary interface layer until we port over the smbg version of trends view to @tidepool/viz
- add the `vizReducer` from @tidepool/viz to the redux root reducer here, just handles a couple of things for trends view at the moment


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;qQqZiU3z&#x2F;3-ux-2-dev-8-tst-3-blip-cgm-in-trends-mvp-release-candidate"> &lt;ux:2 dev:8 tst:3&gt; [blip] [cgm in trends] MVP release candidate</a>

app/components/chart/footer.js
```diff
@@ -74,28 +77,41 @@ var TidelineFooter = React.createClass({
...
 
-    
     return (
       <div className="container-box-outer patient-data-footer-outer">
         <div className="container-box-inner patient-data-footer-inner">
           <div className="patient-data-footer-left">
             <button className="btn btn-chart btn-refresh"
               onClick={this.props.onClickRefresh}>
               Refresh</button>
+            <span className="toggle-container">
```
krystophv: Would it make sense to move this wrapping `<span>` into the `bgDataToggle` instead of having it hang out in the DOM when it has no contents?

app/components/chart/footer.js
```diff
@@ -74,28 +77,41 @@ var TidelineFooter = React.createClass({
...
-    
-    var rightSide = this.props.chartType === 'weekly' ? showValues :
-      this.props.chartType === 'modal' ? modalOpts : null;
-    
+    var rightSide = null;
+    var bgDataToggle = null;
+
+    if (this.props.chartType === 'weekly') {
+      rightSide = showValues;
+    } else if (this.props.chartType === 'modal') {
```
krystophv: Is this `else` necessary, or could it just be `<newline> if (this.props.chartType === 'modal') {`?

app/components/chart/footer.js
```diff
@@ -74,28 +77,41 @@ var TidelineFooter = React.createClass({
...
+      if (this.props.showingSmbg) {
+        rightSide = modalOpts;
+      }
+      bgDataToggle = (
+        <TwoOptionToggle
+          left={{ label: 'BGM', state: this.props.showingSmbg }}
+          right={{ label: 'CGM', state: this.props.showingCbg }}
```
krystophv: Is there a reason that we're storing this state in two props as opposed to the second just being `!this.props.showingSmbg` or something similar? It just feels strange to have essentially one piece of information stored as two things that will always be the inverse of one another.

jebeck: @krystophv and I discussed today, and I explained that this is a relic of the fact that at the beginning of the "pancakes" process we weren't certain whether we might give users the option to layer both BGM and CGM data into the view together (rather than toggle between one or the other). I don't believe we've definitively decided _against_ layering, so @krystophv and I agreed we'll leave this for now. If we do definitely decide against layering in the future, we should consolidate into one variable for cleaner, safer, more understandable code. I will add a comment here in the meantime to explain why it is the way it is.

app/components/chart/footer.js
jebeck: Nah, not necessary. And your suggestion is easier to read, so I've updated.

app/components/chart/footer.js
jebeck: Er, yes. Done.

krystophv:
👍 LGTM :shipit:

PR #322, opened 2016-08-17
get the verifycustodialsuccess to fire after login

krystophv:


krystophv:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;FQZoxati&#x2F;19-blip-in-clinic-account-creation-an-event-should-be-created-when-a-user-verifies-a-custodial-account-from-home"> [blip] [in-clinic account creation] An event should be created when a user verifies a custodial account from home.</a>

krystophv:
ping @jh-bate for review

jh-bate:
LGTM

PR #324, opened 2016-08-18
Updated to new tidepool URL for typeform

HowardLook:
@jebeck @jh-bate @krystophv Can you please do a quick review? At my request, @brandonarbiter updated the Typeform URL from brandon72.typeform.com to tidepool.typeform.com for the clinic survey, but that also changed the sign up URL. The old one is still working and may continue to work, but I'm a little nervous that it may stop if/when typeform updates DNS and propagation occurs.

Thanks!

### DIFF
--- a/app/components/waitlist/waitlist.js
+++ b/app/components/waitlist/waitlist.js
@@ -21,7 +21,7 @@ var WaitList = React.createClass({
   render: function() {
     return (
       <div className="waitlist-container" >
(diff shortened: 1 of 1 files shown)


jebeck:
:shipit: 

@krystophv can you do the merge as you see fit with your sequencing of other changes? Sounds like @HowardLook would like it out ASAP.

krystophv:
It might be able to go out with https://github.com/tidepool-org/blip/pull/322 - I'll try to get an idea of test/sequencing from @brandonarbiter

PR #326, opened 2016-08-24
point to the correct of app/main.(prod.)js

jebeck:
Looks like 'app.js' was being used for both dev and prod for a while, and it was only when I separated out the config bundling that it surfaced as a problem...somehow.

As discussed with @darinkrauss, in the near future (possibly on the 🛩 tomorrow) I'll try to do the hour or so work it would take to cleanly separate the dev and prod webpack configs (as I've seen in quite a few React app boilerplates) so that it's easier to read and something like this doesn't happen again.

Ping @darinkrauss for review?

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -19,8 +19,7 @@ var defineEnvPlugin = new webpack.DefinePlugin({
 });
 
 var plugins = [ defineEnvPlugin, new ExtractTextPlugin('style.[contenthash].css') ];
-var appEntry = './app/main.js';
-var entryScripts = ['babel-polyfill', appEntry];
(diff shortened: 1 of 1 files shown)


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;AMRR0T4w&#x2F;2-ux-0-dev-1-tst-2-blip-cgm-in-trends-object-assign-error-in-blip"> &lt;ux:0 dev:1 tst:2&gt; [blip] [cgm in trends] Object.assign error in blip</a>

jh-bate:
ok -based on that explanation LGTM @jebeck

PR #329, opened 2016-09-16
pass through currentPatientInViewId

jebeck:
What it says on the tin. Required for [viz #13](https://github.com/tidepool-org/viz/pull/13).


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;VJ1JQDTe&#x2F;34-viz-bugfix-for-showing-cgm-trends-view-even-when-not-enough-data-after-logout-and-switching-patients"> [viz] bugfix for showing CGM trends view even when not enough data after logout and switching patients</a>

PR #342, opened 2016-10-24
Browser Metrics

hntrdglss:
Adds tracking:
- when `BrowserWarning` view appears
- when download link is clicked in `BrowserWarning`
- when Blip URL is copied in `BrowserWarning`

Splits tracking for refresh buttons, to differentiate between the footer refresh and "no data" refresh, on `Basics`, `Settings`, and `Weekly` views.

[Trello card](https://trello.com/c/CjmoTwOP/41-capture-metrics-for-non-chrome-switch-to-chrome-dev-1-ux-0-tst-1-blip) for more info!

@jebeck @krystophv @jh-bate


krystophv:
[capture metrics for non-Chrome switch to Chrome [dev: 1 ux: 0 tst: 1] [blip]](https://trello.com/c/CjmoTwOP/41-capture-metrics-for-non-chrome-switch-to-chrome-dev-1-ux-0-tst-1-blip)

hntrdglss:
@krystophv Ah, yes! I meant to ask about that in my description, but forgot while I was waiting for GitHub to come back up. 😞 I updated it to be `readOnly`. I did have concerns if older browsers would allow selecting read only text, because I recall some did way back when, but it seems like most relatively recent ones are good now. (In that they've standardized to allow selecting for `readOnly` and no selecting for `disabled`.)

Also, added the unit test to check for the display metric. Thanks for calling that out!

krystophv:
LGTM 👍

PR #345, opened 2016-10-27
Improve main upload button

hntrdglss:
As the title says, this PR changes the "no data" screen to make the upload CTA more button-like. It also adds some supporting text. See [Trello card](https://trello.com/c/wdJDJVV1) for more info.

Few things I'd like to call out:
- as you'll likely notice and take pause over, this includes some CSS overrides for link colors to be consistent with Uploader branding colors, instead of Blip's (i.e., purple, not green)
- I split `.patient-data-message` and `.patient-data-message-no-data`, since we substantially changed the appearance of some elements defined in `.patient-data-message` (and updated tests accordingly)
- I added a new component, `UploaderButton`, and corresponding tests (Anything else I should test on it? It's pretty basic.)

(N.B.: the Tidepool logo included has a purple background that's just _slightly_ different than the color defined in CSS in the [Uploader](https://github.com/tidepool-org/chrome-uploader/blob/master/styles/core/variables.less#L21). I'm checking with Sara to see which she'd like to change.)

Thanks, @jebeck @krystophv @jh-bate!


test/unit/components/uploaderbutton.test.js
```diff
@@ -0,0 +1,28 @@
...
+
+import UploaderButton from '../../../app/components/uploaderbutton';
+
+describe('UploaderButton', function () {
+  it('should be a function', function() {
+    expect(UploaderButton).to.be.a('function');
+  });
+
+  describe('render', function() {
+    it('should render without problems', function () {
```
jh-bate: I guess we could also test
- the onclick handler works as expected, as making sure the button works feels important
- the text matches what we have given, making sure its not just hardcoded in the `UploaderButton` component

app/components/uploaderbutton/uploaderbutton.js
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+import React, { Component } from 'react'
+
+import logoSrc from './images/T-logo-dark-512x512.png';
+
+export default class UploaderButton extends Component {
+  static propTypes = {
+    buttonUrl: React.PropTypes.string.isRequired,
+    onClick: React.PropTypes.func,
```
jh-bate: the `onClick` should be required shouldn't it?

hntrdglss: My thinking was that maybe you'd want a button to link somewhere, without having an `onClick` handler (for tracking or a callback or whatever). Just an old school link. :) But I'm not opposed to requiring it, especially if that's our convention!

app/pages/patientdata/patientdata.js
```diff
@@ -177,38 +178,37 @@ export let PatientData = React.createClass({
...
-          <a
-            href={this.props.uploadUrl}
-            target="_blank"
-            onClick={handleClickUpload}>Upload data</a>
-          <p>
-            {'Or try '}<a href="" onClick={this.handleClickNoDataRefresh}>refreshing</a>{' the page.'}
+        <div className="patient-data-uploader-message">
+          <h1>To see your data, you’ll need to upload it!</h1>
+          <UploaderButton
+            buttonUrl='http://chrome.google.com/webstore/detail/tidepool-uploader/cabklgajffclbljkhmjphejemhpbghfb'
```
jebeck: Could we pull these URLs out as constants or perhaps defaultProps for easier access in future in case they need to be updated?

jebeck:
As for new component style, I would say we should try to match the set up we have in viz with the AirBnb ESLint, although you won't get hints in your editor to help enforce it in blip (unless we can figure out some kind of complex ESLint config set up). (We have since the redux refactor been refactoring some older components into ES6 as we touch them and writing new components in ES6.) So the the new upload button, @hntrdglss, can you do the small amount of refactoring to make the component a pure functional component since the constructor and lifecycle methods are vacuous? It would also be great if the new test file is in ES6 with `import` instead of `require` etc. but up to you whether you want to change that now.

app/components/uploaderbutton/uploaderbutton.js
hntrdglss: 👍

hntrdglss:
@jebeck @jh-bate Updated with your feedback, when you have time to review.

app/core/less/variables.less
```diff
@@ -26,6 +26,7 @@
 @purple-highcontrast:    #291A45;
 @purple-success:         #7950ea;
 @purple-blip:            #4E4490;
+@purple-uploader:        #291448;
```
jebeck: Was this where there was a question about what purple to use? I wonder if this is supposed to match [the darkest purple we use in the dataviz](https://github.com/tidepool-org/viz/blob/master/src/styles/colors.css#L28) (for avg on smbg trends and median on cbg trends). No action item for now, but just raising awareness among @hntrdglss @krystophv @jh-bate that we should pay attention to this and try to align blip & viz at the next opportunity if indeed there is a mismatch as I suspect.

test/unit/components/uploaderbutton.test.js
```diff
@@ -0,0 +1,48 @@
...
+      expect(wrapper.find(UploaderButton)).to.have.length(0);
+    });
+
+    it('should have provided button text', function () {
+      expect(wrapper.find('.btn-uploader').someWhere(n => (n.text().search(props.buttonText) !== -1))).to.be.true;
+    });
+
+    it('should respond to an onClick event', () => {
```
jebeck: Nitpick, but since sinon spies are stateful I think it's best to include an expectation of a zero callCount before the click simulation. I *think* we tend to do that in this codebase because I think I learned it from Gordon who wrote a lot of the tests in blip...

app/core/less/variables.less
hntrdglss: Yes, exactly. I'll be discussing this with Sara during our design review. I'll be sure to loop you all in after.

jebeck: Oh I didn't realize there was a design review left to do.

test/unit/components/uploaderbutton.test.js
jebeck: I'm not picky about the technique, although I think what we do elsewhere is just check for 1 (but we also include a spy.reset() in an `afterEach` method when there are multiple tests on spies).

PR #350, opened 2016-11-10
Temp fix for Webpack image issue

hntrdglss:
Local environment looks like 
image

Dev environment looks like 
image

Problem is that locally it includes the protocol, hostname, and root path, but on dev it doesn't. So, the image path goes from http://localhost:3000/0f0318e861c4012eb735c227401ebad9.png to https://dev-blip.tidepool.org/patients/<PATIENT_HASH>/0f0318e861c4012eb735c227401ebad9.png

### DIFF
--- a/app/components/uploaderbutton/uploaderbutton.js
+++ b/app/components/uploaderbutton/uploaderbutton.js
@@ -16,9 +16,16 @@
 
 import React, { Component } from 'react'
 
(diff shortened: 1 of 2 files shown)


hntrdglss:
Closing, since partially fixed in #351. We'll still have the issue, if it's a larger image on a non-root page, but I'll submit a separate PR once I find a resolution for that.

PR #353, opened 2016-11-15
Add metric tracking to Blip Notes link

hntrdglss:
We neglected to add metric tracking to a new Blip Notes link on the improved main upload button (PR #345), so here it is! See [Trello card](https://trello.com/c/cb5ujwjI/) for more info.

/cc: @krystophv @jebeck @jh-bate

### DIFF
--- a/app/pages/patientdata/patientdata.js
+++ b/app/pages/patientdata/patientdata.js
@@ -177,6 +177,9 @@ export let PatientData = React.createClass({
     var handleClickUpload = function() {
       self.props.trackMetric('Clicked No Data Upload');
     };
+    var handleClickBlipNotes = function() {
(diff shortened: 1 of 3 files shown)


hntrdglss:
:+1: Yes, I was going to ask about that during stand-up. I see we do it in a few places, but I also see it missing from a lot of others. I wondered if we had a consensus on testing metrics.

jebeck:
I'd say our consensus is to test metrics if at all possible - we learned the hard way when they were borked once in the uploader. Blip, as in lots of ways, is "legacy" and doesn't have full test coverage; this is definitely one of the areas where we'd like to add test coverage as we add/update functionality.

PR #363, opened 2016-12-01
CBG trends display selections

krystophv:
Use the new display toggles from `viz` and track metrics.

PR #366, opened 2016-12-02
perf stuff

jebeck:
I did it @krystophv!!! 🎉 🎊 🎉

We just weren't using webpack right /o\

This PR also contains the addition of the React Perf addon in dev (attached to `window` so you can use in the console, as described e.g., here: http://benchling.engineering/performance-engineering-with-react/). Sticking it in the redux dev mode `<Root />` component seemed like the easiest way (to me) to keep it out of prod, but LMK if you can think of a cleaner way.

@hntrdglss for context: @krystophv noticed a while ago that we were clearly still using the dev build of React in production because propTypes validation shows up all over any perf profiles...

### DIFF
--- a/app/redux/containers/Root.dev.js
+++ b/app/redux/containers/Root.dev.js
@@ -1,4 +1,6 @@
 import React, { Component } from 'react';
+import Perf from 'react-addons-perf';
+window.Perf = Perf;
 import { Provider } from 'react-redux';
 import { Router, browserHistory } from 'react-router';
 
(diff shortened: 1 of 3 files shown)


jebeck:
OK, so let's plan to merge this along with something else upcoming so that the production build gets a smoke test along with whatever else. Either with the smbg Trends port or the CGM trends range selections I guess. Thoughts @krystophv?

jebeck:
Closing to re-open against @krystophv CGM checkboxes work (#363).

PR #387, opened 2017-02-13
Print settings

jh-bate:
@krystophv if you could start having a look at this with regards to https://trello.com/c/CYlJtmZg


jh-bate:
[<ux:2 dev:8 tst:2> Device settings print view in Blip](https://trello.com/c/CYlJtmZg/87-ux-2-dev-8-tst-2-device-settings-print-view-in-blip)

app/pages/patientdata/patientdata.js
```diff
@@ -317,26 +356,13 @@ export let PatientData = React.createClass({
...
-            onSwitchToSettings={this.handleSwitchToSettings}
-            onSwitchToWeekly={this.handleSwitchToWeekly}
-            trackMetric={this.props.trackMetric}
-            uploadUrl={this.props.uploadUrl}
-            ref="tideline" />
-          );
-
+      case 'settingsprintview':
+        // set here so that we default back settings and not print
+        this.setState({ chartType: 'settings' });
```
krystophv: So this is the `setState()`-in-`render()` pattern that React is particularly not a fan of. (since `render()` is called immediately following any `setState()`, this can often lead to a cyclical call and kill everything). Since we call the same `renderSettings()` after this regardless, is there any harm in moving the `window.print()` call directly to the `handleSwitchToSettingsPrintView` function (replacing the `setState()` there) and just dropping this whole `case`? [I tested it quickly, and it seemed to function as I expected and format the printout identically, but I haven't dug real deep into things I might have missed]

app/pages/patientdata/patientdata.js
```diff
@@ -488,6 +514,18 @@ export let PatientData = React.createClass({
...
   },
 
+  handleSwitchToSettingsPrintView: function(e) {
+    this.props.trackMetric('Clicked Switch To Settings Print View', {
+      fromChart: this.state.chartType
+    });
+    if (e) {
+      e.preventDefault();
+    }
+    this.setState({
```
krystophv: [This is where I was recommending putting the `window.print()` call and dropping this `setState()`]

app/pages/print/print.js
```diff
@@ -0,0 +1,380 @@
+/**
```
krystophv: I might be missing it, but I can't find where anything actually `require()`s this page? Is it actually being used anywhere?

jh-bate: ahh crap - yeah thats a relic of the original POC work that I need to cleanup. Ta

jh-bate:
@krystophv good for another look along with https://github.com/tidepool-org/viz/pull/56. The one caveat is that it requires the version of node used in https://github.com/tidepool-org/blip/pull/388 as per our discussion

package.json
```diff
@@ -49,6 +49,7 @@
     "react-addons-update": "15.2.1",
     "react-dom": "15.2.1",
     "react-hot-loader": "3.0.0-beta.2",
+    "react-print": "^1.3.1",
```
krystophv: For now we're always specifying exact versions (no `^`) of dependencies.

krystophv:
Need to merge down from master, but changes LGTM 👍

PR #389, opened 2017-02-16
Add /preferences, /settings endpoints, re-enable siteChangeSource saving

hntrdglss:
This adds new API endpoints for saving `preferences` and `settings`. As [defined by @darinkrauss ](https://tidepoolteam.slack.com/archives/eng/p1486581018000676) in the #eng chat room, `preferences` are visible only to the user, but `settings` are visible to both the user/patient and any caregivers (shared with another account, VCAs, custodial accounts). I believe that's how I've implemented it, and testing seems to agree, but `app/core/api.js` is very convoluted, so please let me know if I missed something.

I'm not using `preferences` for anything here, but went ahead and set up the scaffolding for it. For `settings`, I've updated the `siteChangeSource` to save to that endpoint. We should also pay close attention to the `profile`, as some of the refactoring I did in `platform-client` also updated the API calls there (to reduce redundancy).

Related PRs:
- [`tideline`](https://github.com/tidepool-org/tideline/pull/305)
- [`platform-client`](https://github.com/tidepool-org/platform-client/pull/74)

Related Trello cards:
- [Persist site change source property to the backend](https://trello.com/c/8rMfVYk1)
- [Persisting BG prefs to the backend](https://trello.com/c/rZKNx9bb) (has some information on the endpoint documentation)

@krystophv and/or @jh-bate, would you mind taking a look when you have time?


app/core/api.js
```diff
@@ -458,6 +474,58 @@ api.patient.getAll = function(cb) {
...
 
+// ----- Metadata -----
+
+api.metadata = {};
+
+api.metadata.preferences = {};
+
+api.metadata.preferences.get = function(patientId, cb) {
+  tidepool.findPreferences(patientId, function(err, payload) {
+    if (err && err.status !== 404) {
```
hntrdglss: The `&& err.status !== 404` here and [here](https://github.com/tidepool-org/blip/pull/389/files#diff-a9e125b91e9e2c58c685083cbfa74a49R509) is a bit of an antipattern, compared to other similar functions. My reasoning is that we don't want a login or patient load to fail if they don't have any preferences/settings saved yet. Alternatives could be updating the backend prior to launch, so all accounts have empty objects for these endpoints, or setting a migration plan like [this profile update](https://github.com/tidepool-org/blip/blob/19ae3fccaff5456c871e8f058742a45175ebfebf/app/core/api.js#L188-L205), but this seemed like the cleanest option until the account eventually has preferences/settings saved. I'm not sure what our view is on making an API call (or updating a database) for what essentially is a non-operation.

hntrdglss: The `&& err.status !== 404` here and [here](https://github.com/tidepool-org/blip/pull/389/files#diff-a9e125b91e9e2c58c685083cbfa74a49R509) is a bit of an antipattern, compared to other similar functions. My reasoning is that we don't want a login or patient load to fail if they don't have any preferences/settings saved yet. Alternatives could be updating the backend prior to launch, so all accounts have empty objects for these endpoints, or setting a migration plan like [this profile update](https://github.com/tidepool-org/blip/blob/19ae3fccaff5456c871e8f058742a45175ebfebf/app/core/api.js#L188-L205), but this seemed like the cleanest option until the account eventually has preferences/settings saved. I'm not sure what our view is on making an API call (or updating a database) for what essentially is a non-operation.

hntrdglss:

jebeck:
[Infusion site changes should include tube and cannula primes [blip] [basics] [dev: 5+3 ux: 3 tst: 2]](https://trello.com/c/DEuGTWqq/16-infusion-site-changes-should-include-tube-and-cannula-primes-blip-basics-dev-5-3-ux-3-tst-2)

jebeck:
Tag v0.16.8 is a no-go because the bugfix in #391 affected the VCA behavior. I've found the issue, which is that the `patient` object being passed down to the Basics was a "dirty" one mutated in a strange way (attaching the permissions of the _logged-in user_ to the patient object, not the permissions of the patient on itself; see changes to patients.js in #391). So when I undid the mutation in #391 the VCA's ability to change (but not store) the site change source if it still has `custodian` permissions is not working. Unfortunately what this shows is that there may be other code relying on this "dirty" `patient` object (though I actually suspect there might not be! because we don't use permissions very many places yet), so tomorrow I'm going to a) pass down the logged-in user's permissions on the current patient in view more explicitly/separately to the `Basics` and b) try to sort through all the places where we refer to `patient.permissions` to see if any behavior broke there.

Regression tests will be a goal too.

So contrary to @jh-bate's comment above, I think if you're back Wednesday @hntrdglss you should address his comments on this branch, and we can certainly touch base and try to divide up some of this buggy behavior clean-up work.

app/core/api.js
```diff
@@ -332,7 +338,12 @@ function getPatient(patientId, cb) {
       }
 
       person.permissions = permissions;
-      return cb(null, person);
+
+      api.metadata.settings.get(patientId, function(err, settings) {
```
hntrdglss: Ah, no, that was an oversight. We should check for a non-404 error.

app/core/api.js
```diff
@@ -404,7 +415,12 @@ api.patient.get = function(patientId, cb) {
           return member;
         });
         patient.team = members;
-        return cb(null, patient);
+
+        api.metadata.settings.get(userId, function(err, settings) {
```
hntrdglss: Ah, no, that was an oversight. We should check for a non-404 error.

PR #390, opened 2017-02-20
we require the permissions

jh-bate:
@hntrdglss here is what I was seeing and this update now works for me. Below is what I was seeing in the console which lead me to see in both cases we were getting all the data (even though order was different). Then I noticed on the subsequent load the `patient.permissions` were being overridden.

### On initial load of data

```
action @ 10:02:03.536 @@router/UPDATE_LOCATION 
action @ 10:02:03.558 FETCH_USER_REQUEST 
action @ 10:02:03.600 FETCH_PATIENT_REQUEST 
action @ 10:02:03.612 FETCH_PATIENT_DATA_REQUEST 
action @ 10:02:07.657 FETCH_USER_SUCCESS 
action @ 10:02:07.905 FETCH_PATIENT_DATA_SUCCESS 
action @ 10:02:12.766 FETCH_PATIENT_SUCCESS 
```

### On subsequent load of data when you refresh the page

```
action @ 10:03:22.178 FETCH_USER_REQUEST 
action @ 10:03:22.213 FETCH_PATIENT_REQUEST 
action @ 10:03:22.226 FETCH_PATIENT_DATA_REQUEST 
action @ 10:03:28.331 FETCH_USER_SUCCESS 
action @ 10:03:32.398 FETCH_PATIENT_SUCCESS 
action @ 10:03:32.492 FETCH_PATIENT_DATA_SUCCESS 
```

### DIFF
--- a/app/redux/reducers/misc.js
+++ b/app/redux/reducers/misc.js
@@ -150,10 +150,9 @@ export const allUsersMap = (state = initialState.allUsersMap, action) => {
         });
       } else {
         newState = update(state, {
-          [patient.userid]: { $set: _.omit(patient, ['permissions', 'team'])}
+          [patient.userid]: { $set: _.omit(patient, ['team'])}
         });
       }
-
       const { team } = patient;
       if (team) {
         let others = {};


jebeck:
As an update for anyone else following: @jh-bate asked me to take a look at this, and it appears to me that the combo of @hntrdglss's revised changes (open PRs on tideline, p-c, and #389 here) fix this issue. I'm going to try to trace through so we understand _why_.

app/redux/reducers/misc.js
```diff
@@ -150,10 +150,9 @@ export const allUsersMap = (state = initialState.allUsersMap, action) => {
         });
       } else {
         newState = update(state, {
-          [patient.userid]: { $set: _.omit(patient, ['permissions', 'team'])}
+          [patient.userid]: { $set: _.omit(patient, ['team'])}
```
hntrdglss: I wonder why `permissions` were originally omitted? I guess for `team` it makes sense, due to the below transformation into `others`. (Though I also don't follow why `permissions` are omitted there from the member.)

jebeck:
@hntrdglss please see #391 

This was a much more complex issue. @jh-bate asked me to look into it, and #391 with the fix that actually solves the root issue is still in progress.

We very much intentionally omit `permissions` from the `allUsersMap`, as we store permissions [normalized](http://redux.js.org/docs/recipes/reducers/NormalizingStateShape.html) in two other places in the state tree: `membershipPermissionsInOtherCareTeams` (for the permission the logged-in user has in others' care teams) and `permissionsOfMembersInTargetCareTeam` (corresponds to the `team` on a `patient` object returned from a fetch patient). `permissions` **were** sneaking into the state tree due to a bug where we were mutating the state tree directly in `mapPropsToState` (see the fix in 17cf03129d8e12e6ac4b41eca059546607f82875). (This is also documented to some degree of detail [here](http://developer.tidepool.io/blip/docs/StateTreeGlossary.html#userrelated-state).)

I would have preferred if we had actually discussed how to integrate the new patient `settings` and `preferences` into the state tree prior to your implementation: I think there is an argument to be made for omitting those from `allUsersMap` as well (i.e., normalization), although not quite as strong as the case with the various permission relations. However, there is also a case for putting the `siteChangeSource` in the `viz` branch of the state tree... Overall I think it's a interesting question, and such questions are best solved through discussion.

PR #391, opened 2017-02-21
bugfix race condition state tree

jebeck:
A fix that gets more to the root of the problem than #390 

As discussed @jh-bate I have a bit more to add on this:
- [x] fixing (if they fail)/adding some regression tests
- [x] a bit more state tree reorg re: patient settings

But please try it and see if it solves the issue for you!

Fuller explanation of this bug:

@jh-bate discovered a bug where blip was crashing on some accounts he had in the situation where of the three requests made when hard refreshing (w/remember me) on the data view route (/patients/:id/data) route, the fetch patient (profile, etc.) and fetch patient data requests were coming back prior to the fetch *user* request. This is pretty unusual and I'd guess is connected to very little data being present in the accounts in question. (That doesn't explain why the fetch patient is beating the fetch user request back, but `¯\_(ツ)_/¯`.)

The patient data view renders the loading GIF until both `fetchingPatient` and `fetchingPatientData` are `false`; it was assuming that `fetchingUser` was already complete. So the first step (eb0df6d7cb8fa1b57dcbf8e033e3cc82aee46cf8) was to add a check against `fetchingUser` as well. But this just...delayed the crash until after `fetchingUser` came back. So the next step was to actually figure out the crash.

The crash occurred when trying to access `permissions` on the `patient` object passed through to the Basics code (to determine whether or not the choice of site change source could be persisted). Very long story (and much debugging over Hangouts between @jebeck and @jh-bate) short, it was discovered that `permissions` looked accessible (via the state logged in the console with redux-logger) but were actually **not**, always, because we were incorrectly mutating the redux store's state directly in both app.js and patientdata.js's `mapStateToProps` function, so we had our state mutating out from under us in a way that in this case with the combination of requests coming back in an odd order and the timing of the mutation resulted in no `permissions` being available. 17cf03129d8e12e6ac4b41eca059546607f82875 solves this issue by properly creating fresh `patient` objects (including `permissions` extracted from the `membershipPermissionsInOtherCareTeams` branch of the state tree where they are stored) to pass down as props as the API of the lower components in the tree requires.

The remaining commits in this PR fix and add regression tests, checking for mutation in `mapStateToProps` using our `object-invariant-test-helper` that we already use to check for mutation in our reducer tests (for complex array or object branches of the state tree). I also dropped in the `redux-immutable-state-invariant` (the project our own `object-invariant-test-helper` is based on) in our dev middleware to warn us of any mutations to the redux store state in development (by crashing the app 😮 ). I'll admit I used to think we didn't need this tooling since we test for mutation in our reducers, but my incorrect base assumption here was that the reducers are the only place you might accidentally mutate your redux store state! We may want to move this middleware behind its own flag in the future (but in the meantime just use `export DEV_TOOLS=false` if you need to turn it off) as due to all the deep comparisons it does, it slows down the app *considerably*.

Though @jh-bate discovered this bug on `master` when he pulled in the updates from @hntrdglss's recent site change source setting work, it was **not** related to that work, but rather a combination of an unusual account situation/replicable race condition in the ordering of requests coming back and a *very* old bug dating from our migration to redux for state management in blip.

### DIFF
--- a/app/pages/patientdata/patientdata.js
+++ b/app/pages/patientdata/patientdata.js
@@ -49,6 +49,7 @@ export let PatientData = React.createClass({
     fetchers: React.PropTypes.array.isRequired,
     fetchingPatient: React.PropTypes.bool.isRequired,
     fetchingPatientData: React.PropTypes.bool.isRequired,
+    fetchingUser: React.PropTypes.bool.isRequired,
     isUserPatient: React.PropTypes.bool.isRequired,
     messageThread: React.PropTypes.array,
     onCloseMessageThread: React.PropTypes.func.isRequired,
@@ -122,7 +123,7 @@ export let PatientData = React.createClass({
   },
 
   renderPatientData: function() {
-    if (this.props.fetchingPatient || this.props.fetchingPatientData || this.state.processingData) {
+    if (this.props.fetchingUser || this.props.fetchingPatient || this.props.fetchingPatientData || this.state.processingData) {
       return this.renderLoading();
     }
 
@@ -636,6 +637,7 @@ export function mapStateToProps(state) {
     messageThread: state.blip.messageThread,
     fetchingPatient: state.blip.working.fetchingPatient.inProgress,
     fetchingPatientData: state.blip.working.fetchingPatientData.inProgress,
+    fetchingUser: state.blip.working.fetchingUser.inProgress,
     viz: state.viz,
   };
 }
--- a/app/pages/patientdata/patientdata.js
+++ b/app/pages/patientdata/patientdata.js
@@ -616,22 +616,33 @@ let getFetchers = (dispatchProps, ownProps, api) => {
 };
 
 export function mapStateToProps(state) {
-  var user = null;
-  var patient = null;
+  let user = null;
+  let patient = null;
+  let permissions = {};
+
   if (state.blip.allUsersMap){
     if (state.blip.loggedInUserId) {
       user = state.blip.allUsersMap[state.blip.loggedInUserId];
     }
 
     if (state.blip.currentPatientInViewId){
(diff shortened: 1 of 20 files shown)


jebeck:

app/pages/patientdata/patientdata.js
```diff
@@ -615,27 +616,39 @@ let getFetchers = (dispatchProps, ownProps, api) => {
...
     user: user,
     isUserPatient: personUtils.isSame(user, patient),
-    patient: patient,
+    patient: { permissions, ...patient },
     patientDataMap: state.blip.patientDataMap,
     patientNotesMap: state.blip.patientNotesMap,
     messageThread: state.blip.messageThread,
     fetchingPatient: state.blip.working.fetchingPatient.inProgress,
     fetchingPatientData: state.blip.working.fetchingPatientData.inProgress,
+    fetchingUser: state.blip.working.fetchingUser.inProgress,
```
jh-bate: so we are using this for both fetching a `user` and `patient` ??

test/unit/app/app.test.js
```diff
@@ -225,8 +247,8 @@ describe('App',  () => {
         expect(result.user).to.equal(loggedIn.allUsersMap.a1b2c3);
       });
 
-      it('should retun the current patient in view as patient', () => {
-        expect(result.patient).to.equal(loggedIn.allUsersMap.d4e5f6);
+      it('should retun the current patient in view as patient and empty permissions', () => {
```
jh-bate: spelling of `retun `

app/pages/patientdata/patientdata.js
jh-bate: cancel that - dumb question :) have just looked two lines above

PR #392, opened 2017-02-22
restore VCA ability to change site change source if still custodian

jebeck:
Changes involved:
- membershipPermissionsInOtherCareTeams reducer handles FETCH_PATIENT_SUCCESS so that hard refresh works
- PatientData extracts permsOfLoggedInUser appropriately depending on whether logged-in user is the patient in view or not and passes it through to the Basics (goes with [tideline PR #306](https://github.com/tidepool-org/tideline/pull/306))
- tests for both of these changes, natch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,30 +1,16 @@
 sudo: required
 language: node_js
 
-services:
-  - docker
-
 node_js:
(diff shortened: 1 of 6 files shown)


jebeck:

PR #394, opened 2017-02-27
remove GET on (so far unused) `preferences` in `api.user.get`

jh-bate:
@jebeck adding this just so we have the details of the latest failure 

### Successful signup redirect

```
GET https://stg-api.tidepool.org/metadata/9f4b514ff4/preferences 404 (Not Found)
GET https://stg-api.tidepool.org/auth/user 401 (Unauthorized)
```

### Failed signup redirect

```
GET https://stg-api.tidepool.org/metadata/372109737a/profile 404 (Not Found)
```

404 from here https://github.com/tidepool-org/blip/blob/master/app/core/api.js#L188

which then causes the exception here 

https://github.com/tidepool-org/blip/blob/master/app/routes.js#L164

### DIFF
--- a/app/core/api.js
+++ b/app/core/api.js
@@ -208,14 +208,9 @@ api.user.get = function(cb) {
     });
   };
 
-  var getPreferences = function(cb) {
-    api.metadata.preferences.get(userId, cb);
-  };
-
   async.series({
     account: getAccount,
     profile: getProfile,
-    preferences: getPreferences,
   },
   function(err, results) {
     if (err) {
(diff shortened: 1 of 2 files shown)


jebeck:
It won't let me pick you as a reviewer @jh-bate maybe because you opened the branch? Have a look - discussed briefly w/@darinkrauss and if we understand the info you left (which we may not - we had some questions), removing the `preferences` call (which isn't used for anything yet) may provide the fix we need.

jebeck:
This seems like it was a red herring (especially the change I added), so closing.

PR #395, opened 2017-02-28
"blip" login + header & footer changes (rebranding)

jebeck:
I've tentatively scheduled a design review with SK on Wednesday, but I believe this is ready for someone to start looking at in terms of code. It's mostly styling changes with the most substantial component changes being the addition of the `FooterLinks` component separate from the version footer (since there was a request to hide the links in some places but keep the version always present for debugging/support purposes).

Looking for feedback on both how and how much to try to DRY out some of the styles (see TODO comments). Note that since the bulk of this work was done before our meeting re: the style guide project, I have not introduced new styles (e.g., using `flex` to position footer elements) via CSS modules, but I suppose getting a start on that could be one method of DRYing out what's here. Somewhat a timing question. Do we want to wait for the larger new sign-up flow tasks (and larger styles updates per style guide) to start on that?

Will update from master as soon as current v0.17.0 in test is approved for prod.

PR #396, opened 2017-03-08
"blip" login + header & footer changes (rebranding) + tag fix

jebeck:
This PR will take the place of #395 but please refer to there for PR comments and update history.

The problem is that I created a tag that already existed on the branch behind #395 and pushed the commit before I realized the mistake. (The complaining from Git happened when I tried to push the tag! I guess maybe it's better to do `git push origin --tags` first to fail early if the tag already exists, then you can reset and recreate the right tag.)

@hntrdglss can I get the required stamp of approval?


jebeck:
[<ux:1 dev:5 tst:1> [blip] Update Blip header and footer](https://trello.com/c/p5GATqC4/112-ux-1-dev-5-tst-1-blip-update-blip-header-and-footer)

PR #397, opened 2017-03-15
copy and paste settings

jh-bate:
@krystophv if you could take a quick look at this it would be good. Only a small change despite all the merges into the branch.


krystophv:
[It's hard to copy/paste Device Settings into an email AND edit them (i.e. add a new basal segment)](https://trello.com/c/ICFnscSF/115-it-s-hard-to-copy-paste-device-settings-into-an-email-and-edit-them-i-e-add-a-new-basal-segment)

PR #398, opened 2017-03-20
show print header in all situations

jh-bate:
@krystophv could you take a look at this please? FYI will be doing a design review with @skrugman also.

Here is the bug https://trello.com/c/n2ChaCHf


jh-bate:
[[BUG] Printing in Blip no longer prints PwD Name; no longer allows landscape](https://trello.com/c/n2ChaCHf/149-bug-printing-in-blip-no-longer-prints-pwd-name-no-longer-allows-landscape)

PR #399, opened 2017-03-22
Add customized blood glucose target range to patient profile

hntrdglss:
This adds a new section on the patient profile, where a DSA owner or custodian can set custom low and high bounds for target blood glucose, which will be reflected within `tideline/viz`*. See [Trello card](https://trello.com/c/Pax5chwE) for more info.

_*I have a [question out to Brandon](https://trello.com/c/Pax5chwE/81-customizable-target-range-3-of-3-customization-of-bg-prefs-ui-blip-profile-and-account-settings-ux-1-dev-5-tst-3#comment-58d18929404a934a3b0ff9c3) if that integration should happen as part of this card or another. If it should be part of this work, I'll branch off this and implement it in a separate PR._

PR #400, opened 2017-03-29
Integrate CBGTR into Tideline/Viz

hntrdglss:
Wires up `Tideline` and `Viz` with new customized blood glucose target range settings.

@krystophv and/or @jh-bate, once approved, I'll merge this into the pre-existing #399 pull request, then deploying that to `dev` and `stg` for testing.

### DIFF
--- a/app/pages/patientdata/patientdata.js
+++ b/app/pages/patientdata/patientdata.js
@@ -605,7 +605,9 @@ export let PatientData = React.createClass({
   doProcessing: function(nextProps) {
     var userId = this.props.currentPatientInViewId;
     var patientData = _.get(nextProps, ['patientDataMap', userId], null);
-    var patientSettings = _.get(nextProps, ['patient', 'settings'], DEFAULT_SETTINGS);
+    var patientSettings = _.get(nextProps, ['patient', 'settings'], null);
+    _.defaultsDeep(patientSettings, DEFAULT_SETTINGS);
+
     if (patientData) {
       let combinedData = patientData.concat(nextProps.patientNotesMap[userId]);
       window.downloadInputData = () => {
--- a/app/pages/patientdata/patientdata.js
+++ b/app/pages/patientdata/patientdata.js
@@ -605,7 +605,7 @@ export let PatientData = React.createClass({
   doProcessing: function(nextProps) {
     var userId = this.props.currentPatientInViewId;
     var patientData = _.get(nextProps, ['patientDataMap', userId], null);
-    var patientSettings = _.get(nextProps, ['patient', 'settings'], null);
+    var patientSettings = _.cloneDeep(_.get(nextProps, ['patient', 'settings'], null));
     _.defaultsDeep(patientSettings, DEFAULT_SETTINGS);
 
     if (patientData) {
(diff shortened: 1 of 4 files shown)


hntrdglss:
[Customizable Target Range (3 of 3) - customization of BG prefs UI [blip] [profile and account settings] <ux:1, dev:5, tst:3>](https://trello.com/c/Pax5chwE/81-customizable-target-range-3-of-3-customization-of-bg-prefs-ui-blip-profile-and-account-settings-ux-1-dev-5-tst-3)

app/pages/patientdata/patientdata.js
```diff
@@ -590,18 +592,20 @@ export let PatientData = React.createClass({
...
+    // nextProps patient data exists
+    if (!nextProps.fetchingPatient && !this.state.processedPatientData && nextPatientData) {
       this.doProcessing(nextProps);
     }
   },
 
   doProcessing: function(nextProps) {
     var userId = this.props.currentPatientInViewId;
     var patientData = _.get(nextProps, ['patientDataMap', userId], null);
+    var patientSettings = _.get(nextProps, ['patient', 'settings'], DEFAULT_SETTINGS);
```
krystophv: We might need to employ the `_.defaultsDeep` here as well. The `patient.settings` can be an empty object, and then `utils.processPatientData` gets angry about `settings.units.bg` not being defined.

hntrdglss: Ooh, good catch.

hntrdglss: Yup, was able to replicate. I originally was testing setting the defaults [here](https://github.com/tidepool-org/blip/blob/3a89fe4a8e84e0beef24e8b33497bcf51951e4aa/app/core/api.js#L347), but decided that was less obvious. I must have neglected to nuke my settings after moving it here. Thanks!

hntrdglss: For posterity, I also considered updating [this area](https://github.com/tidepool-org/blip/blob/3a89fe4a8e84e0beef24e8b33497bcf51951e4aa/app/core/api.js#L347), to set `person.settings = settings || *null*` instead of `{}`, and kept the above as-is. But that could affect the site change source code, and I didn't want to open that can of worms again. Might be something to consider in the future, once `platform` takes over the metadata endpoints, though!

app/pages/patientdata/patientdata.js
```diff
@@ -590,18 +592,22 @@ export let PatientData = React.createClass({
...
+    if (!nextProps.fetchingPatient && !this.state.processedPatientData && nextPatientData) {
       this.doProcessing(nextProps);
     }
   },
 
   doProcessing: function(nextProps) {
     var userId = this.props.currentPatientInViewId;
     var patientData = _.get(nextProps, ['patientDataMap', userId], null);
+    var patientSettings = _.get(nextProps, ['patient', 'settings'], null);
+    _.defaultsDeep(patientSettings, DEFAULT_SETTINGS);
```
krystophv: Heh, I think this ends up mutating an object via reference which is part of state. Maybe we need to clone it and then modify?

hntrdglss: Just pushed that, so give it a shot. In the meantime, I'll keep trying to reproduce on my side. Really curious what specific state might be causing that.

krystophv:
Fixes LGTM 👍

PR #402, opened 2017-04-12
filter out those accounts that aren't viewable

jh-bate:
@jebeck I believe this is all we need in this scenario. Local testing for me was good. Hoping that my comments in code code make it more understandable as we spoke about??


jh-bate:
[<M> [web] Make Patient List load faster](https://trello.com/c/VXQ5L102/146-m-web-make-patient-list-load-faster)

PR #403, opened 2017-04-13
fix display name to be fullname of pwd

jh-bate:
@jebeck discovered when resolving this "child account" issue https://trello.com/c/NMr0pTEF. We were displaying the account owners name not the associated patient (aka child account) name.

### DIFF
--- a/app/components/peopletable/peopletable.js
+++ b/app/components/peopletable/peopletable.js
@@ -76,9 +76,10 @@ class PeopleTable extends React.Component {
       if (bday) {
         bday = ` ${sundial.translateMask(bday, 'YYYY-MM-DD', 'M/D/YYYY')}`;
       }
+
       return {
-        fullName: personUtils.fullName(person),
-        fullNameOrderable: personUtils.fullName(person).toLowerCase(),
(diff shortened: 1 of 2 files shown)


jh-bate:
[<S> [hydrophone] update email so it sends the profile name for the PwD, not the User account name.](https://trello.com/c/NMr0pTEF/143-s-hydrophone-update-email-so-it-sends-the-profile-name-for-the-pwd-not-the-user-account-name)

jebeck:
LGTM

Noticed while testing locally that if you accept an invite to a new PWD's team on the VCA /patients page, then click "Show All," the invite you just accepted will not be in the list... Is this a known issue? This should be fixable perhaps with a second look at how we're storing and pulling data for the list in the redux store (i.e., may just require adding handling of the accept invite action to an existing reducer). Not suggesting it for this PR necessarily, just thought I'd mention it while we're noticing existing bugs ;)

jh-bate:
Ta, just taking a quick look now at the issue  to see if its a quick fix. If you search for the person they do show in the search but just seeing about showing in the list as soon as invite is accepted

PR #404, opened 2017-04-20
use viz's reshapeBgClassesToBgBounds function

jebeck:
What it says on the tin. Since this is such a small change, I figure we can merge whenever and just check that target range is set properly on Trends page in whatever release it goes out with. Lemme know if you disagree @krystophv


jebeck:
[<L> [blip] Print View - DAILY](https://trello.com/c/AJDUeLNL/153-l-blip-print-view-daily)

PR #405, opened 2017-04-20
updates for blip to be built/run with node v6.x LTS

jebeck:
Ping @jh-bate and @krystophv 

Can you think of anything I missed? Updated .travis.yml and "engines" in the package.json. Didn't look like we mentioned version of node anywhere in README

Not sure if we have `engines` in tideline and viz, but I'll check.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,18 @@ node_js:
   - "6.10.2"
   - "stable"
 
+addons:
+  artifacts:
+    s3_region: "us-west-2"
+    paths:
+    - $(git ls-files -o deploy/*/*-*.tar.gz | tr "\n" ":")
+    target_paths:
+    - /
+
 script:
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -1,12 +1,8 @@
-sudo: required
+sudo: false
 language: node_js
 
 node_js:
-  - "0.12.7"
-  - "stable"
```
jh-bate: so you don't think we should still keep `stable` for the future?

jebeck:
[<S> Node Update 1 - update Blip and its dependencies to node v6](https://trello.com/c/u82CBVgy/142-s-node-update-1-update-blip-and-its-dependencies-to-node-v6)

jh-bate:
testing all good so merging now cc @jebeck @pazaan

PR #406, opened 2017-04-21
Create deploy artifacts via Travis; update start script to use node version

darinkrauss:
@jebeck @jh-bate @pazaan Adds "deploy" process which builds "binary" artifact for later deployment. Update start script to utilize the targeted version of node. Uses shared environment variable to ensure artifact build and deployed execution use same exact version of node.

To move to a new version of node, we will need to: 1) update infrastructure configuration to automatically install new target version (outside of blip); 2) update `travis.yaml` here to include target version (under `node_js:`); and 3) update `deploy-env.sh` to specify target version used to build artifact and run service.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ _book
 node_modules
 npm-debug.log
 tmp
+deploy
 dist
 config/*
 !config/local.sh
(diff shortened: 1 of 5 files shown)


jebeck:
[<S> Node Update 1 - update Blip and its dependencies to node v6](https://trello.com/c/u82CBVgy/142-s-node-update-1-update-blip-and-its-dependencies-to-node-v6)

PR #407, opened 2017-04-21
Allow clinic role to specified on signup; update platform-client

darinkrauss:
@jebeck @jh-bate @krystophv This allows the signup URL to accept a new parameter `roles` (a comma-delimited array of strings, with `clinic` as the only valid value) that will automatically pass that to the create user call on the backend (`platform-client` has already been updated to pass `roles` through). The net result is that by specifying the `roles` parameter on the URL an account can be created that is immediately a VCA.

After this feature (and `shoreline`) is deployed, but prior to the new clinic signup workflow deployment, we will update the signup TypeForm to allow the user to immediately signup (with the `inviteKey` and optional `roles` parameter depending upon how they answer some of the TypeForm questions). There will be a redirect involved (with the redirect service in our infrastructure).

### DIFF
--- a/app/core/utils.js
+++ b/app/core/utils.js
@@ -24,7 +24,7 @@ var utils = {};
 
 /**
  * Convenience function for capitalizing a string
- * 
+ *
  * @param  {String} str
  * @return {String}
  */
@@ -104,10 +104,10 @@ utils.objectDifference = (destination, source) => {
 
 /**
  * Utility function to get whether page has changed or not
- * 
+ *
  * @param  {Object} oldProps
  * @param  {[type]} newProps
- * 
+ *
  * @return {Boolean}
  */
 utils.isOnSamePage = (oldProps, newProps) => {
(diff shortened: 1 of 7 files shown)


darinkrauss:
[<S> [backend] Sign Up Flow - Phase 1a - Clinician - Back End (2 of 3) - VCA Create](https://trello.com/c/kl2S776E/123-s-backend-sign-up-flow-phase-1a-clinician-back-end-2-of-3-vca-create)

app/core/utils.js
```diff
@@ -196,6 +196,21 @@ utils.getInviteKey = function(location) {
   return '';
 }
 
+utils.getRoles = function(location) {
```
jebeck: We've got tests around other things in this utils module. Could you add some tests around this @darinkrauss? They are in test/unit/utils/utils.test.js it looks like. Will also be helpful as "documentation" since there's a lot going on in your lodash filtering!

darinkrauss: @jebeck Good call. I added tests for `getRoles`. While I was at it, I added tests for `getInviteKey`.

darinkrauss:
@jebeck Added the metric properties that Brandon asked for. Checkout the changes to `trackingMiddleware.js` and let me know if you like where it ended up.

jebeck:
@darinkrauss Travis has choked on a linting error with your more recent changes...

jebeck:
LGTM as soon as Travis passes w/the lint fix.

PR #408, opened 2017-04-22
Remove TOU and PP summary, require acceptance of new terms when they are updated

pazaan:
Add support for an environment variable called `LATEST_TERMS` which should be a string that is parsable by JavaScript Date(), preferably in ISO 8601 combined format so as to be unambiguous.
This should be the date of the most recent Tidepool Terms of Use. When the date is more recent than a users’ ToU acceptance date, the ToU terms page will be shown and the user will need to re-accept the terms before being able to continue to the Tidepool web app.

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -15,9 +15,13 @@
 
 /* global __TEST__ */
 
+if (__TEST__ && (window === undefined || window.config === undefined)) {
+  window.config = {};
+}
+
 var config = window.config;
 
-if (!config && !__TEST__) {
+if (!config) {
   throw new Error('Expected `config` on the global `window` object');
 }
 
--- a/.eslintrc
+++ b/.eslintrc
@@ -19,7 +19,8 @@
     "console": false,
     "__UPLOAD_API__": false,
     "__API_HOST__": false,
-    "__INVITE_KEY__": false
+    "__INVITE_KEY__": false,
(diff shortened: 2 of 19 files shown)


pazaan:
[Remove TOU and PP summary, update date, require acceptance of new terms](https://trello.com/c/gHVbaisM/162-remove-tou-and-pp-summary-update-date-require-acceptance-of-new-terms)

jh-bate:
@pazaan looks like the tests are failing:

see `https://travis-ci.org/tidepool-org/blip/jobs/224684043#L2302`

jebeck:
I just had a few minutes to check in. @pazaan the tests are failing because you have not based this work on the node update. As soon as the node update #405 is approved for prod it needs to be merged to master and this work needs to merge in the update from master. You may also have failing tests after that, as it doesn't look like you've updated any tests yet and there _should_ be existing tests around both the personutils and the terms "page."

app/config.js
```diff
@@ -15,9 +15,13 @@
 
 /* global __TEST__ */
 
+if (__TEST__ && (window === undefined || window.config === undefined)) {
```
jh-bate: @jebeck this was the update that allowed the tests to work. `personutils.js` now has a reference to config and that was killing the tests.

jh-bate:
approved so now merging

PR #412, opened 2017-05-01
Copy settings metrics fix

jebeck:
OK, if there's no other changes required since @darinkrauss's sign-up metrics are working, then this is the only metrics fix that needs to go out, right?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "blip",
-  "version": "1.3.2",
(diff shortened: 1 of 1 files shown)


jebeck:
[[viz] BUG: View Settings for Tandem device hangs on v0.7.14](https://trello.com/c/Tn72jJMW/165-viz-bug-view-settings-for-tandem-device-hangs-on-v0-7-14)

PR #413, opened 2017-05-02
standardize on passing through bgPrefs with bgUnits and bgBounds inside

jebeck:
What it says on the tin. Depends on [viz #71](https://github.com/tidepool-org/viz/pull/71).


jebeck:
[<L> [blip] Print View - DAILY](https://trello.com/c/AJDUeLNL/153-l-blip-print-view-daily)

PR #414, opened 2017-05-03
changes to use Yarn in blip

jebeck:
What it says on the tin: includes checking in yarn.lock, README update, and directing Travis to cache Yarn.

Turns out that `yarn install` will fail hard on a mismatch between node version (see [this Travis build](https://travis-ci.org/tidepool-org/blip/jobs/228490914)) being used and node engine specified in a dependency, so this PR is dependent on [viz #73](https://github.com/tidepool-org/viz/pull/73).


jebeck:
Welp, check out the Travis build @krystophv 

Yarn has errored with: `Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.` when I included a yarn.lock update in the last commit. I can only assume this is due to platform differences: I am running Yarn 0.23.4 on my Mac, and Travis has 0.24.4 on Linux...

(Ooooh, the plot thickens: `brew update` + `brew upgrade yarn` just let me install 0.24.4 on my Mac, so will try generating new yarn.lock and committing...but needing an exact version match between dev boxes and CI boxes seem unsustainable in the long term, potentially...)

What do we think about `--pure-lockfile` instead? I guess it's risky because your package.json and yarn.lock could be out of sync... There seems to be no easy solution here.

jebeck:
Update: as @krystophv has pointed out, the failure that I blamed on different versions of Yarn producing different lockfiles was actually due to a FAIL on my part - I updated the package.json version of @tidepool/viz without regenerating the lockfile in b503c1d05ce16603ce6bf596f1a9291a48120b76...oops. So we'll proceed with `--frozen-lockfile` for replicable builds on Travis.

PR #415, opened 2017-05-15
fix merge issue and add tests so it won't occur again

jh-bate:
@krystophv could you take a look at this please. It resolves the issue seen here https://trello.com/c/Q15cxITp which was the result of a merge that I made.

I have added a test that now asserts that the header is present and will fail if another issue happens in the same manner (comment out the renderHeader call and then the test will fail)

### DIFF
--- a/app/components/printheader/printheader.js
+++ b/app/components/printheader/printheader.js
@@ -29,8 +29,8 @@ const PrintHeader = React.createClass({
     dateLink: React.PropTypes.string,
   },
   render: function() {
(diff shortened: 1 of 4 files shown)


jh-bate:
[[BUG] Header info regression on Device Settings Print View](https://trello.com/c/Q15cxITp/167-bug-header-info-regression-on-device-settings-print-view)

jh-bate:
approved for `prd` so now merging

PR #417, opened 2017-05-18
adding bddp option to DSA signup

krystophv:
Adds the option to share data with the BDDP and specify one of a number of nonprofit partners with which to attribute the proceeds.

This necessitated the addition of a `renderSelect` addition to the `SimpleForm` system, as it previously had no notion of select elements.

### DIFF
--- a/app/components/inputgroup/inputgroup.js
+++ b/app/components/inputgroup/inputgroup.js
@@ -197,7 +197,6 @@ var InputGroup = React.createClass({
   },
 
   renderSelect: function() {
-    var self = this;
     var placeholder = this.props.value === this.props.placeholder;
     var classNames = cx({
       'input-group-control': true,
@@ -224,7 +223,7 @@ var InputGroup = React.createClass({
           name={this.props.name}
           value={this.props.value}
           id={this.props.name}
-          onChange={self.handleChange}
+          onChange={this.handleChange}
           disabled={this.props.disabled}>
             {options}
         </select>
--- a/app/components/inputgroup/inputgroup.js
+++ b/app/components/inputgroup/inputgroup.js
@@ -16,6 +16,7 @@
 
 var React = require('react');
 var _ = require('lodash');
+var cx = require('classnames');
 
 var DatePicker = require('../datepicker');
 
@@ -25,7 +26,7 @@ var InputGroup = React.createClass({
     name: React.PropTypes.string,
     label: React.PropTypes.string,
     items: React.PropTypes.array,
-    text: React.PropTypes.string,
+    text: React.PropTypes.node,
     value: React.PropTypes.oneOfType([
       React.PropTypes.string,
       React.PropTypes.bool,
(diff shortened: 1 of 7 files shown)


krystophv:
[Improve the experience sharing with Big Data Donation upon Tidepool signup](https://trello.com/c/ZnKDZhzv/178-improve-the-experience-sharing-with-big-data-donation-upon-tidepool-signup)

app/components/inputgroup/inputgroup.js
```diff
@@ -191,6 +196,42 @@ var InputGroup = React.createClass({
     );
   },
 
+  renderSelect: function() {
+    var self = this;
+    var placeholder = this.props.value === this.props.placeholder;
```
jh-bate: why not just use the defined `self` from above in all instances where we are still using `this`?

app/pages/patientnew/patientnew.js
```diff
@@ -70,6 +71,37 @@ export let PatientNew = React.createClass({
...
+        {value: 'T1DX', label: 'T1D Exchange'}
+      ]
+    },
+    {
+      name: 'donateExplainer',
+      type: 'explanation',
+      text: (<div style={{'textAlign':'left'}}>
+        By securely and anonymously donating your diabetes data you are helping
+        researchers, device makers, and other innovators. Read everything about
+        this project <a href="http://tidepool.org/blog">here</a>.
```
jh-bate: should this point to the actual data donation post?

test/unit/pages/patientnew.test.js
```diff
@@ -23,6 +23,7 @@ describe('PatientNew', function () {
       console.error = sinon.spy();
       var props = {
         fetchingUser: false,
+        onInviteMember: sinon.stub(),
```
jh-bate: should we have some test that assert we trigger this when data is donated and the member invited?

app/components/inputgroup/inputgroup.js
krystophv: actually, since none of the places are using the `self` in the way that it's classically used (in order to get around internal function scope) it's probably better to not define a `self` here at all - will fix.

app/pages/patientnew/patientnew.js
krystophv: It will. The post isn't up yet and, thus, the URL isn't finalized. Likely will come in sometime tomorrow, but the text change should be easy to get through without much fuss.

test/unit/pages/patientnew.test.js
krystophv: Yup. Incoming.

krystophv:
@jh-bate adjustments/updates made

app/pages/patientnew/patientnew.js
jh-bate: ok

app/components/inputgroup/inputgroup.js
jh-bate: yeah thats better

PR #422, opened 2017-05-29
Daily print view header

cbwebdevelopment:
Passing patient data through the PDF print handler for the daily print view.  To be reviewed alongside the clintb/daily-print-view branch in the `viz` repo.


cbwebdevelopment:
[<L> [blip] Daily print view first production release](https://trello.com/c/tjD655oz/182-l-blip-daily-print-view-first-production-release)

jebeck:
@cbwebdevelopment I'm going to update jebeck/daily-print-view from master to reduce the weird changeset showing up in the diff here... Probably doesn't affect you, but thought I'd give you a heads up.

jebeck:
OK, it did affect you... When you get a chance, you'll need to resolve the conflict now appearing here.

cbwebdevelopment:
Conflicts resolved :)

cbwebdevelopment:
@jebeck There are a few fairly significant changes to the webpack config with the last commit.

I felt that the static pdf-kit.js and blob-stream.js files should be loaded in via Webpack `copy-webpack-plugin` and `html-webpack-include-assets-plugin` to use it's cache-busting and minification.  

This required changing the way the assets are loaded into the main index.html file, which, is now using the`html-webpack-plugin` package.  This simplifies the buildapp.js script significantly, though the custom hashing strategy employed there is now replaced by webpack's, for better or for worse.  

Let me know if you see any red flags here.

jebeck:
@cbwebdevelopment would you be comfortable going back in your branch a bit and creating some new branches and new PRs? I think the Webpack changes need to be reviewed separately, and I don't feel comfortable doing that. I've never fully understood, for example, our custom hashing strategy, but I do know that when we added it was crucial to getting the app to load the correct configs (note that there are two bundles in production: an app bundle, and a config bundle). I'd rather focus on reviewing the PDF code before the end of the week, and I'd advise getting @krystophv to look at the build changes later.

jebeck:
On second thought, I could just review your changes w/o looking at the Webpack stuff. Then just make sure that @krystophv reviews the Webpack commit(s) when the final review of `jebeck/daily-print-view` for deploy & merge into `master` is done.

PR #425, opened 2017-06-07
Update build, test, artifact process

darinkrauss:
@jh-bate


.travis.yml
```diff
@@ -1,31 +1,37 @@
 sudo: false
+
 language: node_js
 
 node_js:
-  - "6.10.2"
-  - "stable"
+  - 6.10.2
+  - node
```
jh-bate: what is just the `node` tag mean??

.travis.yml
```diff
@@ -1,31 +1,37 @@
...
 
 cache: yarn
 
 before_install:
-  - if [[ `npm -v` != 4* ]]; then npm i -g npm@4; fi
+  - if [[ `npm -v` != 4* ]]; then npm install -g npm@4; fi
```
jh-bate: just  a sanity check - its intended we use `npm` for the build and not `yarn`??

.travis.yml
darinkrauss: According to https://docs.travis-ci.com/user/languages/javascript-with-nodejs/, `node` is the "latest stable Node.js release". The same as `tip`, I believe, but it is explicitly *not* io.js.

.travis.yml
darinkrauss: I didn't write this. Jana was the one who updated to `yarn`, but only some of the stuff in the `.travis.yml` and `build.sh` got updated. I don't know the reason behind this, but I am just maintaining the status-quo. If someone else wants to look into whatever the "right thing" is, that'd be great, but it wouldn't be part of this change.

.travis.yml
jh-bate: ok

.travis.yml
jh-bate: yeah I had a recollection of that just wanted to check

PR #427, opened 2017-06-17
Daily Print View

cbwebdevelopment:
This allows the daily chart data for the most recent 6 days to be printed from Tidepool Web when viewing the Daily view.

Goes hand-in-hand with corresponding `clintb/daily-print-view.web-worker` branch on viz.

Details at https://trello.com/c/tjD655oz


cbwebdevelopment:
[<L> [blip] Daily print view first production release](https://trello.com/c/tjD655oz/182-l-blip-daily-print-view-first-production-release)

webpack.config.js
```diff
@@ -19,10 +24,32 @@ var defineEnvPlugin = new webpack.DefinePlugin({
...
+      }
+    }
+  ]),
+  new HtmlWebpackPlugin({
+    template: 'index.ejs',
+  }),
+  new HtmlWebpackIncludeAssetsPlugin({
+    assets: ['pdfkit.js', 'blob-stream.js'],
+    hash: true,
+    append: true,
```
krystophv: Should the external assets here be included before `bundle.js` (i.e. using `append: false`) to ensure that they're available on the page before the application code attempts to run?

app/components/chart/header.js
```diff
@@ -15,21 +15,24 @@
...
+import printPng from './img/print-icon-2x.png';
 
-var tideline = {
-  log: bows('Header')
+const tideline = {
+  log: bows('Header'),
 };
 
-var TidelineHeader = React.createClass({
+const TidelineHeader = React.createClass({
```
cbwebdevelopment: Will do.  Wasn't sure if I was to take on converting it over, though I agree, if I'm going to ES6-ify some of the file, might as well go all the way :)

app/components/chart/header.js
```diff
@@ -146,7 +155,8 @@ var TidelineHeader = React.createClass({
         <div className="grid-item one-whole large-one-third">
           <a href="" className={settingsLinkClass} onClick={this.props.onClickSettings}>Device settings</a>
           <a href="" className={printLinkClass} onClick={this.props.onClickPrint}>
-            <img src={printPng} alt="Print" />
+            {this.props.printReady && <img className="print-icon" src={printPng} alt="Print" />}
+            {!this.props.printReady && <Loading className="print-loading-spinner" delay={0} type="spin" color="#fff" />}
```
cbwebdevelopment: Good catch.

app/pages/patientdata/patientdata.js
```diff
@@ -596,6 +627,20 @@ export let PatientData = React.createClass({
     }
   },
 
+  componentWillUpdate: function(nextProps, nextState) {
+    const pdfEnabled = _.indexOf(['daily'], this.state.chartType) >= 0;
+    const pdfGenerating = nextProps.generatingPDF;
+    const pdfGenerated = nextProps.viz && nextProps.viz.pdf && nextProps.viz.pdf[this.state.chartType];
```
cbwebdevelopment: Agree, `_.get()` is a nice solution here.

test/unit/pages/patientdata.test.js
```diff
@@ -465,30 +468,209 @@ describe('PatientData', function () {
...
+          }
+        },
+        generatingPDF: true,
+      };
+
+      const wrapper = mount(<PatientData {...props} />);
+      const elem = wrapper.instance();
+      sinon.stub(elem, 'generatePDF');
+
+      var callCount = elem.generatePDF.callCount;
```
cbwebdevelopment: good catch 👍

app/pages/patientdata/patientdata.less
```diff
@@ -100,6 +100,15 @@
   float: right;
 }
 
+.patient-data-subnav .print-loading-spinner {
+  width: 16px !important;
```
cbwebdevelopment: Yep, I felt dirty putting that in.  For some reason I thought react-loader didn't allow it to be set via props (should've looked at the gh page instead of a code example from an old project of mine ;) )

webpack.config.js
cbwebdevelopment: In this case, these assets just need to be in place by the time the web worker is invoked to create a print pdf, which is when a user navigates to the daily view.

Technically, it doesn't even need to be loaded by the main thread.  The only reason I have it being loaded there at all is for cache-busting, forcing the browser to reload it when we deploy with a new hash.

I recommend we keep the bundle.js loading first for performance.

cbwebdevelopment:
@krystophv I've updated the PR to address your feedback

PR #430, opened 2017-06-26
fix Basics unmount causing state data restore

krystophv:
See https://trello.com/c/3HKSQKm5 for full context.

This prevents the unmount of the Basics chart causing a setState with stale patient data which will ultimately prevent the app from processing incoming new data (and move beyond the Loading screen).


krystophv:
[Refresh Button: Data Load hangs forever](https://trello.com/c/3HKSQKm5/195-refresh-button-data-load-hangs-forever)

PR #436, opened 2017-08-08
launch Electron app based on custom URI handling

krystophv:
Mostly what it says on the tin. Of note, the callbacks for the protocol detect command (which actually launches the protocol) aren't used because I doubt that they're particularly robust/reliable. For instance, neither are ever called on Opera.

Pairs well with: https://github.com/tidepool-org/chrome-uploader/pull/457


krystophv:
[Electron Uploader - Blip Integration](https://trello.com/c/CTjNregM/212-electron-uploader-blip-integration)

krystophv:
The relevant Electron release that will register itself as the custom protocol handler is here: https://github.com/tidepool-org/chrome-uploader/releases/tag/v0.310.0-alpha.22

test/unit/components/uploadlaunchoverlay.test.js
```diff
@@ -0,0 +1,91 @@
...
+      });
+      expect(wrapper.find('a')).to.have.length(2);
+      expect(wrapper.find('a.disabled')).to.have.length(0);
+    });
+
+    it('should display download link if error retrieving github releases', () => {
+      wrapper.setState({
+        error: 'some error'
+      });
+      expect(wrapper.find(`a[href=${URL_UPLOADER_DOWNLOAD_PAGE}]`)).to.have.length(0);
```
cbwebdevelopment: Should this not be expecting the length to be 1?  Just setting it to 1 would fail, though, as there's an issue with enzyme and href matching, so you'll need to change it. see https://github.com/airbnb/enzyme/issues/283#issuecomment-278086218

Changing to this, for instance, passes:
`expect(wrapper.find({ href: URL_UPLOADER_DOWNLOAD_PAGE }).filter('a')).to.have.length(1);`

app/components/uploadlaunchoverlay/uploadlaunchoverlay.js
```diff
@@ -0,0 +1,123 @@
...
+
+const github = new GitHub();
+
+class UploadLaunchOverlay extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      latestWinRelease: null,
+      latestMacRelease: null,
+      error: null
```
krystophv: I don't think trailing comma is part of our established code style.

krystophv: But I do see that they're part of Airbnb's eslint, so in the interest of moving that direction, I'll tack them in

app/components/navbarpatientcard/navbarpatientcard.js
```diff
@@ -118,16 +122,19 @@ var NavbarPatientCard = React.createClass({
...
       if (e) {
         e.preventDefault();
+        e.stopPropagation();
       }
-      window.open(self.props.uploadUrl, '_blank');
+      self.setState({showUploadOverlay: true});
+      protocolDetect('tidepoolupload://open',
+        () => { /* error callback */ },
+        () => { /* success callback */ }
+      );
```
krystophv: Seems reasonable.

app/components/uploadlaunchoverlay/uploadlaunchoverlay.js
```diff
@@ -0,0 +1,123 @@
...
+    uploaderRepo.listReleases((err, releases, request) => {
+      if(err){
+        this.setState({error: true});
+      }
+      this.setState(this.getLatestGithubRelease(releases));
+    });
+  }
+
+  getLatestGithubRelease = (releases) => {
+    const latestRelease = _.filter(releases, {prerelease: false})[0];
```
krystophv: I had a similar thought, but realized that it wouldn't be able to generate any links at all until we had an Electron version released. Since we _should_ always wait until all the relevant artifacts are attached to a GitHub release before attaching the tag and publishing it, I feel like it's a relatively low risk to presume that the URL exists. If human error creates a situation in which this becomes an issue in the future (and we have releases which have these artifacts at all), then I think we could add some additional checks.

test/unit/components/uploadlaunchoverlay.test.js
krystophv: Good catch. I copied the `.length(0)` logic from some other tests which seem to use it as a verification that the render for a component doesn't error for `shallow`. (I'll have to double check that that actually does what those tests are expecting now).

krystophv:
@cbwebdevelopment - there was a semi-significant set of changes adding the direct download links to the nodata screen as well. as such, I moved the release extraction function to `utils`. Could you look over this and give a review of the changes?

cbwebdevelopment:
The new changes LGTM 👍

PR #445, opened 2017-10-04
restore dev tools by default 

jebeck:
i.e., when `DEV_TOOLS` env variable is unset/undefined.

Unless y'all made a conscious decision to disable them by default? In which case (but I suspect not since I found the regression [here](https://github.com/tidepool-org/blip/commit/dc7ec489699fb5f12bdd285c669f728c674c264f#diff-11e9f7f953edc64ba14b0cc350ae7b9d)), just lemme know and I'll do a PR to update the README for you.

(I have an ulterior motive for fixing this: a small 🐛  that crashes the app in dev mode only (due to the mutation tracker). So look for that fix soon 😉 )


cbwebdevelopment:
[Restore dev tools by default in Blip](https://trello.com/c/1tLMULDP/265-restore-dev-tools-by-default-in-blip)

PR #446, opened 2017-10-18
Fix for displaying inaccurate 'High' and 'Low' smbg values

cbwebdevelopment:
Pulling in new versions of the `tideline` and `viz` repos.

See https://trello.com/c/p7L3RoNn for details


cbwebdevelopment:
[HI/LO SMBG shows actual value instead](https://trello.com/c/p7L3RoNn/19-hi-lo-smbg-shows-actual-value-instead)

PR #447, opened 2017-10-24
Distinguishing bolus override up from override down in basics view

cbwebdevelopment:
On the basics view, remove the 'Manual' bolus filter, and separate the the 'Overrides' filter into 'Overrides' and 'Underrides'

see https://trello.com/c/pQaZdjad for full details.

Pulling in changes from https://github.com/tidepool-org/tideline/pull/320

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "blip",
-  "version": "1.6.12",
+  "version": "1.6.13",
   "private": true,
   "scripts": {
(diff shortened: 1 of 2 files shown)


cbwebdevelopment:
[[data viz] [basics] distinguish override up from override down](https://trello.com/c/pQaZdjad/190-data-viz-basics-distinguish-override-up-from-override-down)

PR #450, opened 2017-11-20
Print View - Combined (Basics, Daily, Settings)

cbwebdevelopment:
See https://trello.com/c/zdyHFY0M for full details.

Goes hand-in-hand with the same-named branches in `tideline` and `viz`.

This PR provides the logic for kicking off the new, combined print PDF's from the basics, daily, and settings view.  (Previously, the PDF was only generated when on the daily view)

It also removes the old settings print components, styling, and tests, as this print view is now completely rewritten in `viz`.

One other notable change is that the `window.downloadDailyPrintViewData` function that was previously used on the daily view to create a json file for local storybook testing on viz has been replaced with a generic `downloadPrintViewData` method that creates and downloads the data for all 3 print views, in both moll/L and mg/dL, into one single file.

NOTE: I haven't yet pulled the new versions of `tideline` and `viz` into this repo, as I was going to wait until those PR's were approved before publishing them on npm.  At that point, I will add them to the package.json here.  If you need to run this locally, I'd recommend checking out the `blip`, `tideline`, and `viz` repos to their respective `clintb/basics-print-view` branches, and npm linking `tideline` and `@tidepool/viz`.


cbwebdevelopment:
[<L> [blip] Print View - Combined (Basics, Daily, Settings)](https://trello.com/c/zdyHFY0M/154-l-blip-print-view-combined-basics-daily-settings)

app/components/chart/basics.js
```diff
@@ -188,6 +192,21 @@ var Basics = React.createClass({
...
+      e.preventDefault();
+    }
+
+    if (this.props.pdf.url) {
+      const printWindow = window.open(this.props.pdf.url);
+      printWindow.focus();
+      printWindow.print();
+    }
+
+    // Send tracking metric
```
jh-bate: maybe the prop name should change to reflect that is firing a metric?? just noting the comment

app/components/chart/settings.js
```diff
@@ -149,37 +155,51 @@ const Settings = React.createClass({
...
       e.preventDefault();
     }
+
+    if (this.props.pdf.url) {
+      const printWindow = window.open(this.props.pdf.url);
+      printWindow.focus();
+      printWindow.print();
+    }
+
+    // Send tracking metric
```
jh-bate: same as pervious comment

app/pages/patientdata/patientdata.js
```diff
@@ -498,7 +506,7 @@ export let PatientData = React.createClass({
     });
   },
 
-  handleSwitchToDailyPrintView: function() {
+  handleSwitchToPrintView: function() {
```
jh-bate: based on comments in the code you should rename this

cbwebdevelopment:
@jh-bate I've renamed the print handlers in the basics, settings, as well as the daily view, and also moved the window opening/focusing logic to the top-level handler in `patient-data.js`

app/components/chart/basics.js
cbwebdevelopment: done

app/components/chart/settings.js
cbwebdevelopment: done

app/pages/patientdata/patientdata.js
cbwebdevelopment: done

PR #451, opened 2017-11-24
Only volume-mount the bare minimum of files docker image.

cbwebdevelopment:
Goes hand-in-hand with:
https://github.com/tidepool-org/tools/pull/64
https://github.com/tidepool-org/viz/pull/93

Changes to Dockerfile to allow us to not volume-mount the `node_modules` and `dist` directories to improve local performance.


cbwebdevelopment:
[Docker performance improvements for 'blip' and 'viz' services](https://trello.com/c/gJMqJ9K6/287-docker-performance-improvements-for-blip-and-viz-services)

cbwebdevelopment:
You're right, @pazaan, there's no good reason for this.  I've removed the unnecessary docker volume definitions.

PR #452, opened 2017-11-28
Dexcom connect banner

cbwebdevelopment:
This displays a 'connect to dexcom' banner in the patient data views.

See https://trello.com/c/YzpXRsNJ for full details.

SPECIAL NOTES:
Due to exceptional circumstances, we are performing QA and the PR review at the same time.  Also, the unit tests for the new functionality are not yet created, but will be while the PR review and staging is happening.

Also, there is code in place to display a 'Connect to Dexcom' button on the banner which has been intentionally commented out for the time being.  We will be enabling that in a later release, but we've decided that we should add a help screen in the initial dexcom connect popup before we do that.


cbwebdevelopment:
[Dexcom API messaging / banner](https://trello.com/c/YzpXRsNJ/278-dexcom-api-messaging-banner)

PR #454, opened 2017-11-30
Print view hotfix

cbwebdevelopment:
Accidentally pushed a prod release with the wrong viz version for the print views.  Simply reverting to the correct viz version


cbwebdevelopment:
[Bug: Print option fails to load with latest release](https://trello.com/c/G3u35uCo/291-bug-print-option-fails-to-load-with-latest-release)

PR #455, opened 2017-12-01
Daily y-scale fix for data sets with no values above target range

cbwebdevelopment:
Data with no above-target values causes Y scale on daily view to be off (the very high boundary is rendered outside of the BG chart area)

See https://trello.com/c/1r7IxaiA for details.

Related PRs
https://github.com/tidepool-org/tideline/pull/324
https://github.com/tidepool-org/viz/pull/94


cbwebdevelopment:
[Data with no above-target values causes Y scale on daily view to be off.](https://trello.com/c/1r7IxaiA/289-data-with-no-above-target-values-causes-y-scale-on-daily-view-to-be-off)

PR #456, opened 2017-12-08
Fix error notification on VCA-created patient account verification

cbwebdevelopment:
See https://trello.com/c/HKrQyncg for details.

In a nutshell, the `getDataSources` fetcher in the main `app.js` was being called before the user was actually authenticated, causing an error message to pop up during the verification process.

This PR just ensures we don't fire that fetcher if the user is not authenticated.


cbwebdevelopment:
[Error when home verifying a VCA-created account](https://trello.com/c/HKrQyncg/298-error-when-home-verifying-a-vca-created-account)

PR #460, opened 2018-01-23
Production tags for Libre fixes

cbwebdevelopment:
Forgot to create production tags before merging branch to master.  Need to do it in separate PR now.


cbwebdevelopment:
[Libre data not being calculated due to insufficient data points](https://trello.com/c/FNlDqd1A/11-libre-data-not-being-calculated-due-to-insufficient-data-points)

PR #461, opened 2018-01-24
Ease Life of developers working on windows hosts/linux guests

franck-fourel:
Add --watch-poll option for webpack-dev-server
This will allow people on Windows to use Linux Vbox guest while code is on windows host,
Small enhancement that may help people on Windows making development. I f you do not have this option the changes are not automatically uploaded and you have to restart the node service after each change.
Still need to validate this change on other platforms.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "nightwatch-single": "./node_modules/nightwatch/bin/nightwatch",
     "browser-tests": "NODE_ENV=test ./node_modules/karma/bin/karma start --browsers Chrome",
     "test-watch": "NODE_ENV=test ./node_modules/karma/bin/karma start --no-single-run",
-    "start": "NODE_ENV=development ./node_modules/webpack-dev-server/bin/webpack-dev-server.js --cache --colors --progress --port 3000 --host 0.0.0.0",
+    "start": "NODE_ENV=development ./node_modules/webpack-dev-server/bin/webpack-dev-server.js --cache --watch-poll --colors --progress --port 3000 --host 0.0.0.0",
     "build": "NODE_ENV=production npm run build-app && npm run build-config",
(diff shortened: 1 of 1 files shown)


cbwebdevelopment:
@franck-fourel Thanks for taking the time to prepare this!  

I'll try to carve out some time this week to look it over and test it out in a few environments.

package.json
```diff
@@ -8,7 +8,7 @@
     "nightwatch-single": "./node_modules/nightwatch/bin/nightwatch",
     "browser-tests": "NODE_ENV=test ./node_modules/karma/bin/karma start --browsers Chrome",
     "test-watch": "NODE_ENV=test ./node_modules/karma/bin/karma start --no-single-run",
-    "start": "NODE_ENV=development ./node_modules/webpack-dev-server/bin/webpack-dev-server.js --cache --colors --progress --port 3000 --host 0.0.0.0",
+    "start": "NODE_ENV=development ./node_modules/webpack-dev-server/bin/webpack-dev-server.js --cache --watch-poll --colors --progress --port 3000 --host 0.0.0.0",
```
cbwebdevelopment: I've tested it out on my Mac (in an alpine linux Docker container, as that is our recommended development environment), and the idle CPU % for the Docker `hyperkit` process went from 2% to 200% on average.

I understand it's needed for Windows users, and this is something we want to accommodate, but I believe a better approach will be to make this configurable via an environment variable in `webpack.config.js`

Adding the following to the `devServer` config object should do it:

```
watchOptions: {
  poll: process.env.WEBPACK_POLL || false,
}
```

cbwebdevelopment: Note: if running in Docker, make sure to add `WEBPACK_POLL:true` to the `environment` block for the `blip` service in your `docker-compose.yml` file.

franck-fourel:
outdated

Issue #449, opened 2017-11-13
Localize the interface


mgu:
It would be great if the strings were translatable.

coyotte508:
I've worked towards that (I'm not part of the tidepool team, or related to the organization) in my own fork.

The localization proposal is [here](https://github.com/coyotte508/blip/issues/2). My fork also acts as a proof of concept where the navbar is translated to French, using the process described in the proposal.

I welcome any feedback, to see if the way it is described is acceptable, and if I should push this further, towards adding language settings in the user interface.

franck-fourel:
@pazaan @cbwebdevelopment 
have you had a chance to have a look at what @coyotte508 (aka Eliott) proposed for multi-language?

pazaan:
This looks good to me. Did you investigate something like react-i18next, considering that it also supports date localisation and plurals? It looks like it might be easy to do this with `marker: 'react-i18next.t'` if we want to add it later...
@krystophv - do you have any comments on this approach?

coyotte508:
The current proposal was designed with minimal changes so it would be more readily acceped, using react-i18next would introduce more changes:

- Using the [translate hoc](https://react.i18next.com/components/translate-hoc.html) on each component, so that it can use `t`
- In each component's code, `t` needs to be extracted from its properties to be used 
- react-i18next recommends using key names, eg `t('welcome.message')` instead of `__('Welcome to Tidepool')`, although the second approach is possible. Using the first approach would be a massive rewrite of the code.

Dates can be handled easily in my opinion, as blip already uses moment for date display. The only thing to do would be to set the language of moment at the same time as setting the language of the application.

Regarding plurals, I don't see it becoming an issue. I didn't see any pluralization dependent on a variable number in the code. If there is a specific example in the code, I can look at it. Depending on how it's done, it can be gracefully handled by the current proposal.

react-i18next does offer more utilities, so if you are okay with the additional changes entailed, I can switch to using it.

By the way, the current code in my repo introduced a language setting in the user interface. I was pleasantly surprised that no change to server code was needed to preserve the language settings across logins. Now the language is set to English by default, and can be changed to French in the Account settings.

krystophv:
[...]

PR #468, opened 2018-03-22
Add multi-language capability to blip with react-i18next

coyotte508:
Related issue: https://github.com/tidepool-org/blip/issues/449

As a proof of concept, the front page, the navbar and the account settings page have translations added for French. Language can be changed in the account settings page.

The workflow changes are described below:

## Extracting translation keys from code

Just run `npm run update-translations`. It uses [i18next-parser](https://github.com/tidepool-org/blip/issues/449) under the hood.

The json files localted in `locales/xx` are updated.

## Translating components

There are two steps: wrapping the component with `translate(namespace?, options?)()` from `react-i18next`, and extracting `t` from `this.props`.

When using `React.createClass`: 
```jsx
import React from 'react';
import { translate } from 'react-i18next';

const Component = translate()(React.createClass({
  // ...
  render: function() {
    // t will translate based on the current language and  namespace given to translate(), 
    // or the default namespace if no namespace was given to translate()
    const { t } = this.props;
    return <span>{t('Hello World')}</span>;
 }
}));
```

When using the ES6 way of creating components, `translate` can be used as a decorator:

```jsx
@translate()
class Component extends React.Component {
  // ...
  render() {
    const { t } = this.props;
    return <span>{t('Hello World')}</span>;
  }
}
```
The default namespace is unchanged and is `'translation'`.

An alternative to using the `translate` decorator on the component is wrapping the rendered content with the `I18n` component, as seen [here](https://react.i18next.com/getting-started.html).

**Warning**: When trying to access member function of decorated components from the outside, you will need to 'unwrap' the component.

For that you need to first add `withRef` as an parameter to `translate`:

```jsx
var Navbar = translate('translation', {withRef: true})(React.createClass({
  hideDropdown: function()  {...}
}
```
With that is it possible to access `hideDropdown` from `app.js`:

```jsx
class App extends React.Component {
  //...

  hideNavbarDropdown() {
    var navbar = this.refs.navbar;

    if (navbar) {
      // Here, .getWrappedInstance() is used to access the component
      navbar.getWrappedInstance().hideDropdown();
    }
  }
}
```

### Interpolation, templated strings

Something like:

```js
const message = `Hello, ${name}!`;
const countMessage = `There are approximately ${Math.round(numHours)} hours left`;
```

Would become:

```js
const message = t('Hello, {{name}}!', {name});
const countMessage = t('There are approximately {{count}} hours left', {count: Math.round(numHours)});
```

### Html mixed with text

Something like:

```jsx
<div>
  Hello, {name}. You have <strong onClick={handleClick}>{messages.length}</strong> notifications.
</div>
```

Would become:

```jsx
import { Trans } from 'react-i18next';

<Trans i18nkey='html.hello-notification'>
  Hello, {{name}}. You have <strong onClick={handleClick}>{{count: messages.length}}</strong> notifications.
</Trans>
```

The key as interpreted by i18next would be `html.hello-noticiation`, and the default value would be this: `Hello, <1>{{name}}</1>. You have <3>{{count: messages.length}}</3> notifications.`. As such, the content of the html doesn't appear in the translation and is not editable by the translator. Moreover, the html can be changed without changing the key.

## Translating submodules

`@tidepool/viz` and `tideline` can add tranlations by adding `i18next` as a dependency, and using `i18next.t()`, blip is already configured to extract translatable strings from them.

They can prefix their translatable strings with `tideline|` or `viz|` to use separate namespaces from blip.

## Considerations

It is possible to implement lazy loading of translations with [i18next-xhr-backend](https://github.com/i18next/i18next-xhr-backend) with a change in webpack's config file, however locize is to be used in the future, with translations stored in the cloud.

Only a small part of the application has been translated. The idea is, if this PR is accepted and multi-language functionnality added to tidepool, to make a pull request to translate every component, maybe another to use keys as translation strings, and another to translate `viz` and `tideline`. Each of these further PRs can be subject to debate (amount of code changed, implementation, possible undetected errors when mass-updating the app...) and this is why I feel that an incremental process is better, rather than a monolithic PR with massive code changes.


app/components/chart/trends.js
```diff
@@ -35,6 +36,7 @@ const TrendsContainer = viz.containers.TrendsContainer;
 const reshapeBgClassesToBgBounds = viz.utils.reshapeBgClassesToBgBounds;
 const Loader = viz.components.Loader;
 
+@translate()
```
krystophv: Since we currently don't have any decorator patterns in the codebase, I'd be hesitant to add it right now. I'm not opposed to the idea, but perhaps I'd like @cbwebdevelopment's opinion as well.

cbwebdevelopment: I agree.  I actually do like decorator patterns (I've used them on past projects), but for consistency with the rest of the codebase at this point, I think we should hold off on introducing them, and use the HOC approach used everywhere else.

coyotte508: I've updated the PR by removing the decorator pattern. I also marked the rest of blip for translation (all tests passing) in the  `language` branch of my fork, ready to make another PR.

cbwebdevelopment:
Closing this proof-of-concept PR, which is superseded by #473
see https://github.com/tidepool-org/blip/pull/473#issuecomment-383304185

PR #470, opened 2018-03-28
Ensure we use up-to-date state timeprefs for daily print view

cbwebdevelopment:
See https://trello.com/c/dr9EeALW

The changes made for the data paging/processing feature meant that at the time the initial print view generation was called, it was not yet set to state in the PatientData instance.

I've changed it to use the incoming `nextState` as supplied from the `componentWillUpdate` hook.

This is actually how I'd intended it to work when the initial print view was released.  I simply forgot to change one instance of `this.state` to `state`.

### DIFF
--- a/app/pages/patientdata/patientdata.js
+++ b/app/pages/patientdata/patientdata.js
@@ -454,7 +454,7 @@ export let PatientData = React.createClass({
           ['basal', 'bolus', 'cbg', 'message', 'smbg']
         ),
         6,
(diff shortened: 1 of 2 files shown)


cbwebdevelopment:
[Print View  - Daily View starts at 5pm rather than midnight](https://trello.com/c/dr9EeALW/351-print-view-daily-view-starts-at-5pm-rather-than-midnight)

PR #472, opened 2018-04-04
Split email verification address to it's own line

cbwebdevelopment:
What it says.  See https://trello.com/c/67H081EV for details


cbwebdevelopment:
[Put the email address on its own line so it doesn't get split.](https://trello.com/c/67H081EV/356-put-the-email-address-on-its-own-line-so-it-doesnt-get-split)

PR #481, opened 2018-05-20
Display date range when using browser print functionality

rspier:
Tweak the CSS so that the current date range is visible in the @print media renderer.  This is a stop-gap until there are proper print views for all visualizations.


rspier:
@cbwebdevelopment

cbwebdevelopment:
[Display date range when using browser print functionality](https://trello.com/c/OKhvKFnN/422-display-date-range-when-using-browser-print-functionality)

rspier:
Thanks!

PR #482, opened 2018-05-25
Add a Dismiss option to the Tidepool Uploader launch popup

courtenayhuffman:


courtenayhuffman:
[Add a Dismiss option to the Tidepool Uploader launch popup](https://trello.com/c/XstB6XTv/398-add-a-dismiss-option-to-the-tidepool-uploader-launch-popup)

app/components/patientcard/patientcard.js
```diff
@@ -203,7 +203,7 @@ var PatientCard = React.createClass({
       <div>
         <div className="ModalOverlay-content">{"Are you sure you want to leave this person's Care Team? You will no longer be able to view their data."}</div>
         <div className="ModalOverlay-controls">
-          <button className="PatientInfo-button PatientInfo-button--secondary" type="button" onClick={this.overlayClickHandler}>Cancel</button>
+          <button className="PatientInfo-button PatientInfo-button--secondary" type="button" onClick={this.modalDismissHandler}>Cancel</button>
```
cbwebdevelopment: This breaks, because the `overlayClickHandler` method below hasn't been renamed to `modalDismissHandler` (it should be).

app/components/patientcard/patientcard.js
```diff
@@ -220,7 +220,7 @@ var PatientCard = React.createClass({
   },
 
   renderUploadOverlay: function() {
-    return <UploadLaunchOverlay overlayClickHandler={()=>{this.setState({showUploadOverlay: false})}}/>
+    return <UploadLaunchOverlay modalDismissHandler={()=>{this.setState({showUploadOverlay: false})}}/>
```
cbwebdevelopment: After renaming the `overlayClickHandler` method  to `modalDismissHandler`, I'd suggest cleaning this duplicated up to use the instance method instead of re-creating the handler:
```
return <UploadLaunchOverlay modalDismissHandler={this.modalDismissHandler}/>
```

app/components/uploadlaunchoverlay/uploadlaunchoverlay.less
```diff
@@ -39,10 +39,18 @@
...
   font-size: @font-size-small;
   font-weight: lighter;
 }
+
+.uploadDismiss {
+  font-size: 30px;
+  font-weight: 500;
+  float: right;
+  color: #6d6d6d;
+  padding-right: 5px;
```
cbwebdevelopment: The hover style is somewhat incomplete:
image

Let's remove the underline on hover, and set the cursor to use a pointer;  I've also, in the example below, used absolute positioning instead of floats/padding, which is more predictable.

I think we should rename the class and move this to `modaloverlay.less`, as I could see this being useful for other modals at some point:
```
.ModalOverlay-dismiss {
  font-size: 30px;
  font-weight: 500;
  color: #6d6d6d;
  cursor: pointer;
  position: absolute;
  top: 16px;
  right: 20px;

  &:hover {
    text-decoration: none;
  }
}
```

You will also need to change the `uploadDismiss` classname in the corresponding components and tests (`uploadlaunchoverlay.js` and `uploadlaunchoverlay.test.js`).

courtenayhuffman:
Ready for PR review again :)

PR #487, opened 2018-06-06
Fix login and share page errors when a shared account is missing a profile

cbwebdevelopment:
See https://trello.com/c/4lnTETQg for details


krystophv:
[Cannot log in to Jill Jellyfish account](https://trello.com/c/4lnTETQg/413-cannot-log-in-to-jill-jellyfish-account)

PR #490, opened 2018-06-14
Fix issue where weekly chart date was set to the next day when switching from daily

cbwebdevelopment:
See https://trello.com/c/Jpi5YurW for details.

Related PR:
tidepool-org/tideline#345


cbwebdevelopment:
[% Automode Widget Weekly view is inaccurate when there is less than 14 days of data in the viewed range.](https://trello.com/c/Jpi5YurW/416-automode-widget-weekly-view-is-inaccurate-when-there-is-less-than-14-days-of-data-in-the-viewed-range)

PR #507, opened 2018-09-11
Add CSP worker-src fallback for older browsers

cbwebdevelopment:
See https://trello.com/c/GbuHSNpr for details.


cbwebdevelopment:
[CSP Error in Safari and MS Edge](https://trello.com/c/GbuHSNpr/470-csp-error-in-safari-and-ms-edge)

PR #519, opened 2019-01-02
Allow medtronic query parameter when fetching device data

darinkrauss:


darinkrauss:
[When insulin healthkit uploads begin, hide Medtronic Direct uploads after insulin healthkit introduction began (September 2017)](https://trello.com/c/a4KNttXb/514-when-insulin-healthkit-uploads-begin-hide-medtronic-direct-uploads-after-insulin-healthkit-introduction-began-september-2017)

PR #541, opened 2019-02-25
Sync 1.17 release from `develop`

krystophv:
Will need to happen again when 1.16 hits production, but this will get the carb rendering work into the 1.17 release branch


krystophv:
Going to merge this as it's only commit is a merge that was reviewed already.

PR #542, opened 2019-02-28
Persist Average Daily Dose input between data views

cbwebdevelopment:
See https://trello.com/c/8w4zfRPq for details.

Goes with tidepool-org/viz#133

### DIFF
--- a/app/components/chart/stats.js
+++ b/app/components/chart/stats.js
@@ -27,6 +27,7 @@ class Stats extends Component {
     chartType: PropTypes.oneOf(['basics', 'daily', 'weekly', 'trends']).isRequired,
     dataUtil: PropTypes.object.isRequired,
     endpoints: PropTypes.arrayOf(PropTypes.string),
+    onAverageDailyDoseInputChange: PropTypes.func,
   };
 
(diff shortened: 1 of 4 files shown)


cbwebdevelopment:
[Regression: User-entered data for the new Avg. Daily Insulin widget does not persist](https://trello.com/c/8w4zfRPq/578-regression-user-entered-data-for-the-new-avg-daily-insulin-widget-does-not-persist)

app/components/chart/stats.js
```diff
@@ -131,15 +132,24 @@ class Stats extends Component {
...
         bgPrefs: {
           bgBounds,
           bgUnits,
         },
         manufacturer,
-      }));
+        ...chartStatOpts,
+      });
+
+      if (statType === 'averageDailyDose' && typeof props.onAverageDailyDoseInputChange === 'function') {
```
krystophv: I feel a slight preference to use `_.isFunction` here, but it's mostly just a preference.

cbwebdevelopment: I agree.  I'll change this.

PR #550, opened 2019-03-25
Added health check endpoint.

derrickburns:
The intent of this is to create a [readiness probe for use in Kubernetes](https://supergiant.io/blog/creating-liveness-probes-for-your-node-js-application-in-kubernetes/).

This adds a `/healthz` endpoint that simply returns a 200 status when called. The idea is that while yarn is build the app on startup, calls to this endpoint will FAIL. Only when the server is started and ready to serve traffic will this Kubernetes send traffic to it, as indicated when this endpoint returns status 200.

### DIFF
--- a/server.js
+++ b/server.js
@@ -87,6 +87,10 @@ app.get('*', (req, res) => {
   res.send(res.locals.htmlWithNonces);
 });
 
+app.get('/healthz', (req, res) => {
+  res.status(200).end();
+});
+
 app.post('/event/csp-report/violation', (req, res) => {
   if (req.body) {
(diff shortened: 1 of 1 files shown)


derrickburns:
Agreed.

Perhaps this is not an "either/or" choice, but a "both" choice.  

Can we start with this?

pazaan:
@cbwebdevelopment - I discussed with @derrickburns that since `blip` is ideally statically served in production, that we should close this PR and instead figure out how to make a "production" image of `blip`.

cbwebdevelopment:
@pazaan - Thanks for the heads-up.

PR #552, opened 2019-03-26
Fix Y-Axis bg tick overlaps on Trends view

cbwebdevelopment:
See https://trello.com/c/EnbpV4VY for details.

See also tidepool-org/viz#139

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "@babel/polyfill": "7.0.0",
     "@babel/preset-env": "7.1.0",
     "@babel/preset-react": "7.0.0",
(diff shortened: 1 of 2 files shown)


cbwebdevelopment:
[On INT (& PROD) with default target BG range, the Trends y-axis displays 70 and 54 on top of one another for T-slim, Animas, Omnipod and VerioFlex](https://trello.com/c/EnbpV4VY/610-on-int-prod-with-default-target-bg-range-the-trends-y-axis-displays-70-and-54-on-top-of-one-another-for-t-slim-animas-omnipod-an)

PR #554, opened 2019-03-28
Ensure select input values have a matching option value

cbwebdevelopment:
Fixes https://trello.com/c/1JmHhwKi

The bug presented itself upon the removal of BDD partners from our list, but the underlying issue was in trying to set values in our Select implementation that didn't have corresponding select options.


cbwebdevelopment:
[Profile Page crashing post removal of BDD target](https://trello.com/c/1JmHhwKi/613-profile-page-crashing-post-removal-of-bdd-target)

cbwebdevelopment:
Yes, I noticed that as well :)

PR #555, opened 2019-03-29
Release 1.18.0 to master

cbwebdevelopment:


cbwebdevelopment:
For this merge, I ended up accidentally merging develop into the branch before merging into master, so I had to revert the merge prior to merging to master.

The end result is that the merge to master is identical to the state at the v1.18.0 tag, as can be verified by comparing here: https://github.com/tidepool-org/blip/compare/4bbbad7..b5d80be

PR #556, opened 2019-03-29
Release 1.18.0 to develop

cbwebdevelopment:


cbwebdevelopment:
@krystophv As mentioned in the corresponding PR to master (https://github.com/tidepool-org/blip/pull/555), I ended up accidentally merging develop into the branch before merging into master, so I had to revert the merge prior to merging to master.

The end result is that the merge to master was identical to the state at the v1.18.0 tag, as can be verified by comparing here: https://github.com/tidepool-org/blip/compare/4bbbad7..b5d80be

Now, to safely merge back to `develop`, I needed to "revert the revert" so that the changes on develop wouldn't be undone. 😞 

The state of this branch is now identical to where it was at commit `7c80a97` (which is where the head of this PR would have been had I not forgotten the merge to master), as can be seen here: https://github.com/tidepool-org/blip/compare/7c80a97..040c4f8

Sorry for the confusion.  Let me know if you want to go over any of this in more detail.

krystophv:
LGTM 🚀

PR #559, opened 2019-04-12
Print view stat calculations

cbwebdevelopment:
See https://trello.com/c/zCsP7dCr for details.

Goes with tidepool-org/viz#143


cbwebdevelopment:
[Use new Agg Stats calculations in Tidepool Web's Print Views](https://trello.com/c/zCsP7dCr/597-use-new-agg-stats-calculations-in-tidepool-webs-print-views)

PR #564, opened 2019-04-24
strip extra space in share email fields

courtenayhuffman:
https://trello.com/c/r3GiIwuA/667-strip-out-extra-spaces-in-share-email-fields


courtenayhuffman:
[Strip out extra spaces in Share email fields](https://trello.com/c/r3GiIwuA/667-strip-out-extra-spaces-in-share-email-fields)

app/pages/patient/patientteam.js
```diff
@@ -131,7 +131,7 @@ var MemberInviteForm = translate()(React.createClass({
 
     var validateEmail = function(email) {
       var re = /^(([^<>()[\]\\.,;:\s@\"]+(\.[^<>()[\]\\.,;:\s@\"]+)*)|(\".+\"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/;
-      return re.test(email);
+      return re.test(email.trim());
```
cbwebdevelopment: The problem with only trimming the email here in the validator is that it will still submit `email` in its untrimmed format, which we wouldn't want.  

I'd suggest undoing the `trim()` call here and instead trimming `email` when you're setting the variable above, and perhaps even falling back to an empty string via `lodash.get()` just in case the email value was something other than a string (like `undefined`, which would throw an error since it doesn't have a `trim` method).

```javascript
var email = _.get(self, 'refs.email.value', '').trim();
```

PR #572, opened 2019-05-14
Build at build time

derrickburns:
This change builds the server when the Docker image is build instead of when it is run. 

The other way does not play well with Kubernetes. :)

See https://trello.com/c/hYwNJsw7 for details

Related PRs
---
tidepool-org/development#21
tidepool-org/viz#151
tidepool-org/tideline#374


derrickburns:
I am happy with a multi-stage build, but I am probably not the best person to write the Dockerfile since my understanding of yarn/npm is minimal.  @cbwebdevelopment can you take a crack at that? 

My only real requirement is that I have a production Docker image that does not attempt to build anything at run time.

config.app.js
```diff
@@ -48,7 +48,7 @@ function integerFromText(value, defaultValue) {
 const config = {
   VERSION: pkg.version,
   UPLOAD_API: __UPLOAD_API__ || 'https://tidepool.org/uploader',
-  API_HOST: __API_HOST__ || 'https://dev-api.tidepool.org',
+  API_HOST: __API_HOST__ || `${window.location.protocol}//${window.location.host}`.replace('app', 'api'),
```
derrickburns: I would suggest changing the replacement from
.replace("app", "api") to
.replace("app.tidepool.org", "api.tidepool.org")
The reason is to avoid possible false positive matches for test environments where the name might be:
"\*app\*-app.tidepool.org".

cbwebdevelopment:
[Update frontend Docker builds to work with Kubernetes flow](https://trello.com/c/hYwNJsw7/705-update-frontend-docker-builds-to-work-with-kubernetes-flow)

PR #577, opened 2019-05-27
bump tideline

courtenayhuffman:
https://trello.com/c/3u361BoE/702-web-fails-to-load-data-views-due-to-negative-bg-value

https://github.com/tidepool-org/tideline/pull/375


courtenayhuffman:
[Web fails to load data views due to negative BG value](https://trello.com/c/3u361BoE/702-web-fails-to-load-data-views-due-to-negative-bg-value)

PR #582, opened 2019-06-08
Allow connections from integration-test.

derrickburns:
This adds a new subdomain to the CSP config to allow integration testing in this subdomain.

### DIFF
--- a/server.js
+++ b/server.js
@@ -28,6 +28,7 @@ const nonceMiddleware = (req, res, next) => {
 }
 
 app.use(helmet());
+
 app.use(nonceMiddleware, helmet.contentSecurityPolicy({
   directives: {
     defaultSrc: ["'none'"],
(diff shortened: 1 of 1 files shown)


server.js
```diff
@@ -69,6 +70,9 @@ app.use(nonceMiddleware, helmet.contentSecurityPolicy({
       'https://chatapi.helpscout.net',
       'https://d3hb14vkzrxvla.cloudfront.net',
       'wss\://*.pusher.com',
+      '*.tidepool.org',
+      '*.integration-test.tidepool.org',
+      'http://*.integration-test.tidepool.org',
```
cbwebdevelopment: Just wondering, wouldn't `*.tidepool.org` cover the other 2 entries?

derrickburns:
No it doesn't. The "*" is not a regex, but a pattern that covers names without periods in them. Also the absence of a protocol means use https.

Derrick

cbwebdevelopment:
Ah.  Sounds good then 👍

PR #583, opened 2019-06-13
Handle negative bg values

courtenayhuffman:
https://trello.com/c/3u361BoE/702-web-fails-to-load-data-views-due-to-nonpositive-integer#

https://github.com/tidepool-org/tideline/pull/378


courtenayhuffman:
[Web fails to load data views due to nonpositive integer](https://trello.com/c/3u361BoE/702-web-fails-to-load-data-views-due-to-nonpositive-integer)

PR #586, opened 2019-06-21
Reduce idle cpu usage on data views

cbwebdevelopment:
https://trello.com/c/kiGrj7Oo

The main fix was to ensure that the CSS loading dot animations are only happening while visible.

The other fix is a workaround for the time being, allowing the the stat animations to be disabled via a `animateStats=false` query param.

Related PR: tidepool-org/viz#156

### DIFF
--- a/app/components/chart/stats.js
+++ b/app/components/chart/stats.js
@@ -91,16 +91,18 @@ class Stats extends Component {
       : false;
   };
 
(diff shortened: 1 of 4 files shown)


cbwebdevelopment:
[High CPU usage in Tidepool Web](https://trello.com/c/kiGrj7Oo/729-high-cpu-usage-in-tidepool-web)

PR #587, opened 2019-06-21
Fix bgLog stat regression in print view

cbwebdevelopment:
When the weekly view was renamed to BG Log, the name change was missed for the summary stats generation for the print view.  This fixes that :) 

See https://trello.com/c/i2k5qiZX

### DIFF
--- a/app/components/chart/stats.js
+++ b/app/components/chart/stats.js
@@ -91,16 +91,18 @@ class Stats extends Component {
       : false;
   };
 
-  renderStats = stats => (_.map(stats, stat => (
+  renderStats = (stats, animate) => (_.map(stats, stat => (
     <div id={`Stat--${stat.id}`} key={stat.id}>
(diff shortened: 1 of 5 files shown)


cbwebdevelopment:
[Avg. BG value missing from Print View PDF](https://trello.com/c/i2k5qiZX/730-avg-bg-value-missing-from-print-view-pdf)

cbwebdevelopment:
@krystophv Just a heads up.  I did it again :|  Accidentally merged in develop before merging to master.  As I had to do the last time this happened, I've reverted the merge, and confirmed that there are zero file changes between the original commit I wanted to merge to master and the current HEAD: https://github.com/tidepool-org/blip/compare/72bc8a9...6694c30

I'll merge this to `master`, then "revert the revert" (so that it's once again a clean merge back to `develop`)

PR #589, opened 2019-06-24
Updated text fix v1

jpreillymb:
Updated text bio bug where a user would put in a long string with no spaces, now properly wraps 
https://trello.com/c/tvrji3bT/576-long-profile-text-bug


jpreillymb:
[Long Profile Text Bug](https://trello.com/c/tvrji3bT/576-long-profile-text-bug)

PR #590, opened 2019-06-24
Added dynamic text count

jpreillymb:
Added a text count for editing a users bio (ie, 26/256 or 276/256). If a user is over the allotted count for a bio, then the counter turns error red.

Part 3 of this Trello 
https://trello.com/c/DI2bGRAa


jpreillymb:
[Tidepool Web – Profile Polish](https://trello.com/c/DI2bGRAa/734-tidepool-web-profile-polish)

PR #624, opened 2019-10-03
Basics and Trends copy to clipboard [WEB-2]

cbwebdevelopment:
Adds copy to clipboard buttons to the basics and trends views

See [WEB-2] for details


thacherT1D:
LGTM 👍

PR #626, opened 2019-10-13
Fix csp violations. Addresses WEB-597

derrickburns:
There are two new subdomains that need to be added.


derrickburns:
@cbwebdevelopment A better idea would be to generate the list of hosts dynamically by querying the DNS SRV record for Tidepool environments....

cbwebdevelopment:
@derrickburns We could certainly do that, but we'd need to have a Jira issue created and prioritized, as it'd be a bit more work than with the uploader, since we'd need to cache the results for a to-be-determined period of time to make sure that we don't fetch the SRV records on each request.

Do you still want to go ahead with this PR in the meantime?

derrickburns:
Yes. Let's continue with the pr

Derrick

PR #628, opened 2019-10-18
Merge 1.29.0-csp into release branch

cbwebdevelopment:


server.js
```diff
@@ -65,14 +65,15 @@ app.use(nonceMiddleware, helmet.contentSecurityPolicy({
     frameSrc: ['https://docs.google.com'],
     connectSrc: [].concat([
       process.env.API_HOST || 'localhost',
-      'https://api.github.com/repos/tidepool-org/uploader/releases',
+      'https://api.github.com/repos/tidepool-org/chrome-uploader/releases',
```
krystophv: This change will need to be dropped.

cbwebdevelopment: Ok. Done

PR #644, opened 2020-01-20
[WEB-4] Add queue to manage concurrent data worker queries

cbwebdevelopment:
While preparing the a release candidate for QA of [WEB-4], I realized that the data view changes weren't changing while the PDF was generating.

I needed to create a queue to manage concurrent data worker queries to make sure each one was completed before another was started, and to make sure that all requested queries were completed.  

As part of this, I'm also sending the patient id across as part of the action metadata for any requests added to the queue so that I can flush the queue if we change patients.

I also fixed a number of small scroll/datetime location issues on daily and bg log views that surfaced while I was checking things over, as well as a fix for the 'refresh' button on the data views, and the date range on the Trends 'Copy to Text' button output, which had gotten broken at some point.

[WEB-4]: https://tidepool.atlassian.net/browse/WEB-4


cbwebdevelopment:
@krystophv I hear you about writing my own queue.  I thought about adding a library, but my needs were about as simple as could be, and it seemed overkill to pull another library in for what's basically a couple of array methods dropped into a class.

I'm divided on this as well.  I could implement queue from the `async` library, which we already have:
https://caolan.github.io/async/v2/docs.html#queue

Should be fairly simple.  I'll give it a quick spin.

cbwebdevelopment:
@krystophv I've implemented `async.queue` and removed mine.  Can you please review that last commit?

PR #658, opened 2020-02-20
[WEB-741] Fix stats missing when using cached patient data

cbwebdevelopment:
Addressed [WEB-741]

Fixes issue found during QA where some stats were missing due the `bgSources` metaData not being preserved when using cached patient data (such as when changing to the profile view and then back to the data views)

Solution was to cache the full metaData object in the redux store when using cache, instead of just the patient id

[WEB-741]: https://tidepool.atlassian.net/browse/WEB-741


cbwebdevelopment:
Oops.  Looks like I forgot to update a unit test.  I'll do this now.

PR #661, opened 2020-02-22
[WEB-742] style linting setup

thacherT1D:


.stylelintrc
```diff
@@ -0,0 +1,7 @@
+{
+  "processors": ["stylelint-processor-styled-components"],
+  "extends": [
+    "stylelint-config-recommended",
+    "stylelint-config-styled-components"
+  ]
+}
```
cbwebdevelopment: Just a note: we typically have adopted the practice of having our editors automatically inserting a newline at the end of each file (in viz, we actually enforce the `eol-last` rule, and I assume we will be pulling that over to this repo as well)

app/themes/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
```
cbwebdevelopment: Ideally, for the new folders, we should use the `.eslintrc` from `viz`.

That one extends the airbnb eslint base, and denotes that we use `single` for our quotes rule.  If there are sections within the theme that require a different rule (because it wouldn't make sense to use single quotes), then we can disable or modify the rule via an `eslint-disable` comment within the code.

package.json
```diff
@@ -20,7 +20,8 @@
     "serve-docs": "./node_modules/.bin/gitbook serve",
     "update-translations": "i18next 'app/**/*.js' 'node_modules/tideline/{plugins,js}/**/*.js' 'node_modules/@tidepool/viz/{storiesDatatypes,storybookDatatypes,stories,src}/**/*.js' -c i18next-parser.config.json -o .",
     "storybook": "./node_modules/.bin/start-storybook -p 6006 --ci",
-    "build-storybook": "build-storybook"
+    "build-storybook": "build-storybook",
+    "lint:css":"stylelint './app/components/**/*.js'"
```
cbwebdevelopment: I'd recommend that we also add the `lint:css` script to the `pretest` script so that a failed css linting will automatically fail the `test` script.

package.json
```diff
@@ -133,6 +134,10 @@
     "source-map-loader": "0.2.4",
     "storybook-addon-designs": "5.2.0",
     "style-loader": "0.23.0",
+    "stylelint": "^13.2.0",
```
cbwebdevelopment: Need to lock down versions here

.stylelintrc
thacherT1D: I was looking for this in the viz .eslintrc for this, but found it was being enforced in an .editorconfig file, so I'm adding that file to blip -- there's nothing in there that I think would negatively affect the rest of blip, but let me know if I'm wrong there

app/themes/.eslintrc
thacherT1D: I've moved over a copy of the viz `.eslintrc` to the `elements` folder for the new components, but I did not see (or find with search) a quotes rule in viz, there is one already in the blip main `.eslintrc` -- seems to be working well when running it, but being mindful of the proliferation of `.eslintrc` files I removed the one specific to the `baseTheme` and just disabled the quote rule with a `eslint-disable` comment at the top of the file

app/themes/baseTheme.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable quotes */
```
cbwebdevelopment: I don't see anything in this file that requires disabling the quotes rule.
The strings surrounded by backticks should be converted to single quotes, and then this file should pass linting.

app/themes/.eslintrc
cbwebdevelopment: The quotes rule comes from the airbnb eslint config that the one from viz extends.  Basically, we use that one as a starting point and then override it as required for our preferences.

.stylelintrc
cbwebdevelopment: That should be fine 👍

app/themes/baseTheme.js
thacherT1D: Got it -- I had thought that the backpacks were required for styled-components, but looking at the rest of this file I agree that converting to single quotes and removing the `eslint-disable` is the way to go.

cbwebdevelopment:
Approved 👍

PR #664, opened 2020-02-28
[WEB-591] Create single and multi day date selector components

cbwebdevelopment:
Addresses [WEB-591]

[Figma designs](https://www.figma.com/file/iuXkrpuLTXExSnuPJE3Jtn/Tidepool-Design-System---Sprint-1?node-id=198%3A6)

Note: the multi date picker design differs from the Figma mocks, but Kelly has approved the design I implemented.

As well, this PR includes a fair bit of revamping to the theming structure in order to bring it up to date with the [System UI spec](https://styled-system.com/theme-specification/), and to incorporated the official [Tidepool color palettes](https://tidepool-design.netlify.com/#/Utilities/Color)

One other note: I did not create any unit tests for these for 2 reasons:
1. I'd like to get this merged to the main feature branch ASAP to allow the devs working on other components to have access to the theme structure updates
2. I feel like the frontend team needs to huddle together and decide how we want to go about testing the new components.  Do we keep doing what we've been doing? Try Jest storyshots? Something else?

[WEB-591]: https://tidepool.atlassian.net/browse/WEB-591


.eslintrc
```diff
@@ -23,13 +23,13 @@
     "no-mixed-requires": [0],
     "no-underscore-dangle": [0],
     "no-bitwise": 2,
-    "camelcase": 1,
+    "camelcase": ["error", { "allow": ["^UNSAFE_"] }],
```
krystophv: Yeah, these are the better way to go about this, good call.

app/components/elements/DatePicker.js
```diff
@@ -0,0 +1,92 @@
...
+
+import { DatePicker as StyledDatePickerBase } from './InputStyles';
+
+import {
+  colors,
+  fontSizes,
+  radii,
+  shadows,
+  space,
+} from '../../themes/baseTheme';
```
krystophv: Is this more for just dropping the repeated object nest reference (`theme.colors`, `theme.fontSizes` etc) from the default export or for hopeful tree-shaking down the line? I'm fine with it either way, just curious.

app/components/elements/DatePicker.js
```diff
@@ -0,0 +1,92 @@
+import React, { useState } from 'react';
+import momentPropTypes from 'react-moment-proptypes';
+import { SingleDatePicker, SingleDatePickerShape } from 'react-dates';
+import NavigateBeforeRoundedIcon from '@material-ui/icons/NavigateBeforeRounded';
+import NavigateNextRoundedIcon from '@material-ui/icons/NavigateNextRounded';
+import CloseRoundedIcon from '@material-ui/icons/CloseRounded';
+import omit from 'lodash/omit';
+import noop from 'lodash/noop';
```
krystophv: I like the specific function imports on `lodash`

app/themes/colorPalette.js
```diff
@@ -0,0 +1,148 @@
...
+  },
+  neutrals: {
+    white: '#FFFFFF',
+    lightestGrey: '#F9F9F9',
+    lightGrey: '#EDEDED',
+    mediumGrey: '#979797',
+    darkGrey: '#606060',
+    black: '#000000',
+  },
+  extended: {
```
krystophv: I know you're just pulling these from the design sheet, but the fact that there aren't an equal number of options in each of these arrays seems like it could get annoying. Thinking along the lines of "I need a darker orange but I can't remember how big the oranges array is to know how far I should go." Not a comment on the code for this review, but a comment about a conversation that might be had with design.

.eslintrc
cbwebdevelopment: 👍

app/components/elements/DatePicker.js
cbwebdevelopment: The first -- just wanted to avoid the nesting.

app/components/elements/DatePicker.js
cbwebdevelopment: Me too :)

app/themes/colorPalette.js
cbwebdevelopment: Yeah, may be worth discussing at our next frontend/design huddle.

PR #665, opened 2020-03-03
[WEB-559] Set up and use standard line heights in theme

cbwebdevelopment:
Noticed that all the line-heights in the `FontStyles` components were hard-coded to pixel values.  I've updated them to use unit-less values from the theme.

I've requested a couple of reviewers, but I'll just merge this in once one of you approves.

### DIFF
--- a/app/themes/baseTheme.js
+++ b/app/themes/baseTheme.js
@@ -78,6 +78,7 @@ export default {
   fonts,
   fontSizes,
   fontWeights,
+  lineHeights,
   mediaQueries,
   radii,
   shadows,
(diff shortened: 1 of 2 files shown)


app/themes/baseTheme.js
```diff
@@ -45,6 +45,8 @@ export const fontWeights = {
   black: 900,
 };
 
+export const lineHeights = [1, 1.25, 1.33, 1.5, 1.75, 2];
```
krystophv: Should this also be an exported property on the default export?

cbwebdevelopment: Oh! :) good call

cbwebdevelopment: Done.  Merging now

PR #667, opened 2020-03-05
[WEB-589] Dialog component

cbwebdevelopment:
See [WEB-589] for issue details
See [Figma Design](https://www.figma.com/file/iuXkrpuLTXExSnuPJE3Jtn/Tidepool-Design-System---Sprint-1?node-id=8%3A826)

[WEB-589]: https://tidepool.atlassian.net/browse/WEB-589

This also wraps in the Button component, which is a sub-task of this one (See [WEB-583]).  I've replaced the sample button component from the website theme with a much lighter themed rebass component, with primary and secondary variations.  One note: there are a few button states missing on Figma.  Kelly is working on these.  If what I've done here stylistically needs to change, I'll address that in a separate PR.

This is basically a minimally themed wrapper around the [MUI Dialog](https://material-ui.com/components/dialogs/), and is intended to be used in the same way.  An example implementation storybook with some customizable knobs has been added.

As well, I've added the storybook 'viewports' addon, which allows us to test our storied components in a number of common device sizes in portrait or landscape to ensure that we are building responsively from the start.

[WEB-583]: https://tidepool.atlassian.net/browse/WEB-583


app/components/elements/Dialog.js
```diff
@@ -0,0 +1,100 @@
...
+  ...DialogActionsProps,
+};
+
+DialogActions.defaultProps = {
+  disableSpacing: true,
+};
+/* Dialog Actions End */
+
+/* Dialog Start */
+const StyledDialog = styled(MuiDialog)``;
```
krystophv: It's not immediately obvious to me why we'd do this. What does running the `MuiDialog` through `styled()` with no styling gain us?

cbwebdevelopment: Good catch.  I had it scaffolded figuring there would be something for me to override, and had planned to remove it at the end if it was still empty.

Will do now.

PR #672, opened 2020-03-09
[WEB-582] text input

krystophv:
See [WEB-582] for details

[WEB-582]: https://tidepool.atlassian.net/browse/WEB-582

### DIFF
--- a/app/components/elements/TextInput.js
+++ b/app/components/elements/TextInput.js
@@ -1,0 +1,68 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { Box } from 'rebass/styled-components';
+import { Label, Input } from '@rebass/forms';
+import { Caption } from './FontStyles';
+import { fonts, fontSizes, colors, borders, radii, space } from '../../themes/baseTheme';
+
+export const TextInput = props => {
+  const StyledInput = styled(Input)`
+    border: ${borders.input};
+    box-shadow: none;
+    border-radius: ${radii.input}px;
+    padding: ${styleprops => (styleprops.space ? space[styleprops.space] : 12)}px;
+    caret-color: ${colors.mediumPurple};
+    font-size: ${fontSizes[1]}px;
+    font-family: ${fonts.default};
+    color: ${colors.mediumPurple};
+
+    &::placeholder {
+      color: ${colors.text.primaryTextSubdued};
+    }
+
+    &.active {
+      color: ${colors.text.primaryTextSubdued};
+      box-shadow: none;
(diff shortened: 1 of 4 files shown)


app/components/elements/TextInput.js
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { Box } from 'rebass/styled-components';
+import { Label, Input } from '@rebass/forms';
+import { Caption } from './FontStyles';
+import { fonts, fontSizes, colors, borders, radii, space } from '../../themes/baseTheme';
+
+export const TextInput = props => {
+  const StyledInput = styled(Input)`
```
cbwebdevelopment: It's important, for performance reasons, to define the styled component outside of the render method or functional component that will be using it.

See https://styled-components.com/docs/basics#define-styled-components-outside-of-the-render-method for reference

app/components/elements/TextInput.js
```diff
@@ -0,0 +1,68 @@
...
+      sx={{
+        width: props.width,
+      }}
+    >
+      {props.label &&
+        <Label htmlFor={props.name}>
+          <Caption>{props.label}</Caption>
+        </Label>
+      }
+      <StyledInput id={props.name} {...props} />
```
cbwebdevelopment: This isn't a big deal, but I'm not sure if you intended to add the `label`, `name`, and `width` props to the `StyledInput` as well.  I don't think they could really cause issues (though I'm wondering if the width being added to the padding applied to the input could be...)

I'd consider pulling the other props out and getting the input props via rest destructuring (unless you intended to pass all the props through)

```
const { label, name, width, ...inputProps } = props;
...
<StyledInput id={props.name} {...inputProps} />
```

app/components/elements/TextInput.js
krystophv: Should we also change it in `IconButton` then? (I used this as a reference to build this component out - https://github.com/tidepool-org/blip/blob/web-559-component-library/app/components/elements/IconButton.js#L8)

app/components/elements/TextInput.js
krystophv: Not entirely intentional, but I didn't see it hurting so I shrugged it off and didn't come back to address it after time of writing. Destructured is much nicer/cleaner and more specific/intentional. 👍

app/components/elements/TextInput.js
cbwebdevelopment: Ah yes!  I'll change it there.  Thanks for pointing that out!

PR #673, opened 2020-03-10
[WEB-590] Popover Component

cbwebdevelopment:
See [WEB-590]

[WEB-590]: https://tidepool.atlassian.net/browse/WEB-590

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "express": "4.16.3",
     "hakken": "0.1.11",
     "helmet": "3.13.0",
-    "material-ui-popup-state": "^1.5.3",
+    "material-ui-popup-state": "1.5.3",
     "react-dates": "21.8.0",
     "react-moment-proptypes": "1.7.0",
     "react-router-dom": "5.1.2",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -36,6 +36,7 @@
     "express": "4.16.3",
     "hakken": "0.1.11",
     "helmet": "3.13.0",
+    "material-ui-popup-state": "^1.5.3",
```
krystophv: Just need to lock this down

cbwebdevelopment: Oops :) Will do.

thacherT1D:
LGTM 🦄 -- the title v. body text colors are different in the figma design, but seem to be the same in the built component. Also, the "i" icon is also a bit different, but I think that's using the material-ui icon that was agreed upon? yes?

cbwebdevelopment:
@thacherT1D Yes, I went with the material icon intentionally here.  For the color difference, I missed that.  The actual component doesn't manage the color of the content placed in it, but I have been trying to keep the storybook examples consistent with the Figma designs.

I've fixed the color.  Will push and merge.

PR #674, opened 2020-03-10
[WEB-593] Checkbox Component

thacherT1D:


app/components/elements/Checkbox.js
```diff
@@ -0,0 +1,77 @@
...
+  background: ${props => (props.checked ? colors.purpleMedium : colors.lightestGrey)};
+  border: 1px solid ${props => (props.checked ? colors.purpleMedium : colors.lightGrey)};
+  border-radius: 2px;
+
+  ${Icon} {
+    visibility: ${props => props.checked ? 'visible' : 'hidden'};
+  }
+`;
+
+export const InputLabel = styled.span`
```
cbwebdevelopment: Again, I think we'd be better to use the `Label` component from `@rebass/forms` instead of creating our own.  This would also keep things consistent with the approach Chris took with the `TextInput` component

app/components/elements/Checkbox.js
```diff
@@ -0,0 +1,77 @@
...
+  padding: 0 0 0 8px;
+  color: ${props => (props.color ? props.color : colors.text.primary)};
+`;
+
+const Checkbox = ({ className, checked, ...props }) => (
+  <div>
+    <CheckboxContainer className={className}>
+      <HiddenCheckbox checked={checked} {...props} />
+      <StyledCheckbox checked={checked}>
+        <Icon viewBox="0 0 24 24">
```
cbwebdevelopment: For icons, we should stick with the material icons.  It will be better for consistency, and is much easier to reason about than looking at svg elements.

stories/Checkbox.stories.js
```diff
@@ -0,0 +1,106 @@
...
+
+import { withDesign } from 'storybook-addon-designs';
+import { withKnobs, text } from '@storybook/addon-knobs';
+import { ThemeProvider } from 'styled-components';
+
+import baseTheme from '../app/themes/baseTheme';
+import Checkbox from '../app/components/elements/Checkbox';
+import { CheckboxGroupTitle } from '../app/components/elements/FontStyles';
+
+/* eslint-disable jsx-a11y/label-has-associated-control */
```
cbwebdevelopment: I'm hesitant to disable accessibility checks without a solid reason.  I think we really should be doing what we can to enforce a11y practices in our new components, and see our storybook serving as examples of best practices of how to use our components so in instances where we can't enforce it via prop-types, we have examples to copy from that have a11y practices built-in.

cbwebdevelopment:
@thacherT1D Just for a quick example, I've pushed a `web-593-rebass-checkbox-alt` branch to show a possible implementation using the rebass checkbox that I think might be simpler in the long run.  It's not styled to the Figma spec yet, and has an arbitrary default 'width' prop that I haven't given much thought to, but I figured I'd pass it along for you to consider.

cbwebdevelopment:
Here's a link for a quick look at the possible rebass implementation:
https://github.com/tidepool-org/blip/compare/web-593-checkbox..web-593-rebass-checkbox-alt

stories/0-Examples.stories.js
```diff
@@ -1,47 +1,21 @@
 import React from 'react';
```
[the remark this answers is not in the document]
thacherT1D: I thought I had -- I think I messed up something in the merge because another file came back too (that was messing up the Travis tests when I first pushed this) -- will remove it

app/components/elements/Checkbox.js
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+import styled from 'styled-components';
+
```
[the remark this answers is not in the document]
thacherT1D: In general I agree, but I had significant trouble getting down to be able to manipulate the rebass component -- I'll take a look at the example that you pushed and see if I missed something when looking at it before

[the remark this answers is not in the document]
cbwebdevelopment: I suppose it depends on what type of manipulation you need.  I'm imagining just changing the sizing/color/label fonts.  If there's more to it, it could get tricky, but then, I'd push back on any design requirements that force us to get tricky.

app/components/elements/Checkbox.js
```diff
@@ -0,0 +1,75 @@
...
+  colors,
+  fonts,
+  fontSizes,
+  fontWeights } from '../../themes/baseTheme';
+
+const StyledCheckbox = styled(Base)`
+  color: ${colors.border.default};
+  height: 1.5em;
+  width: 1.5em;
+  margin-right: .25em;
```
cbwebdevelopment, line 18: I noticed that the checkbox focus background renders really large due to the default padding on the rebass checkbox.

It can reduced, while maintaining the same general sizing/spacing, by updating the styles as:
```css
  height: 1em;
  width: 1em;
  padding: 0;
  margin-right: 0.5em;
```

app/components/elements/Checkbox.js
```diff
@@ -0,0 +1,75 @@
...
+  );
+};
+
+Checkbox.propTypes = {
+  ...CheckboxProps,
+  label: PropTypes.string,
+};
+
+Checkbox.defaultProps = {
+  width: [1/2, 1/4], // eslint-disable-line space-infix-ops
```
cbwebdevelopment, line 71: For the other components, to avoid having to do the eslint exceptions, I've since changed to using percentage-based widths.
```js
  width: ['50%', '25%'],
```

Up to you, though.  This is perfectly valid.

PR #676, opened 2020-03-16
[WEB-592] Dropdown Select Component

cbwebdevelopment:
See [WEB-592] for details

I ended up sticking with the rebass select input, and placing the icon with CSS.  I decided to define some of the style props with the rebass `sx` property, as well as the `width` and `color` props.  I found that this is more flexible than doing these default styles in the styled component root, since the specificity of the resulting styled component cannot easily be overridden by the standard rebass props.  I think this is a pattern worth considering as we go forward, but would to hear any thoughts/cautions you may have.

As well, I found and  added a really cool accessibility storybook addon.  Plug'n'Play, and will provide us with great insights.

[WEB-592]: https://tidepool.atlassian.net/browse/WEB-592


app/components/elements/Select.js
```diff
@@ -0,0 +1,106 @@
...
+  options: PropTypes.arrayOf(PropTypes.shape({
+    value: PropTypes.string.isRequired,
+    label: PropTypes.string.isRequired,
+  })).isRequired,
+};
+
+Select.defaultProps = {
+  width: ['100%', '75%', '50%'],
+  color: colors.text.primary,
+  sx: {
```
krystophv: It looks like `defaultProps` doesn't do a sort of _.extend or _.merge, so if we provide a value for `sx` that doesn't include _all_ of these object properties, then the remaining ones are lost. We may need to break these out?

cbwebdevelopment: I agree.  I took a better approach on the Icons that allowed easy overridding of the default rebass styles.  I'll do the same here.

cbwebdevelopment: I've moved the default styles to a new `inputs.select` variant.

There are a number of styles consistent with the `TextInput` component, and I'm wondering if it's worth migrating the `TextInput` styling over to an `inputs.text` variant, which will afford it the same flexibility we're gaining here.

krystophv: It probably is. Once these get merged in, I'll likely do that as part of the search component.

PR #683, opened 2020-03-23
[WEB-754] Added warning for international Dexcom users

thacherT1D:
Added a note for international Dexcom users who try to link their clarity account

### DIFF
--- a/app/components/datasources/datasources.js
+++ b/app/components/datasources/datasources.js
@@ -214,13 +214,13 @@ export default translate()(class DataSources extends Component {
   }
 
   renderDataSourceWarningMessage(provider) {
-    if(provider.id === 'oauth/dexcom') {
-      return(
+    if (provider.id === 'oauth/dexcom') {
+      return (
         <span>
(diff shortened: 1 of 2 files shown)


app/components/datasources/datasources.js
```diff
@@ -213,6 +213,17 @@ export default translate()(class DataSources extends Component {
     )
   }
 
+  renderDataSourceWarningMessage(provider) {
+    if(provider.id === 'oauth/dexcom') {
+      return(
```
cbwebdevelopment: Just a couple whitespace issues here:
```js
    return (
        <span>
          *For US Dexcom users only. Please contact <a href="mailto:support@tidepool.org">
          support@tidepool.org</a> if you live outside the United States.
        </span>
      )
```

cbwebdevelopment: Just a couple whitespace nitpicks here.  Should be space after the `if` and `return` keywords, and closing parentheses should line up with the `return`

thacherT1D: Thanks 👍-- need to figure out why my editor linting isn't catching that for me

PR #691, opened 2020-03-27
[WEB-584] Text Link Component

thacherT1D:
See [WEB-584]

[WEB-584]: https://tidepool.atlassian.net/browse/WEB-584


app/components/elements/FontStyles.js
```diff
@@ -205,9 +205,10 @@ export const TopNavLink = Styled(NavLink)`
 
 export const TextLink = Styled(Link)`
   color: ${colors.text.link};
-  text-decoration: none;
-  line-height: ${lineHeights[0]};
+  font-size: ${fontSizes[1]}px;
```
cbwebdevelopment: Should we specify a font size?  Would it be better to omit it and let the link size be determined by the parent?

app/components/elements/FontStyles.js
```diff
@@ -205,9 +205,10 @@ export const TopNavLink = Styled(NavLink)`
...
   font-family: ${fonts.default};
+  line-height: ${lineHeights[0]};
```
cbwebdevelopment: As with font size, perhaps we should let the line height be determined by the parent context

app/components/elements/FontStyles.js
```diff
@@ -205,9 +205,10 @@ export const TopNavLink = Styled(NavLink)`
...
+  text-decoration: none;
   &:hover {
```
cbwebdevelopment: Are the hover styles actually required here, since they don't overwrite any of the default styles?

app/components/elements/TextLink.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
```
cbwebdevelopment: This file is not really required.

The `TextLink` could just be imported and used directly from `FontStyles`

app/components/elements/TextLink.js
```diff
@@ -0,0 +1,23 @@
...
+    <div>
+      <Box>
+        <Text href={link} {...textLinkProps}>{linkText}</Text>
+      </Box>
+    </div>
+  );
+};
+
+TextLink.propTypes = {
+  label: PropTypes.string.isRequired,
```
cbwebdevelopment: This prop is not used

app/components/elements/TextLink.js
```diff
@@ -0,0 +1,23 @@
...
+  link: PropTypes.string,
```
cbwebdevelopment: Would t would be better to just pass in as `href`, rather than renaming the passed-in prop as it gets applied to the element?

app/components/elements/FontStyles.js
```diff
@@ -205,9 +205,10 @@ export const TopNavLink = Styled(NavLink)`
 
 export const TextLink = Styled(Link)`
```
cbwebdevelopment: Just a thought, but I think it would be more flexible to not bother with using styled-components for this element and sticking with the rebass `Link` element on it's own, and styling it via a theme variant.

Would allow this PR to be reduced to updating the `baseTheme.js` export to 
```js
export default {
  ...
  link: {
    fontFamily: fonts.default,
    textDecoration: 'none',
    color: colors.text.link,
  },
 ...
};
```

Then, in the storybook, just import `Link` directly from `rebass/styled-components` and it will automatically use the appropriate theme styles.

It would also allow us to leverage theme variants (say, if we wanted to add an `underlined` variant, or an `inverted` variant for links on colored backgrounds) if the need arises, with a couple simple changes to the base theme.

app/components/elements/TextLink.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Box } from 'rebass/styled-components';
+import { TextLink as Text } from '../elements/FontStyles';
+
+export const TextLink = props => {
+  const { linkText, link, ...textLinkProps } = props;
+
+  return (
+    <div>
```
cbwebdevelopment: We shouldn't be wrapping the element in 2 divs.  This prevents the link from being used inline.

app/components/elements/FontStyles.js
cbwebdevelopment: To illustrate how I would add variants, I uploaded a `web-584-text-link-alt` branch that includes a couple of mock variants, and some rough storybook examples of how the various ways we could go about the Links differ (i.e. the `TextLink` component vs. using the `TextLink` styled component export from `FontStyles.js` vs. using the rebass `Link` directly with the theme variants.

thacherT1D: Thanks for the guidance -- I've gone ahead and made the suggested changes 👍

app/components/elements/FontStyles.js
```diff
@@ -205,9 +205,8 @@ export const TopNavLink = Styled(NavLink)`
...
```
cbwebdevelopment: I think we should remove the `TextLink` styled-component here, as it's not being used, and it may confuse us down the road (which component to use for links)

app/components/elements/TextLink.js
cbwebdevelopment: Let's also remove the `TextLink.js` file, since it's not being used

stories/TextLink.stories.js
```diff
@@ -0,0 +1,41 @@
...
+  title: 'Text Link',
+  component: Link,
+  decorators: [withDesign, withKnobs, withTheme],
+};
+
+const linkText = () => text('Link Text', 'Link Text');
+const link = () => text('URL', '');
+
+export const TextLinkStory = () => (
+  <React.Fragment>
```
cbwebdevelopment: No big deal here, but no need to wrap in `React.Fragment` since there's only one root element being rendered.

cbwebdevelopment: Can change or not, at your discretion.

thacherT1D: makes sense to have it as simple as possible

app/components/elements/TextLink.js
thacherT1D: thanks for the catch 🤦‍♂️

app/components/elements/FontStyles.js
thacherT1D: makes sense -- especially following making the best decisions for web rather than deferring to the prototype (as we talked about when starting components)

PR #702, opened 2020-04-08
[WEB-593] Checkbox

thacherT1D:
Small CSS update to styled-component, missed when updating to PR recommendations before merging -- https://tidepool.atlassian.net/browse/WEB-593?atlOrigin=eyJpIjoiNTgzMjJkODdiNGExNDJjM2JkMDVkMDQwMmU4ODRhOTEiLCJwIjoiaiJ9

### DIFF
--- a/app/components/elements/Checkbox.js
+++ b/app/components/elements/Checkbox.js
@@ -22,8 +22,8 @@ const StyledCheckbox = styled(Base)`
   &.checked {
     color: ${colors.purpleMedium};
     background-color: ${colors.white};
-    height: 1.5em;
-    width: 1.5em;
+    height: 1em;
+    width: 1em;
   }
 
   &.disabled {


app/components/elements/Checkbox.js
```diff
@@ -22,8 +22,8 @@ const StyledCheckbox = styled(Base)`
   &.checked {
     color: ${colors.purpleMedium};
     background-color: ${colors.white};
-    height: 1.5em;
-    width: 1.5em;
+    height: 1em;
```
cbwebdevelopment, line 25: Quick question: do the width and height rules need to be specified here at all?  Would they not just be inherited by the ones above?

thacherT1D: It's something about the way rebass does it, they are actually separate svg elements, which in theory would be affected by the same parent element, but in practice I think there is something that's overruling it, so this was the best way I could find to handle it.

PR #708, opened 2020-04-14
[WEB-758] Pagination component

cbwebdevelopment:
See [WEB-758] for details

Includes `default` and `condensed` variants, and a new `pagination` button variant.  As well, I've dropped the `ButtonFont` styled component and implemented it as default styles within the button variants, to allow easier overrides.

I looked at a number of pagination libraries, but found MUI pagination was actually a smaller import cost than the others, and allowed me to control the markup via a hook it exposes which was helpful for managing the variants in the Figma spec.

[WEB-758]: https://tidepool.atlassian.net/browse/WEB-758

### DIFF
--- a/app/components/elements/Pagination.js
+++ b/app/components/elements/Pagination.js
@@ -1,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { default as Base, PaginationProps } from '@material-ui/lab/Pagination';
+import { borders, colors, space } from '../../themes/baseTheme';
+
+const StyledPagination = styled(Base)``;
+
+export const Pagination = props => {
+  return <StyledPagination {...props} />;
+};
+
+Pagination.propTypes = {
+  ...PaginationProps,
+};
+
+export default Pagination;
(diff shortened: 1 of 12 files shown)


app/components/elements/Pagination.js
```diff
@@ -0,0 +1,98 @@
...
+            {map(items, ({ page, type, selected, ...item }) => {
+              const pageClassNames = cx({
+                selected,
+              });
+
+              const itemId = type === 'page' ? `${id}-${page}` : `${id}-${type}`;
+              let children;
+
+              if (type === 'start-ellipsis' || type === 'end-ellipsis') {
+                children = <Button disabled className="ellipsis" variant="pagination"><MoreHorizRoundedIcon /></Button>;
```
krystophv, line 56: The a11y check in Storybook isn't terribly happy with this not having any text: `Element does not have inner text that is visible to screen readers`

cbwebdevelopment: Good catch 👍 
I'll address this before merging.

cbwebdevelopment: Fixed.  Now leveraging our `Icon` component which enforces the aria attributes.

PR #710, opened 2020-04-14
[WEB-586] Avatar Component

thacherT1D:
Jira Ticket: https://tidepool.atlassian.net/browse/WEB-586


app/components/elements/Avatar.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Text, Box } from 'rebass/styled-components';
```
cbwebdevelopment, line 3: One 'nice-to-have' is to include the Prop Types of the root-level rebass component we're using, as it will auto-populate the props in most IDEs.

```js
import { Text, Box, BoxProps } from 'rebass/styled-components';
```
and then in the PropTypes below:
```js
Avatar.propTypes = {
  ...BoxProps,
  initials: PropTypes.string.isRequired,
  variant: PropTypes.oneOf(['default', 'inverse']),
  label: PropTypes.string.isRequired,
};
```

PR #711, opened 2020-04-20
[WEB-585] Table Component

cbwebdevelopment:
See [WEB-585] for details.

Storybook example includes table sorting, filtering by columns by nested data (such as patient name and email), ability to specify render functions for any column, sticky header and row hover options.

Also includes some consolidation of the Select and TextInput component styles, and some fixes to allow easier positioning of them via rebass props.

[WEB-585]: https://tidepool.atlassian.net/browse/WEB-585


cbwebdevelopment:
Yeah, I found it to be a tough call what logic to include within the Table, but landed on a slight preference to keep it modular for the time being.

PR #714, opened 2020-04-22
[WEB-785] add email verification resend banner

krystophv:
For [WEB-785], adds a banner for clinicians viewing accounts with emails that are currently unclaimed allowing the clinician to resend the verification email.

Clears store state regarding sent status on `CLEAR_PATIENT_IN_VIEW` and `FETCH_PATIENT_SUCCESS`.

[WEB-785]: https://tidepool.atlassian.net/browse/WEB-785

### DIFF
--- a/app/components/sendverificationbanner/sendverificationbanner.js
+++ b/app/components/sendverificationbanner/sendverificationbanner.js
@@ -10,7 +10,7 @@ const SendVerificationBanner = (props) => {
     trackMetric,
     resendVerification,
     t,
-    resendEmailVerificationProgress: working,
+    resendEmailVerificationInProgress: working,
     resentEmailVerification: resent,
   } = props;
 
@@ -57,7 +57,7 @@ SendVerificationBanner.propTypes = {
   trackMetric: PropTypes.func.isRequired,
   patient: PropTypes.object.isRequired,
   resendVerification: PropTypes.func.isRequired,
-  resendEmailVerificationProgress: PropTypes.bool.isRequired,
+  resendEmailVerificationInProgress: PropTypes.bool.isRequired,
   resentEmailVerification: PropTypes.bool.isRequired,
 };
 
--- a/app/components/sendverificationbanner/sendverificationbanner.js
+++ b/app/components/sendverificationbanner/sendverificationbanner.js
@@ -1,0 +1,64 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import { translate } from 'react-i18next';
+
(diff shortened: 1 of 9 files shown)


app/components/sendverificationbanner/sendverificationbanner.js
```diff
@@ -0,0 +1,64 @@
...
+
+import personUtils from '../../core/personutils';
+
+const SendVerificationBanner = (props) => {
+  const {
+    patient,
+    trackMetric,
+    resendVerification,
+    t,
+    resendEmailVerificationProgress: working,
```
cbwebdevelopment, line 13: Wondering if this would be better as `resendingEmailVerification` or `resendEmailVerificationInProgress`
The name of this prop doesn't suggest to me that it would be a Boolean.  

No great concern here, and if you feel otherwise, just merge it as-is.

test/unit/app/app.test.js
```diff
@@ -596,6 +603,14 @@ describe('App', () => {
       it('should return null for permsOfLoggedInUser', () => {
         expect(result.permsOfLoggedInUser).to.be.null;
       });
+
+      it('should map working.resendingEmailVerification.inProgress to resendingEmailVerification', () => {
```
cbwebdevelopment, line 607: Looks like you used `resendingEmailVerification` as the expected prop here.  This should be renamed to match the prop, or vice-versa.

test/unit/components/sendverificationbanner.test.js
```diff
@@ -0,0 +1,66 @@
...
+
+  it('should render a resend verification button', () => {
+    const expectedText = 'RESEND VERIFICATION EMAIL'
+    const verificationButton = wrapper.find('button');
+
+    expect(verificationButton).to.have.length(1);
+    expect(verificationButton.text()).contains(expectedText);
+  });
+
+  it('should track when the add email button is clicked', () => {
```
cbwebdevelopment, line 53: Should be when the `resend verification` button is clicked in description

krystophv: Yup, that's what I get for copy paste :)

app/components/sendverificationbanner/sendverificationbanner.js
krystophv: Yeah, going with the clearer and more verbose. Same as the unit test below.

test/unit/app/app.test.js
krystophv: Matching the new renamed prop above

PR #720, opened 2020-04-24
[WEB-803] Hotfix: Ignore future data

cbwebdevelopment:
See [WEB-803] for details

[WEB-803]: https://tidepool.atlassian.net/browse/WEB-803


app/redux/actions/async.js
```diff
@@ -940,25 +942,57 @@ export function fetchPatientData(api, options, id) {
...
           // diabetes datum time and going back 30 days
-          const diabetesDatums = _.reject(results, d => _.includes(['food', 'upload'], d.type));
-          const latestDatumTime = _.max(_.map(diabetesDatums, d => (d.time)));
-          options.startDate = moment.utc(latestDatumTime || options.browserTimeStub).subtract(30, 'days').startOf('day').toISOString();
-          options.endDate = moment.utc(latestDatumTime || options.browserTimeStub).add(1, 'days').toISOString();
+          const diabetesDatums = _.reject(resultsVal.latestDatums, d => _.includes(['food', 'upload'], d.type));
+          const latestDiabetesDatumTime = _.max(_.map(diabetesDatums, d => (d.time)));
+
+          if (moment.utc(latestDiabetesDatumTime).diff(moment.utc(serverTime), 'days') > 1) {
+            _.isFunction(rollbar.error) && rollbar.error(
```
krystophv, line 972: Should we consider using `rollbar.info` (a la https://github.com/tidepool-org/uploader/blob/2752f10c2a4f741785d5811c9cf3929a7fa176d0/lib/commonFunctions.js#L144 ) instead of treating this as an `Error`, since it's largely informational (no specific action for us to take/fix)

cbwebdevelopment: Good idea.  I'll change it.

PR #727, opened 2020-05-05
[WEB-762] accordion

krystophv:
[WEB-762] Accordion component

[WEB-762]: https://tidepool.atlassian.net/browse/WEB-762

### DIFF
--- a/app/components/elements/Accordion.js
+++ b/app/components/elements/Accordion.js
@@ -5,9 +5,9 @@ import { Box, BoxProps } from 'rebass/styled-components';
 import { default as ExpansionPanel, ExpansionPanelProps } from '@material-ui/core/ExpansionPanel';
 import { default as ExpansionPanelSummary, ExpansionPanelSummaryProps } from '@material-ui/core/ExpansionPanelSummary';
 import { default as ExpansionPanelDetails, ExpansionPanelDetailsProps } from '@material-ui/core/ExpansionPanelDetails';
-import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
+import ExpandMoreRoundedIcon from '@material-ui/icons/ExpandMoreRounded';
 
-import { borders, colors, space, fonts, fontSizes } from '../../themes/baseTheme';
+import { borders, colors, space, fonts, fontSizes, fontWeights } from '../../themes/baseTheme';
 
 const StyledAccordion = styled(ExpansionPanel)`
   font-family: ${fonts.default};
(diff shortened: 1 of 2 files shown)


app/components/elements/Accordion.js
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { Box, BoxProps } from 'rebass/styled-components';
+import { default as ExpansionPanel, ExpansionPanelProps } from '@material-ui/core/ExpansionPanel';
+import { default as ExpansionPanelSummary, ExpansionPanelSummaryProps } from '@material-ui/core/ExpansionPanelSummary';
+import { default as ExpansionPanelDetails, ExpansionPanelDetailsProps } from '@material-ui/core/ExpansionPanelDetails';
+import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
```
cbwebdevelopment, line 8: We've been using the rounded variants of all the icons as per design's request:
```js
import ExpandMoreRoundedIcon from '@material-ui/icons/ExpandMoreRounded';
```

app/components/elements/Accordion.js
```diff
@@ -0,0 +1,79 @@
...
+    background-color: ${colors.lightestGrey};
+  }
+`;
+
+const StyledAccordionContent = styled(ExpansionPanelDetails)`
+  padding: ${space[3]}px;
+  color: ${colors.text.primarySubdued}
+`;
+
+export const Accordion = (props) => {
```
cbwebdevelopment, line 39: I played around with this a bit locally, and I think there are a few improvements we can make here.

The first is to use `Box` components and leverage it's `as` prop.  This will allow us to easily apply the rebass theming, while still using the styled expansion panel components.

As well, we should spread the main props supplied to the component to allow us to control the panels via the `onChange` and `expanded` props.  I think that's an important functionality to have, especially if we call it an accordion, which oftentimes is configurable to close others when one is expanded.

On that note, I'm wondering if we should change the component name to `ExpansionPanel`, and use it to build and control accordions as required.

As a small change, I defaulted the header to the `500` font weight as well to match the spec

Here's what I ended up with locally:
```js
export const Accordion = (props) => {
  const { header, children, themeProps, icon, label, ...accordionProps } = props;

  return (
    <Box square as={StyledAccordion} {...themeProps.wrapper} {...accordionProps}>
      <Box
        as={StyledAccordionHeader}
        expandIcon={icon}
        aria-controls={`${label}-content`}
        id={`${label}-header`}
        fontWeight={fontWeights.medium}
        IconButtonProps={{
          disableFocusRipple: true,
          disableRipple: true,
        }}
        {...themeProps.header}
      >
        {header}
      </Box>
      <Box
        as={StyledAccordionContent}
        id={`${label}-content`}
        {...themeProps.panel}
      >
        {children}
      </Box>
    </Box>
  );
};
```

Lot of small changes here, and just my own suggestions.  I've pushed my local changes to a `web-762-accordion-alt` branch in case you'd like to review them locally.  There are a few changes I made to the storybook as well while playing around with this, such as a controlled example and a disabled on (which currently has no styling -- and it's possible that we never need it -- was just trying it out)

krystophv: Ok, yeah. I can see the appeal of going with `Box` by default, it's just not yet intuitively my go-to. I also figured we _might_ want to be dealing with a higher-order (eg "Accordion") component, but it wasn't clear from the design either way, so figured simpler was the first path to take and then build out the HOC later if necessary, but this strikes a decent middle ground that'll allow us to have the HOC easier if we want it. I'll probably just snag the commit and pull it into this branch. I don't see any reason not to accept all the changes suggested.

PR #731, opened 2020-05-13
[WEB-760] Toggle Component

thacherT1D:
WEB-760 Jira Ticket: https://tidepool.atlassian.net/browse/WEB-760


package.json
```diff
@@ -33,7 +33,7 @@
     "@material-ui/core": "4.9.10",
     "@material-ui/icons": "4.9.1",
     "@material-ui/lab": "4.0.0-alpha.49",
-    "@rebass/forms": "4.0.6",
+    "@rebass/forms": "^4.0.6",
```
cbwebdevelopment, line 36: You'll need to remove this carat here so the version is locked down.

PR #746, opened 2020-06-04
Release 1.35.0 to master

krystophv:
Release includes [WEB-829] and [WEB-811]

[WEB-829]: https://tidepool.atlassian.net/browse/WEB-829
[WEB-811]: https://tidepool.atlassian.net/browse/WEB-811


krystophv:
@cbwebdevelopment I needed to tweak the resolve mechanism for `blip` to be less heavy handed in looking at `blip`'s copy of a dependency vs. a dependency's copy so `platform-client` would resolve its own `superagent` instead of `blip`'s (which gets hoisted from `hakken` and is the wrong version). I think that's the only new code in the release branch and should probably be reviewed.

PR #752, opened 2020-06-30
[WEB-822] Prescription review form step

cbwebdevelopment:
See [WEB-822] for details.

Includes a few requested additions, such as displaying any guardrail warnings, and allowing the clinician to navigate back to appropriate steps to edit fields and return to the review step.

[WEB-822]: https://tidepool.atlassian.net/browse/WEB-822


cbwebdevelopment:
@krystophv FYI: Though I try to avoid adding commits after requesting review, I noticed a regression on the back button that needed fixing.

I also realized that it would  be much better if the form automatically loaded at the latest incomplete step (or the review step if all are complete) to avoid the need for clinicians to click their way through steps they've already completed for draft (or submitted) prescriptions.

cbwebdevelopment:
@krystophv I've rerequested a review, since I've made an update to only render the prescription routes when enabled via a feature flag.

I'd rather not have the routes active in our production builds until we're ready to go.

Relevant commit is #0cb45fa

PR #773, opened 2020-07-24
[WEB-716] Change download buttons to use pre-built browser_download_url

thacherT1D:
Jira Ticket: https://tidepool.atlassian.net/browse/WEB-716


thacherT1D:
@krystophv Rolled back to using a class component instead of using a functional component because of complications I started to run into with testing -- confirmed with Clint that this was the better solution for right now. Additionally, wrapping the Button component in <a> tags is definitely not preferable, but based on where we're at it seems like the best solution for the moment -- if you think there's a better way to handle it please let me know 🙂

PR #778, opened 2020-07-29
[WEB-893] Dynamic lower correction range limit

cbwebdevelopment:
See [WEB-893] for details.

I had to re-enable the commented-out `suspendThreshold` field, which is not yet implemented in the backend, but should be soon (via tidepool-org/platform#462).  Currently, this means the suspendThreshold input works, but the value will not be saved to the backend.

This also contains a few small fixes for issues I noticed with the therapy settings section while developing:

1. The threshold warnings now only show when numerical values are present.  This avoids them showing when set to a default empty string value.
2. The input step props were missing from the correction range schedule inputes
3. In addition to the logic defined by the Jira ticket, enforcing that the low correction range targets are not below the suspend threshold, I've also set it so that a High target cannot be set lower than a Low target.

[WEB-893]: https://tidepool.atlassian.net/browse/WEB-893

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -14881,7 +14881,7 @@ reduce-component@1.0.1:
   resolved "https://registry.yarnpkg.com/reduce-component/-/reduce-component-1.0.1.tgz#e0c93542c574521bea13df0f9488ed82ab77c5da"
   integrity sha1-4Mk1QsV0UhvqE98PlIjtgqt3xdo=
 
-"reductio@github:crossfilter/reductio#0c45b03559":
+reductio@crossfilter/reductio#0c45b03559:
   version "0.6.3"
   resolved "https://codeload.github.com/crossfilter/reductio/tar.gz/0c45b0355950101d395d6482e2083ce1cdbcf435"
   dependencies:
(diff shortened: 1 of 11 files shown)


yarn.lock
```diff
@@ -14881,7 +14881,7 @@ reduce-component@1.0.1:
   resolved "https://registry.yarnpkg.com/reduce-component/-/reduce-component-1.0.1.tgz#e0c93542c574521bea13df0f9488ed82ab77c5da"
   integrity sha1-4Mk1QsV0UhvqE98PlIjtgqt3xdo=
 
-"reductio@github:crossfilter/reductio#0c45b03559":
+reductio@crossfilter/reductio#0c45b03559:
```
krystophv, line 14884: I notice that this goes back and forth between these two formats pretty often and I'm really not sure why. It clearly doesn't make any functional difference, but it's a strange behavior that I'd eventually like to know the reason for.

cbwebdevelopment: Yeah, I'm not sure why either.  I'm wondering if it's a yarn version thing.

PR #792, opened 2020-08-22
[WEB-979] Add success toast upon completion of prescription form

cbwebdevelopment:
See [WEB-979] for details

[WEB-979]: https://tidepool.atlassian.net/browse/WEB-979


cbwebdevelopment:
Regarding stacking the Toasts, as you mentioned in your edit, MUI recommends against it from a UX perspective, though they do send an example library that implements it for those that desire/need the functionality:
https://material-ui.com/components/snackbars/#notistack

If the time comes that we do need it, we could update the ToastProvider to allow pushing/unshifting of toasts, and either using `notistack` or handling the stacking in our own component (where the only tricky bit would be the animations/positioning)

PR #829, opened 2020-09-29
[WEB-981] Customizable print date ranges

cbwebdevelopment:
See [WEB-981] for details.  Involved an increased scope over the original print dialog, as it seemed necessary to be able to set different date ranges per chart type.  Also added was the ability to disable chart types.

The PDF is no longer generated ahead of time.  It is now generated when the print dialog is submitted.  There were 2 main scenarios to account for upon submission:

1. If the earliest date requested for any of the print charts was within the already-fetched data, we simply generate the PDF and open it in a new window when ready.
2. If the earliest date requested was outside of the already-fetched data, we store the print options submitted, fetch data to the required date, and once the data is added to the data worker, we then generate the PDF

Other small updates include:

1. Updating the Switch rebass component styling to match our Figma spec
2. A few small theming improvements/fixes to Accordion, Date Range and Dialog components
3. Removal of dynamic loading icon from header "Print" button.

Related PR: tidepool-org/viz#232

[WEB-981]: https://tidepool.atlassian.net/browse/WEB-981


thacherhussain (approved):
LGTM 🦄 -- pending Travis' expected check and the `package.json` / `yarn.lock` conflicts

PR #830, opened 2020-10-01
Fix for broken production config stemming from webpack html plugin bump

cbwebdevelopment:
Thanks for finding the fix so quick @krystophv
Just wrapped it up in a PR for your review


krystophv (approved):
LGTM 👍

PR #832, opened 2020-10-07
[WEB-54] Custom Basics date ranges

cbwebdevelopment:
See [WEB-54]

Implements a custom date range picker on the basics view via a calendar icon beside the dates.  Similar to the implementation of the print view date dialog, we have a 90-day limit, and identical defaults/presets to the print view to facilitate a 1:1 experience.

Supporting PR: tidepool-org/tideline#408

[WEB-54]: https://tidepool.atlassian.net/browse/WEB-54

### DIFF
--- a/.nlu.example.json
+++ b/.nlu.example.json
@@ -18,6 +18,7 @@
     "/.git/"
   ],
   "list": [
+    "blip",
     "tidepool-platform-client",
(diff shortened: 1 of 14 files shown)


thacherhussain (approved):
LGTM 🦄

PR #833, opened 2020-10-08
Replace nlu linking by leveraging webpack aliases

cbwebdevelopment:
I felt there may be a better way to manage local npm linking than NLU.  The main thing I like about NLU is that links are managed in a declarative fashion.  It did have 3 shortcomings, IMO, though:

1. It doesn't work for Windows -- no biggie, but...
2. It was still kind of slow when linking/unlinking via `nlu run`
3. Because the config file was JSON, we couldn't just comment out packages, but had to manually add and remove them.  Who wants to type ?!?

I came across this article: https://medium.com/@tnrich_29519/a-better-alternative-to-npm-yarn-link-for-the-front-end-667f03d497a6, which advocated for leveraging webpack aliases for this, and it seems to work great in my testing. 

It solves all 3 shortcomings, and I've set it up in a way that will only add these aliases if we're running in `DEV` mode, which I think is ideal, since we obviously don't want it for `PRD` builds, and we don't need to forget to unlink when running unit tests.

Looking for feedback.


cbwebdevelopment:
Just to note: I initially had missed some local changes to the example file in my initial commit for the PR

webpack.config.js
```diff
@@ -10,11 +10,15 @@ const terser = require('terser');
 const fs = require('fs');
 const pkg = require('./package.json');
 const cp = require('child_process');
+const optional = require('optional');
 
 const isDev = (process.env.NODE_ENV === 'development');
 const isTest = (process.env.NODE_ENV === 'test');
 const isProd = (process.env.NODE_ENV === 'production');
 
+const linkedPackages = isDev ? optional('./config/linked-packages').packages || {} : {};
```
krystophv, line 19: Should this be:
```suggestion
const linkedPackages = (isDev || isTest) ? optional('./config/linked-packages').packages || {} : {};
```
to handle the case that a unit test's success depends on changes in a locally linked and yet unpublished package?

krystophv: Actually, I see that will probably have implications on the CI servers. Perhaps we could use a `isCI` check? Not sure if that's adding too much complexity though.

cbwebdevelopment: Well, in my description, I noted that I'd initially excluded TEST on purpose.  I suppose I'm good either way though.  What I wanted to avoid, and what has happened, is having tests pass, then having them fail in CI since the dependancy wasn't published.

That's just my workflow, though -- I like to get the tests working in the linked package, then publish and run the test-watcher in `blip`.

I'll commit the suggestion.  Makes sense to have it allow for the workflow you've outlined

cbwebdevelopment: Ah, you're right about the CI issue.  And I've already committed the suggestion while you were typing!  Let me look into the CI check option.  Otherwise, we could add a secondary means of running the tests with the linked packages, but not by default, which may be less flaky than whatever CI check I come up with.

Thoughts?

cbwebdevelopment: Actually, I don't think it would be an issue, since the linked-packages.js file is excluded from VC, and would not be present in CI builds, and would thus just spread an empty object into the aliases as intended.

Gonna revert my revert :)

cbwebdevelopment: Tests passing in CI (build is not finished as of time of writing this, but the `npm test` portion has) with latest commit.  I think we're good on this front.

krystophv: Cool, seems good to me. If forgetting to publish packages becomes a common issue, then we can look into ways to solve it (whether that's moving to local/linked-specific testing `npm scripts` or something else).

cbwebdevelopment: It's honestly not a problem for me, and it would only result a failed CI build.  I have seen interns trip over it, so that's all I had in mind -- kind of avoiding that scenario for those who may not be mindful of it.

After giving it some thought, I see no real downside to it. and could be seen as a convenience at times.

krystophv (approved):
LGTM 🚀

thacherhussain (approved):
LGTM 🦄

PR #835, opened 2020-10-09
[WEB-54] Set basics dates default extent size to 14 days, and send metrics on date changes

cbwebdevelopment:
Some additional requirements stemming from the comments on [WEB-54]

Default dates for basics is now the latest 14 days to match the new print view defaults.

As well, added a kissmetric event to track date changes, and a note regarding the 90 days max custom range.

[WEB-54]: https://tidepool.atlassian.net/browse/WEB-54


thacherhussain (approved):
LGTM 🦄

PR #841, opened 2020-10-15
[WEB-1064] Fix for double clicking datums on BG Log not navigating to correct date in Daily

cbwebdevelopment:
See [WEB-1064] 

Basicslly, the BG Log view was sending the date as a UTC timestamp, but the handler for switching to Daily was only working when the provided datetime was an ISO string.  The handler now works with both formats.

[WEB-1064]: https://tidepool.atlassian.net/browse/WEB-1064


thacherhussain (approved):
LGTM 🦄

PR #843, opened 2020-10-20
[WEB-993] Add option to exclude days with no boluses from basics calendar aggregations and stats

cbwebdevelopment:
See [WEB-993]

This amalgamates the Basal bolus stat with the total insulin stat (as outlined in comments in Jira ticket), and adds a checkbox to exclude days without boluses in the Basics date range from chart aggregations and those stats.

When checked, it will also cause the calendar days to adopt the 'out of range' styling in both the web and print.  The calculations will also be affected in the copy-to-text output, and will have a note added regarding the omission of empty bolus days.

Goes with:
tidepool-org/viz#234
tidepool-org/tideline#409

[WEB-993]: https://tidepool.atlassian.net/browse/WEB-993


test/unit/pages/patientdata.test.js
```diff
@@ -2791,6 +2853,39 @@ describe('PatientData', function () {
...
+  describe('toggleDaysWithoutBoluses', () => {
+    let wrapper;
+    let instance;
+
+    beforeEach(() => {
+      wrapper = shallow(<PatientData.WrappedComponent {...defaultProps} />);
+      instance = wrapper.instance();
+    });
+
+    it('should call `updateChartPrefs with the `excludeDaysWithoutBolus` chartPrefs state toggled', () => {
```
krystophv, line 2865:
```suggestion
    it('should call `updateChartPrefs` with the `excludeDaysWithoutBolus` chartPrefs state toggled', () => {
```

krystophv (approved):
LGTM 👍

PR #844, opened 2020-10-22
[WEB-981] Pre-open a PDF window async data fetch is required

cbwebdevelopment:
Small fix required for [WEB-981], where in scenarios that an async data fetch was required prior to generating a PDF report, the print window would be blocked by the browser's popup blocker.

The solution is to open the new window BEFORE sending the request for earlier data, and then updating the `location.href` of it once the data is received and the PDF is ready.  We show a short message in the new blank window while the user waits.

In cases where this pre-fetching is not required, the workflow has not been updated.  It will stay on the print modal until the processing is complete, and then open the new tab.

While implementing this fix, I also noticed some inconsistencies on hover states and flex positioning within the updated header bar, so I've cleaned up some styles as well.

[WEB-981]: https://tidepool.atlassian.net/browse/WEB-981

### DIFF
--- a/app/components/chart/header.js
+++ b/app/components/chart/header.js
@@ -4,8 +4,8 @@ import React, { Component } from 'react';
 import cx from 'classnames';
 import { translate } from 'react-i18next';
 import DateRangeRoundedIcon from '@material-ui/icons/DateRangeRounded';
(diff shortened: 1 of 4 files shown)


thacherhussain (approved):
LGTM 🦄

PR #851, opened 2020-11-02
[WEB-1089] Catch pdf window focus errors caused by popup blocker

cbwebdevelopment:
Hotfix. See [WEB-1089] for details.

Mainly, this PR will catch errors caused by the popup blocker being invoked when a request to generate a PDF takes too long, and display a notification to the user.

It also adds an action prop to the Toast component that can be used to render a CTA within a notification, and fixes a couple of warnings on the proptypes for the print and chart date range modals, since some of the props are not immediately available on first rendering.

The fix on this one was relatively easy.  Getting the patientdata unit tests working ended up being a rather large trial, due to a 'perfect storm' of sorts in the way the ToastProvider context had to be applied to the 'create-react-class' component, which didn't play nicely with i18n's context within the enzyme mounts.  I ended up having to export the `create-react-class` component separately, and pass it to the translate HOC wrapped in a minimal functional component.

[WEB-1089]: https://tidepool.atlassian.net/browse/WEB-1089

### DIFF
--- a/app/components/ChartDateRangeModal.js
+++ b/app/components/ChartDateRangeModal.js
@@ -214,9 +214,9 @@ export const ChartDateRangeModal = (props) => {
 };
 
 ChartDateRangeModal.propTypes = {
-  chartType: PropTypes.string.isRequired,
+  chartType: PropTypes.string,
   maxDays: PropTypes.number.isRequired,
-  mostRecentDatumDate: PropTypes.number.isRequired,
+  mostRecentDatumDate: PropTypes.number,
   onClose: PropTypes.func.isRequired,
   onDatesChange: PropTypes.func.isRequired,
   onSubmit: PropTypes.func.isRequired,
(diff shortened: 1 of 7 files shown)


krystophv (approved):
LGTM 🚀 

Good work on this. The unit testing contortions seem like a pain. It might be worth finding a place to document the conflict and throw out a couple of suggested more permanent fixes for the issue going forward (i.e. which of the libraries we need to update, etc) since I'm sure this won't be the only time we hit the issue.

cbwebdevelopment:
Thanks, Chris.  I agree that we should find and document a more permanent solution.

All in, I probably spent about a half a day just debugging the tests and trying to figure our a better way.  I'm not certain of what that is yet, though I imagine it will likely be one of:

1. Dropping `create-react-class` and moving to class components, which would allow us to define context via a `contextTypes` static class property, instead of the `contextType` which only allows for one.
2. It may be that we don't have to do (1), but rather, updating the `react-i18next` to the latest _may_ fix the conflict.  I started down this road, but with the sheer amount of changes required, felt it was best to avoid this for the time being, given that this is a hotfix.

PR #858, opened 2020-12-11
[WEB-1074] guardrail updates

cbwebdevelopment:
See [WEB-1074]

Includes updated guardrails, as well as new temporary correction ranges [WEB-1062] and updated therapy settings ordering based on the new guardrail dependancies [WEB-1093]

While tackling the new conditional threshold dependancies, it became clear that there were some performance issues to address, and in the process, some reduction in code complexity to apply by removing the creation and use of custom metadata to track form state, and passing around the already provided formik context where needed.  The bulk of these changes are in the following commits (JIRA [WEB-1121]):

- https://github.com/tidepool-org/blip/pull/858/commits/4eb9c14a8637d9ddc671210e578c704bbd26cd8e
- https://github.com/tidepool-org/blip/pull/858/commits/fc7a2f0814e0de1f44baf80b76a9f98c4bb07c6b
- https://github.com/tidepool-org/blip/pull/858/commits/562bc9bb96ba4d8ed45699ac17e4ef3df020a44e

Other misc updates:
1. Showing fields in an error state on the review step similar to how we were showing the threshold warnings.  It's an unlikely scenario for a user to fall into if they follow the form steps, but it can happen if a prescription is saved in a draft state and the error thresholds change (as happened to me here, which is how I noticed it).  It would place a clinician in a state where they could not submit the review step, and would not know why.  This way they can see the field(s) in error and easily edit them to acceptable values
2. Rounding values to the nearest allowed increment on blur for when a user types in the values.

[WEB-1074]: https://tidepool.atlassian.net/browse/WEB-1074

[WEB-1062]: https://tidepool.atlassian.net/browse/WEB-1062
[WEB-1093]: https://tidepool.atlassian.net/browse/WEB-1093
[WEB-1121]: https://tidepool.atlassian.net/browse/WEB-1121


app/pages/prescription/ScheduleForm.js
```diff
@@ -74,23 +84,27 @@ const ScheduleForm = props => {
             innerRef={refs[index]}
             id={`${fieldArrayName}.${index}.start`}
             name={`${fieldArrayName}.${index}.start`}
-            error={getFieldError(fieldArrayMeta, index, 'start')}
+            error={getFieldError(`${fieldArrayName}.${index}.start`, useFormikContext())}
```
krystophv, line 87: The use of a hook here will cause issues when the number of schedules changes since the set of hooks executed per render cycle will change. I see that you're calling the hook pretty often to pass to `getFieldError` - is there an issue with calling the hook to get the context once at the top of the render and then just passing a reference to it here (and elsewhere) as opposed to calling the hook each time? 

The only other solution I've found for this is to break out the component that needs the hook call into its own component.

krystophv (changes requested):
Looks pretty good except for the somewhat obscure Hook requirement where React requires the number and order of hooks called per render be identical across renders. I _think_ it can be fixed relatively easily by just moving the hook call up and passing by reference.

app/pages/prescription/ScheduleForm.js
cbwebdevelopment: Good catch.  No issue with simply passing a common reference.

cbwebdevelopment:
@krystophv Updated to only call useFormikContext once within each component.  
Commit ce9771e

krystophv (approved):
LGTM :shipit:

PR #859, opened 2021-01-08
Webpack local declarative config

cbwebdevelopment:
This expands on the use of local webpack config that we've been using to link npm packages for local development, adding the ability to specify the remote backend environment to connect to, and which feature flags to enable.


krystophv (approved):
Sorry this one slipped through. LGTM 👍

PR #868, opened 2021-01-22
[WEB-1129] Basics date selection fix

cbwebdevelopment:
See [WEB-1129] for details.

Basically the wrong date was being used for requests for earlier data due to a missing `startDate` option being set when requesting a range for the basics view that required earlier data to be fetched.

[WEB-1129]: https://tidepool.atlassian.net/browse/WEB-1129


krystophv (approved):
LGTM 👍

PR #889, opened 2021-04-13
Attempt to address max file watchers warnings that are clogging the travis logs

cbwebdevelopment:
Seems to work.  Previously, Travis logs were getting inundated with error messages from `chokidar` along the lines of:
```
Error: ENOSPC: System limit for number of file watchers reached
```


krystophv (approved):
LGTM 👍🏼

PR #896, opened 2021-05-12
Travis: docker login before pull

cbwebdevelopment:
Docker pull rate limits are capped at 100 for unauthenticated requests.  Authenticating prior to pulling allows us to avoid these rate limit errors during high travis build usage.


krystophv (approved):
LGTM 👍🏼

PR #904, opened 2021-05-21
[WEB-1135] Prescription flow updates

cbwebdevelopment:
See [WEB-1135]

This ticket was originally meant to address some specific additional feeback/changes to the prescription flow coming down from product/design, but it kind of became a catch-all for small changes/improvements.

Rather than wait until product has a finalized list of feedback items, we've decided to merge the current changes back to the develop branch, and any future feedback will be addressed in new Jira tickets.

Here's a quick summary of the changes/additions in this PR:

- New 'onEnter' hook for the Stepper component.  Our first specific use case is to allow us to redirect to the start of the settings calculator step when navigating back from the Therapy settings step, especially important in cases where the calculator was skipped, as the 2nd calculator step is contingent on the first being filled.
- More granular methods for clearing calculator inputs and results - for instance when the calculator method is changed
- Some fixes for blur handling on initially focused inputs that was causing flashes of the error state on checkboxes and radios
- Moving calculator results to it's own component to allow rendering them within the calculator step, and at the top of the therapy settings form
- Some additional validation on the calculator schema fields
- consistently birthdate format throughout
- rearranging the order of some therapy settings form fields
- some general styling fixes for bordered checkbox and radio groups
- Remove bolusAmountMaximum default value

[WEB-1135]: https://tidepool.atlassian.net/browse/WEB-1135


test/unit/pages/prescription/PrescriptionForm.test.js
```diff
@@ -83,19 +85,64 @@ describe('PrescriptionForm', () => {
...
-      clearCalculator(setFieldValue);
-      sinon.assert.calledWithExactly(setFieldValue, 'calculator.method', undefined, false)
-      sinon.assert.calledWithExactly(setFieldValue, 'calculator.totalDailyDose', undefined, false)
-      sinon.assert.calledWithExactly(setFieldValue, 'calculator.totalDailyDoseScaleFactor', undefined, false)
-      sinon.assert.calledWithExactly(setFieldValue, 'calculator.weight', undefined, false)
-      sinon.assert.calledWithExactly(setFieldValue, 'calculator.weightUnits', undefined, false)
-      sinon.assert.calledWithExactly(setFieldValue, 'calculator.recommendedBasalRate', undefined, false)
-      sinon.assert.calledWithExactly(setFieldValue, 'calculator.recommendedInsulinSensitivity', undefined, false)
-      sinon.assert.calledWithExactly(setFieldValue, 'calculator.recommendedCarbohydrateRatio', undefined, false)
+      clearCalculator(formikContext);
```
krystophv, line 131: I think this should also continue to test that `calculator.method` is cleared as well.

krystophv (approved):
Looks like we lost one unit test in the shuffle, but other than that, all LGTM 👍🏼

test/unit/pages/prescription/PrescriptionForm.test.js
cbwebdevelopment: Good catch.  Will get that back in and merge.

PR #937, opened 2021-09-14
[WEB-1219] Update prescription API to use new clinic endpoints

clintonium-119:
See [WEB-1219]

Contains the necessary API updates for clinics prescriptions

**A few other changes to note:**
 - I felt it made sense at this time to remove the `withDevices`, `withPrescriptions`, and `withAssociatedAccounts` HOCs that I'd put together while initially working on the prescription form since I had to make changes to the HOCs to accommodate passing in the `selectedClinicId` state, and I'd hoped to find an opportunity to update to the `useSelector` approach at some point before launch.
 - The `selectedClinicId` state is now persisted to (and hydrated from) `localStorage`.  This was needed since we persist the prescription form state in case of refreshes or interruptions, but needed to have reference to the active clinic upon reload.  This also was something I generally had wanted to implement for some of the other clinic interfaces as well.
 - Some small form label updates to the prescription flow. 

Related PR: https://github.com/tidepool-org/platform-client/pull/127

[WEB-1219]: https://tidepool.atlassian.net/browse/WEB-1219


krystophv (approved):
LGTM 🚀 

I have to admit to being partial to the `useSelector` pattern over HOC's. It's a little more verbose, but I just like having all the state dependencies collected into a single file. Great work!

PR #941, opened 2021-09-23
[WEB-1233] Fix for incorrect account name in NavigationMenu

clintonium-119:
See [WEB-1233]
Navigation menu was incorrectly showing the name from the patient profile rather than the correct name from the account profile.

[WEB-1233]: https://tidepool.atlassian.net/browse/WEB-1233


krystophv (approved):
LGTM 👍🏼

PR #946, opened 2021-10-05
[WEB-1238] notify on resend invite

krystophv:
[WEB-1238] - moving the resend invitation action behind a modal confirmation to avoid accidental spam

### DIFF
--- a/app/pages/share/AccessManagement.js
+++ b/app/pages/share/AccessManagement.js
@@ -17,6 +17,7 @@ import InputIcon from '@material-ui/icons/Input';
 import DeleteForeverIcon from '@material-ui/icons/DeleteForever';
 import PublishRoundedIcon from '@material-ui/icons/PublishRounded';
 import CircularProgress from '@material-ui/core/CircularProgress';
+import sundial from 'sundial';
 
 import {
   Title,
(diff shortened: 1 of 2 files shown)


app/pages/share/AccessManagement.js
```diff
@@ -623,6 +636,42 @@ export const AccessManagement = (props) => {
...
+        <DialogContent>
+          <Body1>
+            {t(
+              'You invited {{inviteName}} to view your data on {{inviteDate}}. Are you sure you want to resend this invitation?',
+              {
+                inviteName: selectedSharedAccount?.name || selectedSharedAccount?.email,
+                inviteDate: formattedInviteDate,
+              }
+            )}
+          </Body1>
```
clintonium-119, line 657: The only thing I notice here is that the Figma design shows the `inviteName` and `inviteDate` in bold text, so it may be worth using the `<Trans />` component here to allow that.  

However, if you do go ahead with this, we should likely bold the email in the "Revoke Invitation" dialog as well to match it.  Up to you.

I also feel it may read a little better if the "Are you sure you want to resend this invitation?" part is on a new line (i.e. in separate `Body1` element), but again, that's more of a question of design intent in the Figma file (i.e. it's on a new line there, but is that the intent, or just coincidental).

clintonium-119 (approved):
Nice work 🎉 

I did comment on a deviation in font weight from the Figma design for a couple elements, so I'd suggest either implementing them, or clearing the difference with Design if you haven't already before merging.

PR #950, opened 2021-10-06
[WEB-1159] Only show resend invite button for non-declined invites

clintonium-119:
Addresses some QA feedback, where the action to resend declined invites was present in the UI, but disallowed (by design) on the backend.


krystophv (approved):
LGTM 🚢

PR #965, opened 2021-11-18
[WEB-1301] Fix for missing profile page custodial permissions

clintonium-119:
See [WEB-1301] for details.

I also noticed a couple of small issues that I fixed at the same time:

1. Since we no longer show the private workspace option on the 'Workspaces' page, it no longer makes sense to redirect there when a clinician is a member of only 1 clinic - even if they have an active private workspace.
2. There was a bug where the workspace switcher was sometimes not showing the private workspace option when it should.  This is now fixed

[WEB-1301]: https://tidepool.atlassian.net/browse/WEB-1301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/clinic/WorkspaceSwitcher.js
+++ b/app/components/clinic/WorkspaceSwitcher.js
@@ -72,7 +72,7 @@ export const WorkspaceSwitcher = props => {
 
       setMenuOptions(options);
     }
-  }, [clinics]);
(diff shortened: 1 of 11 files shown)


krystophv (approved):
LGTM 🚢

PR #984, opened 2021-12-22
[WEB-1204, WEB-1182] Stat updates

clintonium-119:
See [WEB-1204], [WEB-1182].  Adds `Avg Daily Insulin Ratio` stat to trends view, and sets any of the vertically-tall stats to collapsible.  Stat collapse state is persisted in local storage on a per-view basis.

Related PR: https://github.com/tidepool-org/viz/pull/285

[WEB-1204]: https://tidepool.atlassian.net/browse/WEB-1204?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WEB-1182]: https://tidepool.atlassian.net/browse/WEB-1182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/unit/components/chart/deviceSelection.test.js
+++ b/test/unit/components/chart/deviceSelection.test.js
@@ -169,7 +169,6 @@ describe('DeviceSelection', () => {
 
   it('should track a metric when device selection accordion is collapsed or expanded', () => {
     const collapseIcon = wrapper.find('.MuiExpansionPanelSummary-expandIcon').hostNodes();
-    console.log('collapseIcon.debug()', collapseIcon.debug());
     expect(collapseIcon).to.have.lengthOf(1);
 
     sinon.assert.callCount(props.trackMetric, 0);
--- a/test/unit/components/chart/deviceSelection.test.js
+++ b/test/unit/components/chart/deviceSelection.test.js
@@ -17,6 +17,7 @@ const expect = chai.expect;
 
 describe('DeviceSelection', () => {
   const props = {
+    chartType: 'basics',
     devices: [
       { id: 'device1' },
       { id: 'device2' }
@@ -26,6 +27,7 @@ describe('DeviceSelection', () => {
     },
     updateChartPrefs: sinon.stub(),
     removeGeneratedPDFS: sinon.stub(),
+    trackMetric: sinon.stub(),
   };
 
   let wrapper;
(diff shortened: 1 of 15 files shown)


test/unit/components/chart/deviceSelection.test.js
```diff
@@ -160,4 +166,17 @@ describe('DeviceSelection', () => {
     );
     sinon.assert.callCount(props.removeGeneratedPDFS, 4);
   });
+
+  it('should track a metric when device selection accordion is collapsed or expanded', () => {
+    const collapseIcon = wrapper.find('.MuiExpansionPanelSummary-expandIcon').hostNodes();
+    console.log('collapseIcon.debug()', collapseIcon.debug());
```
krystophv, line 172:
```suggestion
```

test/unit/components/chart/stats.test.js
```diff
@@ -106,5 +107,106 @@ describe('Stats', () => {
...
+              // collapsed state of stats
+              averageDailyDose: true,
+              carbs: true,
+            },
+          },
+          sinon.stub()
+        ]));
+
+        wrapper.setProps();
+
```
krystophv, line 171: It seems per the comment above that we should verify that `carbStat` is still opened and title hasn't changed even if the localStorage state reflects a collapsed state.

krystophv (approved):
Code LGTM 👍🏼 

Would suggest adding a couple assertions for one test for completeness.

test/unit/components/chart/stats.test.js
clintonium-119: I've added a couple more test assertions to cover this.  Thanks 👍

krystophv: I notice that dailyDoseStat is the return from a function containing a find. Will the `carbStat` reference be properly updated in the new assertions, or would it also need an arrow function to ensure that we're checking the updated value?

clintonium-119: Let me check.

clintonium-119: Yes, good catch.  I've updated that one to a function as well.

PR #998, opened 2022-01-28
[WEB-1440] Route redirect hotfixes

clintonium-119:
See [WEB-1440] for details.

[WEB-1440]: https://tidepool.atlassian.net/browse/WEB-1440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/core/personutils.js
+++ b/app/core/personutils.js
@@ -46,7 +46,8 @@ personUtils.isPatient = (person) => {
 };
 
 personUtils.isClinicianAccount = (user) => {
-  return (_.indexOf(_.get(user, 'roles', []), 'clinic') !== -1 || user?.isClinicMember);
+  const userRoles = _.get(user, 'roles', []);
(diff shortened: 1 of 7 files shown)


krystophv (changes requested):
Largely LGTM, just one dropped/ignored error that it may be best to return early on. If there's justification for ignoring it, feel free to just comment as such and re-request review 👍🏼

clintonium-119:
I've updated the Auth routes gate to return early for user and clinics fetch errors.  

Also, I noticed an issue where I needed to ensure that when we submit the confirmation modal without having triggered a previous submission via the clinician profile or clinic details forms, the `submitting` state was set to true so that the forced logout would occur (otherwise it would navigate on to the workspaces page before the logout timeout resolves).

krystophv (approved):
LGTM 🚢

PR #1004, opened 2022-02-02
[WEB-1458] Check for clinic invites on initial browser load for authenticated routes

clintonium-119:
See [WEB-1458]

[WEB-1458]: https://tidepool.atlassian.net/browse/WEB-1458?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/routes.js
+++ b/app/routes.js
@@ -1,5 +1,6 @@
 import _ from 'lodash';
 import React from 'react';
+import async from 'async';
 import { Route, Switch, Redirect } from 'react-router-dom';
 import { push } from 'connected-react-router';
 
@@ -60,6 +61,14 @@ export const requireChrome = (next, ...args) => (dispatch) => {
 export const requireAuth = (api, cb = _.noop) => (dispatch, getState) => {
   const { blip: state, router: routerState } = getState();
 
+  const routes = {
+    patients: '/patients',
+    workspaces: '/workspaces',
+    clinicDetails: '/clinic-details',
+    clinicianDetails: '/clinician-details',
+    clinicWorkspace: '/clinic-workspace',
+  };
+
   if (!api.user.isAuthenticated()) {
     dispatch(push('/login'));
   } else {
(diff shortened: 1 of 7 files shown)


app/routes.js
```diff
@@ -112,14 +130,19 @@ export const requireAuth = (api, cb = _.noop) => (dispatch, getState) => {
             // Navigate to the appropriate page for a clinician user or team member who needs to
             // complete the setup process
             if (firstEmptyOrUnmigratedClinic) dispatch(actions.sync.selectClinic(firstEmptyOrUnmigratedClinic.clinic.id));
-            dispatch(push(state.clinicFlowActive || state.selectedClinicId ? '/clinic-details' : '/clinician-details'));
+            dispatch(push(state.clinicFlowActive || state.selectedClinicId ? routes.clinicDetails : routes.clinicianDetails));
+          } else if (values.invites?.length) {
```
krystophv, line 134: I have concerns that this redirect may trigger more often than we want. Though gated by the `fetchingClinicsForClinician` working state, this has the _potential_ to redirect on every authenticated route. If the user has a pending invite that they don't wish to immediately address (neither accepting nor denying it), then they will be redirected to the `clinicDetails` or `workspaces` page each time the working state gets reset. (This happens when switching patients (`DATA_WORKER_REMOVE_DATA_SUCCESS`) and any other time that the data worker is cleared due to that action resetting to initial working state). Is this the desired behavior (i.e. do we want to be that persistent/aggressive about getting the user to take action on the invite)?

clintonium-119: I'm thinking this _should_ be fine.  
I don't think the `blip.working` state is completely reset by that action.  I believe only `blip.working.queryingData` is set to the default working state when `DATA_WORKER_REMOVE_DATA_SUCCESS` fires ([here in working.js](https://github.com/tidepool-org/blip/blob/716da5575592e454b068f31ed01843a7d0f75af1/app/redux/reducers/working.js#L152)).
I could also gate it on `fetchingClinicianInvites` working state.  I thought about it, but then it seemed redundant to me.
Your thoughts?

clintonium-119: Great catch, Chris.  I've addressed this.  I also realized that we also needed to add the "Manage Workspaces" nav menu item if pending clinic invites exist for the user.

krystophv (approved):
LGTM 🎸

PR #1005, opened 2022-02-08
[WEB-1407] add research organization

krystophv:
[WEB-1407] adds research organization option for clinic types


clintonium-119 (approved):
LGTM 💯

PR #1016, opened 2022-02-16
[WEB-1449] Consolidate clinic team member name fields, and pre-fill fields when appropriate

clintonium-119:
See [WEB-1449]


krystophv (approved):
LGTM 🚢

PR #1031, opened 2022-03-15
[WEB-1372] keycloak

krystophv:
For [WEB-1372], adds basic integration with keycloak behind an environment variable. 

See also: https://github.com/tidepool-org/platform-client/pull/135, https://github.com/tidepool-org/tools/pull/92

[WEB-1372]: https://tidepool.atlassian.net/browse/WEB-1372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


webpack.config.js
```diff
@@ -26,6 +26,9 @@ const featureFlags = _.get(optional('./config/local'), 'featureFlags', {
   pendoEnabled: process.env.PENDO_ENABLED || true,
   clinicsEnabled: process.env.CLINICS_ENABLED || false,
 });
+const keycloakUrl = _.get(optional('./config/local'), 'keycloakUrl', process.env.KEYCLOAK_URL || null);
```
clintonium-119, line 29: Perhaps we should update `config/local.example.js` to include the `keycloakUrl, keycloakRealm, keycloakClientId` exports with default (falsey) values.

config.webpack.js
```diff
@@ -26,6 +26,9 @@ const defineEnvPlugin = new webpack.DefinePlugin({
   __RX_ENABLED__: JSON.stringify(process.env.RX_ENABLED || false),
   __PENDO_ENABLED__: JSON.stringify(process.env.PENDO_ENABLED || true),
   __CLINICS_ENABLED__: JSON.stringify(process.env.CLINICS_ENABLED || false),
+  __KEYCLOAK_URL__: JSON.stringify(process.env.KEYCLOAK_URL || null),
```
clintonium-119, line 29: We will need to update the `Dockerfile` to accept these as build arguments in order to run keycloak on remote docker builds.

As well, when the time comes to test these on the remotes, we'll need to update the `artifact.sh` script in the `tools` repo to pass them in from Travis.  Should that be done now?

clintonium-119 (changes requested):
Great work @krystophv!  

I just have a couple small suggestions, and we'll need to update the docker build process to pass the keycloak env variables in.  Details comments.

app/core/api.js
```diff
@@ -85,22 +85,40 @@ api.user.login = function(user, options, cb) {
...
+};
 
+api.user.saveSession = function (userId, token, options, cb) {
+  api.log('saveSession');
+  options = options || {};
+  if (typeof options === 'function') {
+    cb = options;
+    options = {};
+  }
+  tidepool.saveSession(userId, token, options, function (err, session) {
```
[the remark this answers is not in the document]
krystophv: Indeed it didn't. Just following along with the general pattern in `blip` got me here. This should be resolved now in the `platform-client` PR.

webpack.config.js
krystophv: Good call, adding these in.

config.webpack.js
krystophv: `Dockerfile` not being my forte, I'll rely on you to let me know if I've updated it properly. Also added a PR for the `tools` repo at: https://github.com/tidepool-org/tools/pull/92 and have tagged you for review on it (`bash` script also not being something really in my wheelhouse).

clintonium-119 (approved):
Great work on all this.  All looks good 👍

clintonium-119 (commented):
Great work, @krystophv :ok_hand:  All looks good :)
 
Prior to approval, though, I had one question/comment about whether or not the UPLOAD_API change was intended to be included here.

config.app.js
```diff
@@ -51,7 +48,7 @@ function integerFromText(value, defaultValue) {
 
 const config = {
   VERSION: __VERSION__,
-  UPLOAD_API: __UPLOAD_API__ || 'https://tidepool.org/uploader',
+  UPLOAD_API: __UPLOAD_API__ || `${window.location.protocol}//${window.location.host}`,
```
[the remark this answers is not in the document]
krystophv: It is. It turns out that this URI is sort of a relic of the past, but it gets passed in to `platform-client`'s config and is used for the `/info` endpoint (which now provides the keycloak config per environment). Up till needing this call, this URI was completely unused in `blip`, but had already been updated in `uploader` - this basically brings `blip` up to parity. This isn't supposed to be a link to uploader, but the API URI base at which uploads can happen - and for some while now the uploads happen at the same API URI as everything else in our apps.

clintonium-119 (approved):
Nice work 👏 
Approved!

app/pages/login/login.js
```diff
@@ -78,19 +82,36 @@ export let Login = translate()(class extends React.Component {
...
+      !loggingIn &&
+      !this.props.isAuthenticated
+    ) {
+      keycloak.login({ loginHint: this.props.seedEmail });
+    }
+
+    // forward to keycloak login when available
+    if(keycloakConfig.initialized && !loggingIn && !this.props.isAuthenticated){
+      keycloak.login();
+      return (<></>);
```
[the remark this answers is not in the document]
krystophv: There's not a particular reason. The code at one point was returning other JSX and the solution collapsed on returning nothing, so I just made it an empty fragment. Functionally it renders the same as returning `null`, but technically creates an object to track rendering nothing, so I can switch it out to avoid confusion.

PR #1034, opened 2022-03-21
[WEB-1522] Clinic route restriction fixes

clintonium-119:
Hotfix for [WEB-1522]
Primarily addresses users potentially being able to skip the route restriction checks on clinic UI routes if one of the backend calls made prior to the checks returns an error.

I also fixed a small bug where the nav menu wasn't updating to the default options if the user navigated away from the clinic profile page, and cleaned up the error handling on login failures a bit too, including dispatching the logout action if the backend calls fail.

[WEB-1522]: https://tidepool.atlassian.net/browse/WEB-1522?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/redux/actions/async.js
+++ b/app/redux/actions/async.js
@@ -239,7 +239,7 @@ export function login(api, credentials, options, postLoginAction) {
             const userHasClinicProfile = !!_.get(user, ['profile', 'clinic'], false);
             const isClinicianAccount = personUtils.isClinicianAccount(user);
 
-            if (config.CLINICS_ENABLED && isClinicianAccount) {
+            if (config.CLINICS_ENABLED) {
               // Fetch clinic-clinician relationships and pending clinic invites, and only proceed
               // to the clinic workflow if a relationship with a clinic object or an invite exists.
               const fetchers = {
--- a/app/redux/actions/async.js
+++ b/app/redux/actions/async.js
@@ -237,8 +237,9 @@ export function login(api, credentials, options, postLoginAction) {
             handleLoginFailure(ErrorMessages.ERR_FETCHING_USER, err);
           } else {
             const userHasClinicProfile = !!_.get(user, ['profile', 'clinic'], false);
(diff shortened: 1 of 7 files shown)


app/redux/actions/async.js
```diff
@@ -237,8 +237,9 @@ export function login(api, credentials, options, postLoginAction) {
             handleLoginFailure(ErrorMessages.ERR_FETCHING_USER, err);
           } else {
             const userHasClinicProfile = !!_.get(user, ['profile', 'clinic'], false);
+            const isClinicianAccount = personUtils.isClinicianAccount(user);
 
-            if (config.CLINICS_ENABLED) {
+            if (config.CLINICS_ENABLED && isClinicianAccount) {
```
krystophv, line 242: If we gate this login redirection path on `isClinicianAccount`, then won't that prevent non-clinic (classic or new) users who receive a clinician invite from a clinic from getting to the clinician workflow?

krystophv (changes requested):
Overall looks good, just one concern about new clinician invitees not getting routed properly.

app/redux/actions/async.js
clintonium-119: Hmm. I just tested it with an invite using a new email address and it did in fact land the new clinician on the clinicDetails page as desired. Now, this is because I chose a 'Clinician' Account from the user sign up flow, and thus the new user had the clinic role.

Now, if I had chosen to create a standard patient account, or had invited via email an existing standard patient account, you're correct that they would not see the invitation. Should we allow invites to non-clinician accounts?

Either way, I suppose I'm not gating the corresponding route checks on refresh in routes.js, so perhaps I should just remove this gate here. My intention was to not bother with additional server calls if it wasn't a clinician user, but it's rather insignificant and harmless to leave them in. Any other thoughts on this?

krystophv (approved):
LGTM 🚢

PR #1037, opened 2022-03-21
[WEB-1078] daily view date picker

clintonium-119:
See [WEB-1078]

While I had initially wrote this with most of the changes done to the `Daily` component, having it reacting post-mount to updates to the `initialDatetimeLocation` prop, this proved to be very troublesome, and hard to reason about from a code perspective.  As I was preparing the PR for review, I went about commenting on some of the more 'odd' bits of code required to make this work, and felt I needed to rework my approach this one, as it was apparent that this method was too flaky, and very complicated with the way the current tideline view handles changes to this prop.

I ended up with what I feel is a much cleaner solution that handles the date changes in the parent `PatientData` component and triggers a remount when the initial datetime needs to be updated via updating the `key` prop (updating a key forces a remount).  The remount was the surest way to avoiding edge case side-effects that I'd been chasing down with any other approach I tried.

I've also made an update to how the `queryDataCount` prop was tracked, as there was a bug that was causing the Daily view chart to update a few extra times after data updates.  It is now incremented in the data worker itself, rather than as a piece of state tracked by the `PatientData` wrapper.  This also fixes some general render jitters when initially loading or locating the daily chart tideline components.

Related PR: https://github.com/tidepool-org/viz/pull/313

Other Notes:
- I noticed a couple of minor issues with the `ChartDateRange` modal component: the latest selectable date should not be beyond the latest diabetes datum time, and the time zone was not being taken into account on the startDate returned from the component.
- This also fixes an existing bug where the setting of the most recent datum time was not being set properly in the `handleSwitchToDaily` method, resulting in a click to the most recent datetime via the chart navigation icons would not always get all the way to end of the scrollable area.

[WEB-1078]: https://tidepool.atlassian.net/browse/WEB-1078?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


krystophv (approved):
Great work on this! I think this is a solid strategy to deal with the potential edge case scenarios and keeps the lower components from needing to keep up with too much derived state. LGTM 🚢

PR #1040, opened 2022-03-27
Add french locale

gcharest:
I agree to the terms of Tidepool Project’s Volunteer/Contributor License Agreement v1.1 as it exists at http://tidepool-org.github.io/files/TidepoolVCLA-1.1.pdf on 2022-03-27.

This PR is to propose the addition of the French locale in Tidepool Web.

The changes include the modification of the following components where strings were hardcoded:
- app/components/datasources/datasources.js
- app/components/footerlinks/footerlinks.js

> Note: Other components may have hardcoded strings but through testing locally the web application, they were the only ones I could identify.

The changes also include the addition of many strings initially absent from the translation.json files, using i18next-parser, then their translation in French, which explains why the EN and ES json files were also modified:
- locales/en/translation.json
- locales/es/translation.json
- locales/fr/translation.json


clintonium-119 (approved):
Approved 👍 
Amazing work, @gcharest !
We very much appreciate all that you've put into this :)

I made a couple of small changes:

1. Switched to using the `<Trans>` component to handle the translation of content with nested markup in `datasources.js`
2. Updated the translation files with the `update-translations` NPM script.  Otherwise, translated strings from our `tideline` and `@tidepool/viz` repos could get stripped.

gcharest:
That's awesome! Thanks a lot!

I'll double check the script you mentioned as I must have missed it.

PR #1043, opened 2022-03-30
Language config fix

clintonium-119:
This fixes the `update-translations` yarn script to properly parse the translatable strings in the viz source code.
Accordingly, the translations files are also updated here to include the numerous missing strings.

One other fix is to only change the locale based on a user's stored preferences when the `I18N_ENABLED` flag is set


krystophv (approved):
LGTM 👍🏼

PR #1051, opened 2022-04-14
Translate missing French strings

gcharest:
This is a follow up PR to translate the missing French strings found by @clintonium-119.


gcharest:
Ref #1040

The Viz strings were not initially captured by the `update-translations` script, which was since fixed.

clintonium-119 (approved):
Thanks so much for this @gcharest !
Merging :)

PR #1056, opened 2022-05-09
[WEB-1577] enforce date format for new patient birthday

krystophv:
To remedy [WEB-1577], this uses `moment`'s strict parsing for the birthday field for new patients to avoid the default behavior of allowing "overflow" months/days which lets invalid birthdays (generally mistaking MM/DD/YYYY for DD/MM/YYYY) through.

[WEB-1577]: https://tidepool.atlassian.net/browse/WEB-1577?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clintonium-119 (approved):
Looks good 🚢 
Curiosity question: I see that you've set a hotfix semver (`1.53.1-...`).  Is this going out as a hotfix, in which case we should change the base to master?

krystophv:
I hadn't really thought about it being a hotfix vs just a bugfix. I suppose it could go through as a hotfix since the priority of the ticket is pretty high and it's a relatively small fix in scope.

PR #1071, opened 2022-06-13
Debug cause of preferred bg units not working in production

clintonium-119:
Need to add some debugging log statements to production to investigate further.


krystophv (approved):
LGTM 👍🏼

PR #1072, opened 2022-06-14
[WEB-1660] Fix for clinic preferred BG units not being applied correctly in certain circumstances.

clintonium-119:
See [WEB-1660]

The main issue was that the code block that was intended to perform the override had some incorrect conditional logic for when to apply the overrides.  

Specifically, the override units from the clinic were stored in the following formats:  `mmol/L | mg/dL` 
The conditional code was validating that it matched one of `[mmoll, mgdl]`

I also updated it to allow the query param `units` override to take precedence over the clinic's preferred bg units so that it's still a functional option within a clinic context.

[WEB-1660]: https://tidepool.atlassian.net/browse/WEB-1660?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/core/utils.js
+++ b/app/core/utils.js
@@ -341,29 +341,36 @@ utils.getTimePrefsForDataProcessing = (latestUpload, queryParams) => {
   return timePrefsForTideline;
 };
 
-utils.getBGPrefsForDataProcessing = (patientSettings, { units: overrideUnits, source }) => {
-  var bgUnits = _.get(patientSettings, 'units.bg', overrideUnits?.replace('/', '').toLowerCase() === 'mmoll' ? MMOLL_UNITS : MGDL_UNITS);
+utils.getBGPrefsForDataProcessing = (patientSettings, { units: overrideUnits, source: overrideSource }) => {
+  // Allow overriding stored BG Unit preferences via query param or preferred clinic BG units
+  // If no override is specified, use patient settings units if availiable, otherwise 'mg/dL'
+  // const patientSettingsBgUnits = _.get(patientSettings, 'units.bg',  MGDL_UNITS);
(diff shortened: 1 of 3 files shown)


app/core/utils.js
```diff
@@ -341,29 +341,36 @@ utils.getTimePrefsForDataProcessing = (latestUpload, queryParams) => {
   return timePrefsForTideline;
 };
 
-utils.getBGPrefsForDataProcessing = (patientSettings, { units: overrideUnits, source }) => {
-  var bgUnits = _.get(patientSettings, 'units.bg', overrideUnits?.replace('/', '').toLowerCase() === 'mmoll' ? MMOLL_UNITS : MGDL_UNITS);
+utils.getBGPrefsForDataProcessing = (patientSettings, { units: overrideUnits, source: overrideSource }) => {
+  // Allow overriding stored BG Unit preferences via query param or preferred clinic BG units
+  // If no override is specified, use patient settings units if availiable, otherwise 'mg/dL'
+  // const patientSettingsBgUnits = _.get(patientSettings, 'units.bg',  MGDL_UNITS);
```
krystophv, line 347:
```suggestion
```

krystophv (approved):
One dead code line hanging around, but otherwise LGTM 🚀

clintonium-119:
@krystophv Just re-requested review.  I made a small change, which is to default the patient settings units to mg/dL if not present, which is the default - i.e. the units are only stored on the patient settings object once they've been modified in the patient profile form.

It was technically still working in those cases, but for a different reason (same reason we didn't notice the error that would crop up in production with the original release).  The change should not affect the fix that needed to be applied for the production bug.

PR #1076, opened 2022-06-15
[WEB-1672] Fix issues with patient data loading

clintonium-119:
See [WEB-1672]
Fixes a regression introduced in `v1.53.0` where switching between the profile page and the patient view causes the patient view to not load.  This was due to moving the queryDataCount metadata to the Data worker, and not resetting it to zero when we drop the redux state data, but preserve the cached data in the worker, as happens when navigating to the profile page from a data view.

With the queryDataCount set to zero, it now properly fetches the initial data (from the worker) and renders the view.

As well, I noticed that some old top-level code in `app.js` that was used to ensure we had `selectedClinicId` state set was affecting the page refreshes when viewing clinic custodial account data, setting the `selectedClinicId` arbitrarily to the first available clinic.  This was causing that clinic's preferred bg units to prevail over the custodial account units.  This code should have been deleted when we transitioned to storing the `selectedClinicId` state in localstorage for persistence. 

[WEB-1672]: https://tidepool.atlassian.net/browse/WEB-1672?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/pages/app/app.js
+++ b/app/pages/app/app.js
@@ -175,12 +175,6 @@ export class AppComponent extends React.Component {
       this.doFetching(nextProps);
     }
 
-    if (!this.props.clinicFlowActive && nextProps.clinicFlowActive && !selectedClinicId && _.keys(clinics).length) {
-      // We keep the selectedClinicId state at it's default 'null' if the app loads on the legacy
-      // patients page. Otherwise, we select the first available clinic.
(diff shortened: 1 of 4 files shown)


krystophv (approved):
I suppose cache invalidation is one of the hard parts of programming :)

LGTM 🎸

PR #1078, opened 2022-06-16
[WEB-1682] Fix for recursive patient data fetching immeditately after login

clintonium-119:
See [WEB-1682]

Basically, when the (broken) refresh functionality was fixed in `v1.52.0`, it was set to detect changes to react to the removal of patient data from the data worker.  Because the completed state of that action was never being set back to null, it was causing, in some instances, [the data fetch to fire over and over again](https://github.com/tidepool-org/blip/blob/24ee0e8a866545f6e4bbb23f8cab8e984d18fd6c/app/pages/patientdata/patientdata.js#L1457). 

[WEB-1682]: https://tidepool.atlassian.net/browse/WEB-1682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


krystophv (approved):
LGTM 🚢

PR #1100, opened 2022-08-11
[WEB-1644] summary stat percentage rounding

krystophv:
For [WEB-1644], applies a number of rounding rules to prevent rounding display values into filter boundaries.

[WEB-1644]: https://tidepool.atlassian.net/browse/WEB-1644?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/clinic/BgRangeSummary.js
+++ b/app/components/clinic/BgRangeSummary.js
@@ -43,7 +43,7 @@ export const formatValue = (inputValue, key) => {
         precision = 2;
         percentage = roundUp(percentage, precision);
       }
-      if (percentage > 0.5 && percentage < 1) {
+      if (percentage >= 0.5 && percentage < 1) {
         precision = 1;
         percentage = roundDown(percentage, precision);
       }
--- a/app/components/clinic/BgRangeSummary.js
+++ b/app/components/clinic/BgRangeSummary.js
@@ -17,22 +17,71 @@ import { space, shadows } from '../../themes/baseTheme';
 import { utils as vizUtils } from '@tidepool/viz';
 const { reshapeBgClassesToBgBounds, generateBgRangeLabels } = vizUtils.bg;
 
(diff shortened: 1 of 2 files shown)


app/components/clinic/BgRangeSummary.js
```diff
@@ -17,23 +17,72 @@ import { space, shadows } from '../../themes/baseTheme';
...
+  }
 
-    return format(`.${precision}f`)(percentage);
+  switch (key) {
+    case 'veryLow':
+      if (percentage > 0 && percentage < 0.5) {
+        precision = 2;
+        percentage = roundUp(percentage, precision);
+      }
+      if (percentage > 0.5 && percentage < 1) {
```
clintonium-119, line 46: I realize what you have matches the done criteria of the ticket, but, would this be better as:
```suggestion
      if (percentage >= 0.5 && percentage < 1) {
```
If I'm tracking this correctly, a `percentage` value of _exactly_ `0.5` would end up with a precision of `0`, and would be formatted as `0` (I don't think `d3-format` rounds for the`f` notation)

If this is correct, we should also update the done criteria of the ticket to suit.

clintonium-119 (changes requested):

app/components/clinic/BgRangeSummary.js
krystophv: Ah, yeah - good catch. I'll make the adjustment and add a unit test for that case.

krystophv: @clintonium-119 added in the unit test, adjusted the boundary condition and updated the done criteria 👍🏼

clintonium-119 (approved):
Looks great :fire: 
Approved.

PR #1104, opened 2022-08-29
[WEB-1701] scroll to top on page, performance pass, eslint react-hooks/recommended

krystophv:
As part of the latest design tweaks for [WEB-1701] (scrolling to the top of the patient interface on pagination) it was discovered that there were performance issues causing loading state hangs in the interface. As a start, I grabbed the eslint react-hooks/recommended plugin and went about trying to satisfy all of its errors and warnings for the ClinicPatients page. This lead to a lot of dependency adding, cascading into needing to move hook dependencies into hooks themselves and use-before-define errors to solve by rearranging significant portions of code. A couple sections were moved into sub-components in order to be able to move their prop processing into their own hooks as the renders for them were already in a hook (and, of course, one does not simply nest hooks).

In trying to keep the modifications scoped, I only edited files elsewhere in the codebase to resolve eslint errors (as opposed to warnings) or things that caused other errors (adding setToast from ToastProvider as a hook dependency resulted in an infinite loop when running the test suite which was _particularly_ difficult to debug).

[WEB-1701]: https://tidepool.atlassian.net/browse/WEB-1701?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -14,6 +14,9 @@
     "__INVITE_KEY__": false,
     "__LATEST_TERMS__": false
   },
+  "extends": [
+    "plugin:react-hooks/recommended"
+  ],
   "rules": {
     "quotes": [2, "single", {
       "avoidEscape": true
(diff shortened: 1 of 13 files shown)


clintonium-119 (approved):
Thanks for digging deep on this.  Great work :fire:  Approved.

PR #1118, opened 2022-10-07
[WEB-1838] force remount on data fetch after refresh

krystophv:
For [WEB-1838]. If the `chartKey` happens to match up with the incoming query metadata on a query that tries to force a remount, the result of the update is nil. This can happen because the chartKey only gets updated on forced remount attempts and can recycle through values as the query count gets reset to 0 on use of the Refresh button. If we force a remount (and thus update the chartKey state value) on "initial" queries when we've already got a value for `state.chartKey`, we can keep the values better in sync post-Refresh. 

[WEB-1838]: https://tidepool.atlassian.net/browse/WEB-1838?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/pages/patientdata/patientdata.js
+++ b/app/pages/patientdata/patientdata.js
@@ -1542,6 +1542,7 @@ export const PatientDataClass = createReactClass({
           excludedDevices: undefined,
           timePrefs,
           bgPrefs,
+          forceRemountAfterQuery: this.state.chartKey > 0
         });
       }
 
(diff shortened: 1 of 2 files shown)


clintonium-119 (approved):
LGTM :1st_place_medal: 
Great job debugging this.

PR #1123, opened 2022-10-14
Web 1807 modify browser warning page

ginnyyadav:
Modified browser warning page. This page shows whenever a user tries to access their own data as well as when a clinician tries to access a user's data. This should show up on mobile devices as well as unsupported browsers. I've only modified this to catch major browsers. So many have the word "chrome" in their UA string, so I don't think we can target them all.  I've tried to leave most of the original page intact. 

Browser 
image
Mobile 
image


clintonium-119 (approved):
Great work on this :fire: 
Approved - though I have a suggestion to add a missing semicolon prior to merging in the review comments.

PR #1134, opened 2022-10-20
[WEB-1939] Update regex for tag name validation

clintonium-119:
See [WEB-1939]

[WEB-1939]: https://tidepool.atlassian.net/browse/WEB-1939?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/core/clinicUtils.js
+++ b/app/core/clinicUtils.js
@@ -139,7 +139,7 @@ export const clinicSchema = yup.object().shape({
 export const clinicPatientTagSchema = yup.object().shape({
   name: yup.string()
     .max(20, t('Tag name max length is ${max} characters'))
-    .matches(/^[\p{L}\p{N}\s_+><-]*$/u, t('Allowed special characters: - _ + > <'))
+    .matches(/^[\p{L}\p{N}_+><-]{1}[\p{L}\p{N}\s_+><-]*$/u, t('Allowed special characters: - _ + > <'))
 })
 
 export const patientSchema = yup.object().shape({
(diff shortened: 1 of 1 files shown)


app/core/clinicUtils.js
```diff
@@ -139,7 +139,7 @@ export const clinicSchema = yup.object().shape({
 export const clinicPatientTagSchema = yup.object().shape({
   name: yup.string()
     .max(20, t('Tag name max length is ${max} characters'))
-    .matches(/^[\x21-\x7E\s]*$/, t('Only standard characters are allowed'))
+    .matches(/^[\p{L}\p{N}\s_+><-]*$/u, t('Allowed special characters: - _ + > <'))
```
krystophv, line 142: I think this may need to be adjusted to prevent leading spaces.

clintonium-119: I actually have our fields automatically trimming leading and trailing whitespace prior to validation in our default usage here:
https://github.com/tidepool-org/blip/commit/0316cee886bc838ae22338d42df907d02550efb0

I did get more explicit on the backend regex validation, ensuring that there were no leading spaces (even though I do a similar leading/trailing trim before writing to the DB).  Agree that I should update it either here or there (or both) for consistency.

krystophv (approved):
LGTM 👍🏼

PR #1148, opened 2022-11-17
Make sure table border can be seen in mozilla

ginnyyadav:
Went down this rabbit hole  and seems to be an issue specifically with Mozilla Firefox. https://bugzilla.mozilla.org/show_bug.cgi?id=688556 
 
background-clip: padding-box seems to fix the issue.

### DIFF
--- a/app/components/elements/Table.js
+++ b/app/components/elements/Table.js
@@ -180,7 +180,6 @@ export const Table = React.memo(props => {
                   key={`${id}-header-${col.field?.replace(/\./g, '-')}`}
                   align={col.align || (index === 0 ? 'left' : 'right')}
                   sortDirection={orderBy === colSortBy ? order : false}
-
                 >
                   <Box
                     className="table-header-inner-cell"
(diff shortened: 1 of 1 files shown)


app/components/elements/Table.js
```diff
@@ -72,6 +72,7 @@ const StyledTable = styled(Base)`
     color: inherit;
     font-size: inherit;
     font-family: inherit;
+    background-clip: padding-box
```
clintonium-119, line 75:
```suggestion
    background-clip: padding-box;
```
Should always have a semicolon after style declarations.  No harm here, but any subsequent styles added would be ignored.

app/components/elements/Table.js
```diff
@@ -179,6 +180,7 @@ export const Table = React.memo(props => {
                   key={`${id}-header-${col.field?.replace(/\./g, '-')}`}
                   align={col.align || (index === 0 ? 'left' : 'right')}
                   sortDirection={orderBy === colSortBy ? order : false}
+
```
clintonium-119, line 183:
```suggestion
```
Suggest cleaning up this extra line that was added

clintonium-119 (changes requested):
Just a couple of tiny change requests.  Just non-breaking nits.

ginnyyadav:
Thanks @clintonium-119 good catches 🙏

clintonium-119 (approved):
LGTM :)

PR #1160, opened 2022-12-13
[WEB-1998] redirect logout through keycloak if available

krystophv:
For [WEB-1998], will send the user to the keycloak logout URL if it's available instead of redirecting to `\`, which can potentially beat out the `keycloak.logout()` call we were previously using, resulting in a not-logged-out state.

[WEB-1998]: https://tidepool.atlassian.net/browse/WEB-1998?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/loginnav/loginnav.js
+++ b/app/components/loginnav/loginnav.js
@@ -6,6 +6,8 @@ import { keycloak } from '../../keycloak';
 import { translate } from 'react-i18next';
 var Link = require('react-router-dom').Link;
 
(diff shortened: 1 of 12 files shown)


clintonium-119 (approved):
Nice work.  LGTM :fire:

PR #1161, opened 2022-12-13
Add local environment to apiHost config options list

clintonium-119:
Local option was never added to the example file.  Now added.


gniezen (approved):
LGTM! :shipit:

PR #1176, opened 2023-02-08
[WEB-2064] Hotfix: Ensure latest pump settings are fetched

clintonium-119:
[WEB-2064] 
Currently, we exclude the latest `pumpSettings`, `upload`, and `food` datums when determining the fetch range for the initial data fetch.

This can lead to a situation, however, where the latest `pumpSettings` are not fetched and added to the data worker, so this fix still uses only the latest diabetes (insulin or bg) datums to determine the start date of the fetch range, but ensures the end date accounts for the non-diabetes datums as well. 

[WEB-2064]: https://tidepool.atlassian.net/browse/WEB-2064?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


krystophv (approved):
LGTM 🔥

PR #1183, opened 2023-02-16
[UPLOAD-996] add uploader URI scheme to frameSrc CSP

krystophv:
for [UPLOAD-996], pretty much what it says on the tin, we need to allow our custom URI scheme on frameSrc for the mechanism that the custom URI launcher library uses for Firefox


clintonium-119 (approved):
Approved :100:

PR #1186, opened 2023-03-02
Add firefox as supported browser

### LINKED ISSUE #418: Only works in Google Chrome
Hi. The 90s called, and they want their "Only works in Netscape" error page back.

What about Windows 10 S where only Microsoft Edge browser is available? What about iOS with Safari only? or Nintendo Wii U or Xbox One. On these platforms you can’t install another browser. I personally prefer Firefox and have no intention of installing Chrome or another browser.

There should at least be a way to say “I understand that you guys don’t bother testing in other browsers, now let me proceed anyway.”

### LINKED ISSUE #1029: Blip requires Chrome, although site works fine in Firefox
If you are going to test browser compatibility, test on availability of required features, not useragent string. All required Web APIs should be checked in utils instead of the browser. Technically it would be possible to run an outdated version of Chrome, although it requires some work, which could render the site useless.

The site works fine in Firefox (only tested in version 97), but requires altering the userAgent string to get past the browser compatibility test.

Are there specific features you require that are not available in other browsers? I understand you might be interested in blocking IE, but Safari is likely to be compatible as well as FF. There are Chromium (Google specific) Web APIs that other browser engines haven't implemented, but unless they are strictly necessary, I don't see why you should be blocking FF and Safari.

Alternatively, you can warn the user that the application might not work correctly in other browsers, and you recommend using Chrome, but allow the user to override anyway. Similar to allowing running unsigned applications on most modern OSes.

nielsvanvelzen:
I recently switched to Firefox as my primary browser and was surprised to see Tidepool blocking browsers without any clear reason. As explained in the mentioned issues, if there are specific functionalities that require Chrome you should use feature testing instead of blocking a complete browser. This is a really dark UX pattern.

This PR adds Firefox to the "supported browsers" list, which obviously is still a hack because Tidepool still blocks browsers for not being Chrome.

Fixes #418 
Fixes #1029

### DIFF
--- a/app/core/utils.js
+++ b/app/core/utils.js
@@ -69,7 +69,7 @@ utils.getIn = (obj, props, notFound) => {
 
 utils.isSupportedBrowser = () => {
   var userAgent = navigator.userAgent.toLowerCase();
-  return (userAgent.indexOf('chrome') > -1 && userAgent.indexOf('opr') === -1 && userAgent.indexOf('mobi') === -1);
+  return (userAgent.indexOf('chrome') > -1 && userAgent.indexOf('opr') === -1 && userAgent.indexOf('mobi') === -1 && userAgent.indexOf('firefox') === -1);
 };
 
 utils.isMobile = () => {


krystophv:
Thanks for this, you can follow along with our efforts to expand browser support here: https://tidepool.atlassian.net/browse/WEB-1806

nielsvanvelzen:
Thanks for the fast response. It's great to see Firefox support is already in progress. The linked issues didn't have any official responses so I wasn't aware of that. It does look like the last changes to the ticket happened back in August, does that mean the project is currently on hold? 

Would it be possible to create some kind of "ignore this warning" button in the meantime as an easy-win for the Firefox users?

krystophv:
The linked issue has many sub-tasks which describe testing and bug reports from cross-browser verification. Firefox support is the next priority and, to my knowledge, the testing and bug fixes for Firefox are now complete. All that remains is the task to adjust our supported browser list and page. I'll bring this up in the next sprint prioritization for consideration.

We cannot add a warning bypass for browsers that we haven't yet verified and documented functionality for.

nielsvanvelzen:
Thanks for the responses, appreciate the efforts! I'll close the PR so you guys can add the support properly.

vitawasalreadytaken:
Hi, apologies if I'm hijacking an old thread, but this seems to be the one place on GitHub which aggregates several duplicate issues on the topic of browser compatibility. 

For what it's worth, I'd like to express support for this bugfix as well. I deliberately use the word "bugfix" rather than "feature". Google Chrome is a browser that comes with serious privacy issues (just like its parent company, and that's no accident). I am disappointed that I am being forced to use it if I want to use Tidepool. I am a big fan of Tidepool's open-source ethos, and this odd privacy-harming restriction doesn't quite fit it 😕 

I found some quite recent updates on child issues of https://tidepool.atlassian.net/browse/WEB-1806, so it seems there is some progress on this. Perhaps the delivery could be sped up by expanding browser support one by one? 🙏 

(I am aware that I can circumvent this block by modifying the User-Agent string in the meantime.)

Thank you for your work – I hope this doesn't come across as ungrateful. It is amazing how much quality software you put out there for free use!

scl67:
Any reason why this was closed?

aleksandrs-ledovskis:
FWIW, the login and functionality works just fine if forcing User-Agent override via an extension.

For example, am using https://addons.mozilla.org/en-GB/firefox/addon/user-agent-string-switcher/ and "Custom mode" for `"app.tidepool.org"`

```
{
  "app.tidepool.org": "<your choice of Chrome's User-Agent>"
}
```

ryceg:
There's no reason to not simply test for the features which you require- 
```js
if ("hid" in navigator) {
  console.log("WebHID is supported!");
} else {
  console.log("WebHID is NOT supported.");
}
```

Plus, you safeguard against old versions of Chrome that are not compatible. Then when Firefox and Safari eventually add support, you don't have to do anything.

PR #1191, opened 2023-03-14
[WEB-2115] disable leaving if last other admin is only invited

krystophv:
for [WEB-2115] adjusts logic to only consider non-invited clinic admins

[WEB-2115]: https://tidepool.atlassian.net/browse/WEB-2115?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/pages/clinicianedit/clinicianedit.js
+++ b/app/pages/clinicianedit/clinicianedit.js
@@ -10,6 +10,8 @@ import map from 'lodash/map';
 import includes from 'lodash/includes';
 import cloneDeep from 'lodash/cloneDeep';
 import filter from 'lodash/filter';
+import indexOf from 'lodash/indexOf';
+import isUndefined from 'lodash/isUndefined';
 import * as yup from 'yup';
 import { Box, Flex, Text } from 'rebass/styled-components';
 import { components as vizComponents } from '@tidepool/viz';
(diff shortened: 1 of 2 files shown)


app/pages/clinicianedit/clinicianedit.js
```diff
@@ -60,7 +62,13 @@ export const ClinicianEdit = (props) => {
...
   const selectedClinicianRoles = selectedClinician?.roles;
   const clinicianName = selectedClinician?.name;
-  const adminCount = filter(clinicians, { roles: ['CLINIC_ADMIN'] }).length;
+  const activeAdmins = filter(
+    clinicians,
+    (clinician) =>
+      indexOf(clinician.roles, 'CLINIC_ADMIN') !== -1 &&
+      isUndefined(clinician.inviteId)
+  );
+  const adminCount = filter(activeAdmins, { roles: ['CLINIC_ADMIN'] }).length;
```
clintonium-119, line 71: I don't think this needs to be filtered here, since all `activeAdmins` are guaranteed to have the `CLINIC_ADMIN` role.
```suggestion
  const adminCount = activeAdmins.length;
```

clintonium-119 (changes requested):
Looks good. Just one small code simplification suggestion in comments to address.

app/pages/clinicianedit/clinicianedit.js
krystophv: Ah yeah, that logic moved up so this can be simplified. Good call 👍🏼

clintonium-119 (approved):
Approved :fire:

PR #1195, opened 2023-03-22
[WEB-2098] logged out page and 401/403 redirects

krystophv:
for [WEB-2098] creates logged out page which users will be sent to when keycloak token authentication/refresh fails. Additionally triggers a keycloak token refresh whenever any API call returns a 401/403 error to verify session validity.

[WEB-2098]: https://tidepool.atlassian.net/browse/WEB-2098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/keycloak.js
+++ b/app/keycloak.js
@@ -65,6 +65,7 @@ export const onKeycloakEvent = (store) => (event, error) => {
     }
     case 'onAuthRefreshError': {
       store.dispatch(sync.keycloakAuthRefreshError(event, error));
+      store.dispatch(async.loggedOut(api));
       break;
     }
     case 'onTokenExpired': {
@@ -73,6 +74,7 @@ export const onKeycloakEvent = (store) => (event, error) => {
     }
     case 'onAuthLogout': {
       store.dispatch(sync.keycloakAuthLogout(event, error));
+      store.dispatch(async.loggedOut(api));
       break;
     }
     default:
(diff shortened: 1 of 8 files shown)


test/unit/keycloak.test.js
```diff
@@ -264,6 +266,37 @@ describe('keycloak', () => {
       expect(updateKeycloakConfigMock.callCount).to.equal(0);
       keycloak.__ResetDependency__('_keycloakConfig');
     });
+
+    it('should call keycloak.updateToken() when action has 401 or 403 error', () => {
```
tjotala, line 270: Stylistic nit: these might be "cleaner" as two separate test cases.

app/keycloak.js
```diff
@@ -103,6 +105,16 @@ export const keycloakMiddleware = (api) => (storeAPI) => (next) => (action) => {
       break;
     }
     default:
+      if (
+        action?.error?.status === 401 ||
+        action?.error?.originalError?.status === 401 ||
+        action?.error?.status === 403 ||
```
tjotala, line 111: I can see how this makes perfect sense for 401.

More of a philosophical question here: will re-login solve a 403?

I suppose won't hurt to try, but if they didn't have the permission before, it seems unlikely that they would gain it after logging in again - unless someone just granted them the permission.

tjotala (approved):
LGTM - no blockers, just nits

app/keycloak.js
krystophv: There's some amount of historical ambiguity around which status the backend returns in which situation. There have been cases where we should have used one but return the other, so this code treats both the same way.

test/unit/keycloak.test.js
krystophv: Wouldn't hurt to split this out. 👍🏼

PR #1205, opened 2023-04-04
[WEB-2107] user profile email in use message

krystophv:
for [WEB-2107]. also decided to use the `working` state to determine whether or not a save was successful instead of just _always_ optimistically telling the user that it was and disabling the save button while the call is in progress.

[WEB-2107]: https://tidepool.atlassian.net/browse/WEB-2107?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clintonium-119 (approved):
Nice job.  Good call on properly tracking progress.  Approved :fire:

PR #1227, opened 2023-04-24
[WEB-2248] just use host for rollbar environment

krystophv:
for [WEB-2248], Rollbar has the `environment` as part of their URL structure for some pieces of their application, so using a full protocol+host breaks the URLs for those pieces of the application (and they don't urlencode it...) so we can just use the `host` going forward.

[WEB-2248]: https://tidepool.atlassian.net/browse/WEB-2248?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clintonium-119 (approved):
LGTM :+1:

PR #1240, opened 2023-05-31
[WEB-2395] Add GA4 page view and sign up redirect tracking

clintonium-119:
[WEB-2395]

Related PRs:
 - https://github.com/tidepool-org/development/pull/230
 - https://github.com/tidepool-org/tools/pull/97

[WEB-2395]: https://tidepool.atlassian.net/browse/WEB-2395?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/routes.js
+++ b/app/routes.js
@@ -1,6 +1,5 @@
 import _ from 'lodash';
 import React from 'react';
-import ReactGA from 'react-ga4';
 import async from 'async';
 import { Route, Switch, Redirect } from 'react-router-dom';
 import { push } from 'connected-react-router';
@@ -327,21 +326,6 @@ export const getRoutes = (appContext) => {
   return (
     <Route path='/' {...props} render={routeProps => (
       <AppComponent {...routeProps} {...props}>
-        <Route
-          path="/"
-          render={({ location }) => {
-            if (ReactGA.isInitialized) {
-              ReactGA.send({
-                hitType: 'pageview',
-                page: location?.pathname + location?.search + location?.hash,
-                'page_search': location.search,
-                'page_hash': location.hash,
-              });
-            }
-            return null;
-          }}
-        />
-
         <Switch>
           <Route exact path='/' render={routeProps => (<Gate onEnter={boundRequireNoAuth} key={routeProps.match.path}><Login {...routeProps} {...props} /></Gate>)} />
           <Route path='/login' render={routeProps => (<Gate onEnter={boundRequireNoAuth} key={routeProps.match.path}><Login {...routeProps} {...props} /></Gate>)} />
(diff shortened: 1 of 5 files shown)


app/pages/signup/signup.js
```diff
@@ -168,6 +169,14 @@ export let Signup = translate()(class extends React.Component {
...
       }
       keycloak.register(options);
+
+      if (ReactGA.isInitialized) {
+        ReactGA.event({
+          category: 'redirect',
+          action: 'Keycloak Sign Up',
+          label: inviteEmail ? 'Has inviteEmail' : 'No inviteEmail',
+        });
+      }
```
krystophv, line 179: The `keycloak.register()` call above this is ultimately a redirect, so I'm concerned that this will end up being a race condition as to whether or not this event gets sent properly. Given the async nature of _both_ of the calls, it might not help a ton, but could be worth moving this above the `keycloak.register()`

app/routes.js
```diff
@@ -326,6 +327,21 @@ export const getRoutes = (appContext) => {
...
+              ReactGA.send({
+                hitType: 'pageview',
+                page: location?.pathname + location?.search + location?.hash,
+                'page_search': location.search,
+                'page_hash': location.hash,
+              });
+            }
+            return null;
+          }}
+        />
```
krystophv, line 343: While I do like the concept of tracking the page views at the route level, I'm not a huge fan of having a `/` level `Route` that's doing duty as a makeshift middleware and always returns `null`. I think that using the `useLocation` hook in the `AppComponent` might be a cleaner/more idiomatic way to go: https://v5.reactrouter.com/web/api/Hooks/uselocation

krystophv (changes requested):
I've got a couple small concerns. Lemme know if you disagree or if I've missed something.

app/routes.js
clintonium-119: I could also use the `withRouter` HOC, combined with `history.listen` directly in app.js.  Something like: https://stackoverflow.com/a/45373907

clintonium-119: I think I may have played around with this approach as well (I think I had access to the history within the location prop, IIRC) and had it working with the component lifecycle methods, but had opted for the vanity route method (had found a couple examples of doing that online, but I get the distaste for a 'fake' route, though it seemed like having the logic within the routing section was a nice location)

clintonium-119:
@krystophv I've gone and moved the GA init and pageview logic to a wrapper component.  I think it's better than the vanity route we were going to settle with.  Requesting re-review

krystophv (approved):
LGTM 💯

PR #1251, opened 2023-06-09
[WEB-2431] handle selecting clinic a patient is a member of in all cases

krystophv:
for [WEB-2431], expands selecting a clinic to handle cases where there _is_ a selected clinic, but it's not a clinic that the current patient-in-view is a member of instead of only handling the case where there is no selected clinic

[WEB-2431]: https://tidepool.atlassian.net/browse/WEB-2431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clintonium-119 (approved):
LGTM :+1:

PR #1275, opened 2023-08-15
[WEB-2498] Handle TIDE unhappy paths

clintonium-119:
[WEB-2498]

Adds empty results handling for individual sections, or the view as a whole, as well as:
- Disables dashboard CTA if clinic has not created any patient tags
- validates stored config, and opens automatically when CTA is selected, or tide dashboard is loaded, if invalid
- clears stored results on dashboard unmount

[WEB-2498]: https://tidepool.atlassian.net/browse/WEB-2498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/redux/reducers/misc.js
```diff
@@ -1044,6 +1044,7 @@ export const tideDashboardPatients = (state = initialState.tideDashboardPatients
     case types.FETCH_TIDE_DASHBOARD_PATIENTS_SUCCESS:
       return action?.payload?.results || initialState.tideDashboardPatients;
     case types.LOGOUT_REQUEST:
+    case types.CLEAR_TIDE_DASHBOARD_PATIENTS:
```
krystophv, line 1047: I think this could use a unit test (more of a guarantee of functionality into the future than a "does this do what we want")

krystophv (approved):
LGTM 🚀

app/redux/reducers/misc.js
krystophv: Oh, right - sorry; my mind was in Uploader where the reducer unit tests are under `misc.test.js` instead of individual files per reducer and I was just quickly trying to match up file-for-file unit tests at the end of the review (I had already hit the "viewed" checkbox on that unit test, so out-of-sight/out-of-mind took over). All good 👍🏼

PR #1281, opened 2023-08-29
[WEB-2543,WEB-2542] Clear patient in view state when mounting dashboard

clintonium-119:
[WEB-2543], [WEB-2542]

Similar to what we do for the clinic and clinician patients lists, we need to clear the currently viewed patient here as well.

While it would be simpler to do it on unmount of the patient data view component, we currently don't because we intentionally retain the currently loaded patient so while the user navigates between the data views and profile/share pages.

[WEB-2543]: https://tidepool.atlassian.net/browse/WEB-2543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WEB-2542]: https://tidepool.atlassian.net/browse/WEB-2542?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/pages/dashboard/TideDashboard.js
+++ b/app/pages/dashboard/TideDashboard.js
@@ -532,6 +532,7 @@ export const TideDashboard = (props) => {
   const { set: setToast } = useToasts();
   const selectedClinicId = useSelector((state) => state.blip.selectedClinicId);
   const loggedInUserId = useSelector((state) => state.blip.loggedInUserId);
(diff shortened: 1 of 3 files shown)


krystophv (approved):
LGTM 🚢

PR #1325, opened 2024-01-12
[WEB-2633,WEB-2634] Add additional visitor and account info to Pendo

clintonium-119:
See [WEB-2633], [WEB-2634]

In working on this, I also updated the existing middleware login and select clinic handlers to ensure that initialize is not only called on login, since the `selectClinic` action is sometimes fired before login.  It will now always use pendo.initialize for the first pendo call made, regardless of the action handler being called.

One other additional change was that we had previously been not setting a clinic as the account context on login if the user had multiple clinics.  I've added an exception to that, where, if the user has the `selectedClinicId` available in state at time of login, it will use that as the account context.

[WEB-2633]: https://tidepool.atlassian.net/browse/WEB-2633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WEB-2634]: https://tidepool.atlassian.net/browse/WEB-2634?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


krystophv (approved):
Nice work, LGTM 💯

PR #1372, opened 2024-04-24
[WEB-2933] Consolidate links styles in Limit UI

clintonium-119:
[WEB-2933] 
Some of the link styles stopped working correctly when we switched to theme-ui.
This consolidates them by defining them at the theme level (and in the correct location for theme-ui)


krystophv (approved):
LGTM 🚢

PR #1377, opened 2024-05-09
[WEB-2945] RPM config form range selector tweaks

clintonium-119:
[WEB-2945] 

Basically, adjusts the date-blocking logic for the RPM date range UX, and also ignores the browser time zone, and treats all calendar dates as UTC, only converting to the requested time zone on submission, which fixes a few edge case issues that could affect selecting dates at the start of the range for certain time zones.

It also reduces the max days in past to 59 days from 60, since we only generate pop health stats on the backend for the past 60 days in UTC time, leaving some time zones with no data for up to 12 hours on the 60th day back.

[WEB-2945]: https://tidepool.atlassian.net/browse/WEB-2945?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/clinic/RpmReportConfigForm.js
+++ b/app/components/clinic/RpmReportConfigForm.js
@@ -2,7 +2,6 @@ import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { withTranslation } from 'react-i18next';
 import { useDispatch, useSelector } from 'react-redux';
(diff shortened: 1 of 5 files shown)


krystophv (approved):
LGTM 💯

PR #1389, opened 2024-06-03
[WEB-2969,WEB-2974] Date range selection fix for certain time zones

clintonium-119:
[WEB-2969]

There were a few timezone-related bugs that surfaced for users ahead of UTC time.  These were addressed in 2 main ways (I've added some comments to sections of code as I saw fit, but wanted to call them out here too):

1. I've set the global `moment` object to default to "UTC" whenever the rpm form modal is open.  It automatically resets to the default whenever the modal is closed, or the component unmounts.  [This issue](https://github.com/react-dates/react-dates/issues/1138#issuecomment-870240972) (among some others I found while debugging how to get `react-dates` to work with UTC internally) is one I came across earlier in development, and had hoped to avoid, but it's necessary to get this working as designed, and has the benefit of removing some workaround code cruft I had in place.
2. We needed the selectable dates to be relative to the timezone chosen in the UI.  What I have is a `utcDayShift` that gets tracked when the timezone changes, and adjusts the available chart dates appropriately, whether it is currently a calendar date ahead or behind of UTC.  This day shift also gets passed to the schema validation whenever the timezone changes.

Also addresses [WEB-2974] by deleting empty `search` string parameters from the report config

[WEB-2969]: https://tidepool.atlassian.net/browse/WEB-2969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[WEB-2974]: https://tidepool.atlassian.net/browse/WEB-2974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/clinic/RpmReportConfigForm.js
+++ b/app/components/clinic/RpmReportConfigForm.js
@@ -16,7 +16,7 @@ import { Element, scroller } from 'react-scroll';
 
 import { useLocalStorage } from '../../core/hooks';
 import { addEmptyOption, getCommonFormikFieldProps } from '../../core/forms';
-import { rpmReportConfigSchema as validationSchema, timezoneOptions, dateRegex } from '../../core/clinicUtils';
+import { rpmReportConfigSchema, timezoneOptions, dateRegex } from '../../core/clinicUtils';
 import { Body0 } from '../../components/elements/FontStyles';
 import Select from '../elements/Select';
 import DateRangePicker from '../elements/DateRangePicker';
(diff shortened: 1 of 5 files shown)


app/components/clinic/RpmReportConfigForm.js
```diff
@@ -157,33 +154,85 @@ export const RpmReportConfigForm = props => {
...
   } = formikContext;
 
   // Set to default state when dialog is newly opened
   useEffect(() => {
     if (open) {
+      // Set global moment to use UTC timezone when config modal opens. This is required for our
+      // use within this form, since React-dates (used for our datepicker components) does not allow
+      // overriding the default moment-provided timezone, and we need to work with UTC dates for
+      // consistent behaviour across various timezones
+      setMomentToUTC();
```
krystophv, line 175: Minor nit: this has the feel of a use-before-define to me since these utility functions are defined after the `useEffect` sets.

krystophv (approved):
LGTM. Excellent work on this, timezones are the worst. 🚀

app/components/clinic/RpmReportConfigForm.js
clintonium-119: Because the functions are declared as functions, they do get hoisted above where they're called, but I'm fine with shifting their definitions upward to read better.

clintonium-119: Done

PR #1393, opened 2024-06-07
[WEB-2995] Fix day prior to start date not always being disabled

clintonium-119:
[WEB-2995]

When I removed the `getCalendarDate` method to simplify the code in the previous PR, I missed wrapping this one check in `moment.utc` like I had with all other, and, more importantly, needed the precision set to `true`.

Technically, the `moment.utc` wrapping could be left out, since the global moment should default to UTC while the form is open, but I want it to be consistent with the others, and explicit that we're dealing exclusively with UTC when glancing at the code.

[WEB-2995]: https://tidepool.atlassian.net/browse/WEB-2995?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


krystophv (approved):
LGTM 👍🏼

PR #1394, opened 2024-06-07
[WEB-2995] Small tweak to blocked end date range when end date input is selected

clintonium-119:
[WEB-2995] Noticed that there was a similar intermittent error on the end date.

Also, resetting the form whenever the modal closes so it will automatically reset the default dates when it next opens (fixes a very remote UI edge case where a user can change timezone that allows a future date to be selected, then closes the modal, and reopens it in a non-future-date time zone).

[WEB-2995]: https://tidepool.atlassian.net/browse/WEB-2995?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/clinic/RpmReportConfigForm.js
+++ b/app/components/clinic/RpmReportConfigForm.js
@@ -172,6 +172,7 @@ export const RpmReportConfigForm = props => {
     values,
     setValues,
     validateForm,
+    resetForm,
   } = formikContext;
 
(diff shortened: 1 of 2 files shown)


krystophv (approved):
LGTM 🚀

PR #1397, opened 2024-06-25
Fix travis builds by removing buildx install step

clintonium-119:
Fix travis builds by removing buildx install step


krystophv (approved):
LGTM 👍🏼

PR #1413, opened 2024-08-05
Only call generateAGPImagesSucces if no errors were thrown

clintonium-119:
I realized that one of the tests was broken due to firing the generate PDF success handler even if an error was thrown.  Also, the tests weren't always firing in the anticipated order which caused indeterminite results.

Once approved, I'll also back-merge this to `develop` so that we have it in a passing state


krystophv (approved):
LGTM 🚀

PR #1419, opened 2024-08-16
[WEB-3068] Skip device selection if only one pump and cgm option available

clintonium-119:
[WEB-3068]

I spent an inordinate amount of time trying to unit test this one, but the stepper form wrapper has proved to be difficult to navigate.  Given that this is considered a temporary solution/situation, and the plans to alter our testing approach, I felt it fine to forego unit-testing this change.

Happy to hear your thoughts if you have any pushback on this.

[WEB-3068]: https://tidepool.atlassian.net/browse/WEB-3068?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/pages/prescription/PrescriptionForm.js
+++ b/app/pages/prescription/PrescriptionForm.js
@@ -53,6 +53,7 @@ import * as actions from '../../redux/actions';
 import { components as vizComponents } from '@tidepool/viz';
 
 import {
+  cgmDeviceOptions,
   defaultUnits,
   deviceIdMap,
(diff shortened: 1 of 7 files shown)


krystophv (approved):
LGTM 🚢 

Not terribly worried about unit testing given the broader context and that this will be a high-touch-point experience for a bit.

PR #1420, opened 2024-08-21
[WEB-3022] Implement Palmtree guardrails

clintonium-119:
[WEB-3022]

Implements the guardrails according to the ticket.

Other changes:

- Had to re-implement the time options for the ScheduleForm component as well, since the html5 time field no longer supports the 'step' prop as it had when this was originally developed.
- Also, made the time increment and max number of schedules configurable
- Fixed a navigation error when single-step editing from the review page
- Remove the `createdUserId` from the revision hash to account for a similar backend change.

[WEB-3022]: https://tidepool.atlassian.net/browse/WEB-3022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/pages/prescription/PrescriptionForm.js
+++ b/app/pages/prescription/PrescriptionForm.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { withTranslation } from 'react-i18next';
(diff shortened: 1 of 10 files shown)


krystophv (approved):
LGTM 🔥 

Excellent work

PR #1466, opened 2024-11-18
[WEB-3276] Fix for invite expired status not being correctly determined on TIDE dash

clintonium-119:
[WEB-3276]

Also includes small UI regression fix where name column was no longer being truncated by an ellipsis


krystophv (approved):
LGTM :shipit:

PR #1477, opened 2024-11-26
[WEB-3277] - Preserve Patient List Filter State

henry-tp:
[WEB-3277]

### Summary of changes:

- When navigating away from the patient list and then navigating back, the show/hide and search filters remain applied

-----

### Note to reviewer: 

Can you please double check that made the necessary change to the relevant views? I'm still not fully aware of all of the different ways of viewing the patient list so I'm kind of worried I may have missed some.

[WEB-3277]: https://tidepool.atlassian.net/browse/WEB-3277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/unit/pages/ClinicPatients.test.js
```diff
@@ -577,7 +581,6 @@ describe('ClinicPatients', () => {
         </Provider>
       );
 
-      wrapper.find('#patients-view-toggle').hostNodes().simulate('click');
```
henry-tp, line 580: Hope you're okay with this. Because the state is defined in the redux store, and we're testing all of the functionality where the patient list is visible, I need to default the redux store value of "Show Patients" to true here, which is not actually the default state of the page.

I'm not aware of any capability we have to actually hook up a functioning redux store inside the test such that we can actually default "Show Patients" to false and then add in the simulated interaction.

So my thought is that we test all of this functionality while the redux state is set to true, and then also add tests to make sure the the redux state will get changed by the inputs.

test/unit/pages/ClinicPatients.test.js
```diff
@@ -149,6 +149,10 @@ describe('ClinicPatients', () => {
         settingClinicPatientLastReviewed: defaultWorkingState,
         revertingClinicPatientLastReviewed: defaultWorkingState,
       },
+      patientListFilters: {
+        patientListSearchTextInput: '',
+        isPatientListVisible: true
```
henry-tp, line 154: The variable names are quite long here but I slightly prefer greppability over readability when dealing with redux because the variables end up getting referenced everywhere.

henry-tp:
Hey @krystophv or @clintonium-119 , could I get a review on this one? :)

test/unit/pages/ClinicPatients.test.js
krystophv: This approach seems reasonable to me. The alternative, as I see it, would be to just use the _actual_ redux store with our reducers wrapped to capture actions on the way through, but that really starts to fall outside the scope of "unit" tests.

app/redux/reducers/patientListFilters.js
```diff
@@ -0,0 +1,22 @@
+import update from 'immutability-helper';
+import * as actionTypes from '../constants/actionTypes';
+
+const initialState = {
+  isPatientListVisible: false,
+  patientListSearchTextInput: '',
+}
```
krystophv, line 7: I'm a little torn on this one. On one hand I do see the value/desire to keep the initial state colocated with the rest of the code, but on the other I do like the self-documenting nature of the `initialState.js` that effectively shows what the whole Redux state tree should look like in the application. I don't have _strong_ feelings about it, and in the RTK world, this pattern feels more idiomatic and since that's where we're 🤞🏼 headed, I'm inclined to let this be the start of the colocation.

app/pages/clinicworkspace/ClinicianPatients.js
```diff
@@ -57,15 +57,16 @@ export const ClinicianPatients = (props) => {
...
-  const [searchText, setSearchText] = React.useState('');
   const [page, setPage] = useState(1);
   const [pageCount, setPageCount] = useState();
   const [selectedPatient, setSelectedPatient] = useState(null);
   const [loading, setLoading] = useState(false);
   const [patientFormContext, setPatientFormContext] = useState();
   const rowsPerPage = 8;
 
+  const patientListSearchTextInput = useSelector(({ blip }) => blip.patientListFilters.patientListSearchTextInput);
+  const isPatientListVisible = useSelector(({ blip }) => blip.patientListFilters.isPatientListVisible);
```
krystophv, line 68: Just curious about why this is split across two `useSelector` calls instead of the object deconstruction in `ClinicPatients`

henry-tp: No real reason, just my inattention. I believe that splitting into multiple useSelector calls is technically better to for reduce re-rendering when returning primitives, but I think in this case we consume both values in the state so it doesn't really matter.

Unified here: a0c47bdd69b300108edb1f6ba9d717e9ff06b9d6

app/redux/actions/sync.js
```diff
@@ -2462,3 +2462,17 @@ export function setPendoData(data) {
     payload: { data },
   };
 }
+
+export function setIsPatientListVisible(data) {
+  return {
+    type: ActionTypes.SET_IS_PATIENT_LIST_VISIBLE,
+    payload: data,
```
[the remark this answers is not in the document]
henry-tp: Oh, I had no idea! This is really interesting, and logical that there should be a certain structure to adhere to. Makes sense. Also I realized that `data` is not a good variable name on my part. Thanks for this catch!

Use objects as payloads: e6d7e9a81ef3ade6fc6c1c41c86d4d4d0cdb2ed4

Added tests: aba2abe3d91b2d4ad93d9980def606bd16a64fa1

app/redux/reducers/patientListFilters.js
henry-tp: Ahh! This was not an intentional decision I made, just my inattention, and we used RTK at my previous role so I probably just put it here by muscle memory. Funny, I think we talked about this during my interview with you. If you're good with this, then perhaps we can leave it here and start the co-location.

henry-tp:
@krystophv  I was doing some manual testing and I realized that the filters would be applied to a new workspace if you were to switch workspaces, which I believe is _not_ the intent of this ticket

https://github.com/user-attachments/assets/7f9fbc88-3cb4-4dcb-a84b-c2ec7596a6f9

I'm slightly afraid of other situations like this =( If you can think of any let me know.

For now, I have added added some calls to reset the values if you switch the workspace this way: 751b9369

krystophv:
> I'm slightly afraid of other situations like this =( If you can think of any let me know.
> 
> For now, I have added added some calls to reset the values if you switch the workspace this way: [751b936](https://github.com/tidepool-org/blip/commit/751b9369603da7cdd85845b1de8934fe85e50a97)
> 
> As an alternative to the above, I was considering clearing the filters whenever the user dispatches `actions.async.selectClinic()` (by putting code in `async.js:3010`) which might solve the issue more holistically. Let me know if that's a better idea.

I think that, at least for now, I prefer the explicit resets instead of having side effects from other actions. I don't think that we'd run into any specific issues for our current use cases, but I'd rather keep the actions pretty atomic and isolated to their own state tree branches. The only other place I think might need to be handled is in the workspaces page: https://github.com/tidepool-org/blip/blob/0cf8f8327e1686a8ab9db0cfa127cbccfd141bdc/app/pages/workspaces/workspaces.js#L288-L296

app/pages/workspaces/workspaces.js
```diff
@@ -291,6 +291,8 @@ export const Workspaces = (props) => {
       : ['Clinic - Workspaces - Go to private workspace'];
 
     trackMetric(...metric);
+    dispatch(actions.sync.setPatientListSearchTextInput(''));
+    dispatch(actions.sync.setIsPatientListVisible(false));
```
henry-tp, line 295: @krystophv makes sense, and I agree that resetting these filters shouldn't be a concern of `selectClinic`.

Added one more set of reset calls for the Go To Workspace button: a7b65eb3

krystophv (approved):
All updates LGTM. Nice work! 🚀

PR #1478, opened 2024-11-26
[WEB-3297,WEB-3298] Ensure connection status updates after resending email

clintonium-119:
[WEB-3297] [WEB-3298]

### DIFF
--- a/app/components/clinic/PatientForm.js
+++ b/app/components/clinic/PatientForm.js
@@ -227,7 +227,9 @@ export const PatientForm = (props) => {
 
     if (!isFirstRender && !inProgress) {
       if (completed) {
+        // Close the resend email modal and refetch patient details to update the connection status
         setShowResendDexcomConnectRequest(false);
+        fetchPatientDetails();
 
         setToast({
           message: successMessage,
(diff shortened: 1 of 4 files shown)


app/components/clinic/PatientForm.js
```diff
@@ -329,6 +331,10 @@ export const PatientForm = (props) => {
     dispatch(actions.async.sendPatientDexcomConnectRequest(api, selectedClinicId, patient.id));
   }
 
+  function fetchPatientDetails() {
+    dispatch(actions.async.fetchPatientFromClinic.bind(null, api, selectedClinicId, patient.id)());
```
krystophv, line 335: Is there an advantage to using the `.bind()` here if the function is always immediately executed?

test/unit/redux/reducers/tideDashboardPatients.test.js
```diff
@@ -69,4 +69,23 @@ describe('tideDashboardPatients', () => {
       ] } });
     });
   });
+
+  describe('fetchPatientFromClinicSuccess', () => {
+    it('should set state to initial empty state', () => {
```
krystophv, line 74: Description didn't get updated from it's original copy :)

app/components/clinic/PatientForm.js
clintonium-119: Ah, good catch.  Will forego the bind.

test/unit/redux/reducers/tideDashboardPatients.test.js
clintonium-119: Yep :)

PR #1483, opened 2024-12-05
[WEB-2800] - Print Modal Styles

henry-tp:
[WEB-2800]

[Figma Link ](https://www.figma.com/design/8TgXWZZNFpBIeXJzkzB1S8/Uploader-in-Web-(Updated-Styles)?node-id=613-337&node-type=canvas&t=TDlqHv9nYhIDkKWW-0) (Mock at the bottom)

[WEB-2800]: https://tidepool.atlassian.net/browse/WEB-2800?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/themes/baseTheme.js
```diff
@@ -225,6 +225,7 @@ const text = {
   mediumTitle: {
     ...titleText,
     fontWeight: 'medium',
+    fontSize: 3
```
henry-tp, line 228: I searched through blip and it looks like all instances of `<MediumTitle>` are used within the `<DialogTitle>`, meaning that this is somewhat a dialog-specific configuration. So I think that this is okay, as opposed to spinning off a new `variant`.

app/components/elements/Dialog.js
```diff
@@ -24,13 +24,20 @@ export function DialogTitle(props) {
...
     <Flex
       p={3}
-      sx={{ alignItems: 'center', justifyContent: 'space-between', borderBottom: props.divider ? borders.divider : 'unset', ...sx }}
+      sx={{ alignItems: 'center', borderBottom: props.divider ? borders.divider : 'unset', ...sx }}
       {...dialogTitleProps}
     >
-      {children}
+      <Flex 
+        mx={[0, 'auto']} 
+        pl={closeIcon ? ICON_WIDTH : 0} // ensure Title is properly centered when icon present
```
henry-tp, line 37: This is somewhat a hack to offset the width of the icon when trying to center the title
image

I couldn't think of a better way to easily do it. Typically I would set the icon position to absolute, but I was unable to make that happen with the `sx` prop, I don't think it is supported.

app/components/elements/Dialog.js
```diff
@@ -24,13 +24,20 @@ export function DialogTitle(props) {
     ...dialogTitleProps
   } = props;
 
+  const ICON_WIDTH = 28;
+
   return (
     <Flex
       p={3}
-      sx={{ alignItems: 'center', justifyContent: 'space-between', borderBottom: props.divider ? borders.divider : 'unset', ...sx }}
+      sx={{ alignItems: 'center', borderBottom: props.divider ? borders.divider : 'unset', ...sx }}
```
henry-tp, line 32: Based on the Figma doc, my understanding is that we are trying to align all modal styles, so I think having the following changes reflected on all modals is okay:
- Center Title
- Reduce Title Font Size

app/components/elements/Dialog.js
```diff
@@ -24,13 +24,21 @@ export function DialogTitle(props) {
...
       {...dialogTitleProps}
     >
-      {children}
+      <Flex
+        // offset by Icon width to ensure Title is properly centered when closeIcon is present
+        pl={closeIcon ? ICON_WIDTH : 0}
```
henry-tp, line 37: This is somewhat a hack to offset the width of the icon when trying to center the title
image


I couldn't think of a better way to easily do it. Typically I would set the icon to `position: absolute`, but I was unable to make that happen with the `sx` prop.

krystophv: I think that, as far as CSS hacks go, this is pretty readable for future readers, which is about the best you can hope for. 😄

krystophv (approved):
LGTM 🚢

app/components/elements/Dialog.js
clintonium-119: I missed this similar PR to the one for WEB-2802, where I left the following comment:

| It's tricky, but I believe the best way to do have middle piece of content be true-centered in-between elements of potentially variable sizes with flex is to use flex-grow: 1, flex-basis: 0 on the outer elements.

| I ended up having to do this on the Dialog title to accommodate adding a back button to some modals as part of the data connection work. [Here's a link to the commit](https://github.com/tidepool-org/blip/pull/1485/commits/15bc53b554512f24df2d833541c67df34b2d67bb).

I recommend the flex-based approach, as it will be necessary when we add a potential back button and text on the other side of the title.

henry-tp: Thanks @clintonium-119, that seems to work!

image

If you don't mind, I've cherry-picked your commit into my branch c7603f2977439e099e0b33c9adf43711c8456575

henry-tp: Thanks @clintonium-119 and @krystophv for the review :)

The approach works great. If you don't mind, I've cherry-picked your commit into my branch to avoid merge error, since your commit only touches this file. 

c7603f2977439e099e0b33c9adf43711c8456575

image

PR #1486, opened 2024-12-10
[WEB-3301] - Preserve Patient Tags

henry-tp:
[WEB-3301]

[WEB-3301]: https://tidepool.atlassian.net/browse/WEB-3301?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/pages/clinicworkspace/ClinicPatients.js
```diff
@@ -724,6 +725,9 @@ export const ClinicPatients = (props) => {
   }, [deletingClinicPatientTag, handleAsyncResult, handleCloseClinicPatientTagUpdateDialog, previousDeletingClinicPatientTag?.inProgress, t]);
 
   useEffect(() => {
+    // Prevent this effect from firing on logout, which would clear all activeFilters from localStorage
+    if (!clinic) return;
+
```
henry-tp, line 730: The patientTags are actually being stored properly in localStorage. However, they are being wiped out by the firing of this function on user logout. 

This function's responsibility is to remove tags from the currently applied set of filters if they have been removed from the clinic's catalogue of tags. When the clinic's tag library is modified, the effect is fired, and the function un-applies all tag filters that don't exist in the clinic tag library anymore.

The issue is that when the user logs out, the clinic is momentarily set to `undefined`. This component sees that the clinic has no tags and assumes ALL tags have been removed from the clinic library, and therefore we should clear all filters.

app/pages/clinicworkspace/ClinicPatients.js
```diff
@@ -531,7 +531,8 @@ export const ClinicPatients = (props) => {
     [clinicBgUnits]
   );
 
-  const [activeFilters, setActiveFilters] = useLocalStorage('activePatientFilters', defaultFilterState, true);
+  const activeFiltersStorageKey = `activePatientFilters/${loggedInUserId}/${selectedClinicId}`;
```
henry-tp, line 534: Question for sanity check - a "clinic" and a "workspace" are functionally the same thing in our platform, correct? The requirements are that the filters are stored `per user, per workspace`. This seems to work with that assumption in mind, but I just want to triple check.

clintonium-119: Yes, that's correct.

app/pages/clinicworkspace/ClinicPatients.js
clintonium-119: Good job diagnosing this :+1:

clintonium-119 (approved):
Great job, @henry-tp :fire:

PR #1489, opened 2024-12-13
[WEB-3336] - TIDE Dashboard - Drawer Prototype

henry-tp:
[WEB-3336]

[WEB-3336]: https://tidepool.atlassian.net/browse/WEB-3336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/pages/dashboard/buildGenerateAGPImagesFunction.js
```diff
@@ -0,0 +1,58 @@
+import _ from 'lodash';
+import { utils as vizUtils } from '@tidepool/viz';
+import Plotly from 'plotly.js-basic-dist-min';
+import * as actions from '../../redux/actions';
+
+const buildGenerateAGPImages = (dispatch) => {
```
henry-tp, line 6: Copied from `app/pages/patientdata/patientdata.js:922`. The fake "props" object is just to prevent needing to change the code.

app/pages/dashboard/getQueries.js
```diff
@@ -0,0 +1,162 @@
...
+  agpCGM: {
+    bgSource: 'cbg',
+  },
+  settings: {
+    touched: false,
+  },
+  excludedDevices: [],
+};
+
+const getQueries = (
```
henry-tp, line 70: To create this function `getQueries`, I started at [this function call to generatePDF()](https://github.com/tidepool-org/blip/blob/WEB-3336-tide-drawer/app/pages/patientdata/patientdata.js#L1061) and worked backwards up the callback chain, which should return that `queries` argument. 

Questions to follow.

app/pages/dashboard/getQueries.js
```diff
@@ -0,0 +1,162 @@
...
+  })();
+
+  const timePrefs = (() => {
+    const latestTimeZone = data?.metaData?.latestTimeZone;
+    const queryParams = {};
+
+    const localTimePrefs = utils.getTimePrefsForDataProcessing(latestTimeZone, queryParams);
+    
+    return localTimePrefs;
+  })();
```
henry-tp, line 99: Working backwards in the callback chain to get to the `queries` arg, I get to these 2 values which are defined by `this.state.bgPrefs` and `this.state.timePrefs`. Because they are held in component state, the process of how they get their value is a little nebulous.

From code reading and testing, it seems like `bgPrefs` represent the `clinic's` preferences for bg readings, while timePrefs represent the _patient's_ device preferences for time. Is this understanding correct?

I _believe_ they are defined [here in componentWillReceiveProps](https://github.com/tidepool-org/blip/blob/WEB-3336-tide-drawer/app/pages/patientdata/patientdata.js#L1694-L1718). Unfortunately, I can't simply copy and paste that code because it is coupled to component lifecycle, so I have written this batch of code to get us `bgPrefs` and `timePrefs`, and I think it works and is accurate. Would be awesome if you could help me double check this.

app/pages/dashboard/getQueries.js
```diff
@@ -0,0 +1,160 @@
...
```
henry-tp, line 97: Working backwards in the callback chain to get to the `queries` arg, I get to these 2 values which are defined by `this.state.bgPrefs` and `this.state.timePrefs`. Because they are held in component state, the process of how they get their value is a little nebulous.

From code reading and testing, it seems like `bgPrefs` represent the `clinic's` preferences for bg readings, while timePrefs represent the _patient's_ device preferences for time. Is this understanding correct?

I _believe_ they are defined [here in componentWillReceiveProps](https://github.com/tidepool-org/blip/blob/WEB-3336-tide-drawer/app/pages/patientdata/patientdata.js#L1694-L1718). Unfortunately, I can't simply copy and paste that code because it is coupled to component lifecycle, so I have written this batch of code to get us `bgPrefs` and `timePrefs`, and I think it works and is accurate. 

Would be awesome if you could help me double check this work.

app/redux/actions/async.js
```diff
@@ -1169,7 +1170,10 @@ export function fetchPatientData(api, options, id) {
 
       // We only add the data to the worker if another patient id has not been fetched
       // while we waited on this one, and we are still on an app view specific to that patient
-      if (location.pathname.indexOf(id) >= 0 && (!fetchingPatientId || fetchingPatientId === id)) {
+      if (
+        options.forceDataWorkerAddDataRequest ||
+        (location.pathname.indexOf(id) >= 0 && (!fetchingPatientId || fetchingPatientId === id))
+      ) {
```
henry-tp, line 1176: Here, I force `actions.async.fetchPatientData()` to call `actions.worker.dataWorkerAddDataRequest()`.

I didn't want to do it this way, I would have preferred to call `actions.worker.dataWorkerAddDataRequest()` from my component after we verify that the patient is loaded. However, the call to `dataWorkerAddDataRequest()` requires `options.startDate` as an arg, and the startDate is generated in a complex fashion within this function. I couldn't come up with a better way.

app/pages/dashboard/getOpts.js
```diff
@@ -0,0 +1,102 @@
+import moment from 'moment-timezone';
+import _ from 'lodash';
+import get from 'lodash/get';
+import { utils as vizUtils } from '@tidepool/viz';
+
+const getTimezoneFromTimePrefs = vizUtils.datetime.getTimezoneFromTimePrefs;
+
+const getOpts = (
+  data // data from redux (state.blip.data)
+) => {
```
henry-tp, line 10: To create this function `getOpts`, I started at [this call to onClickPrint()](https://github.com/tidepool-org/blip/blob/WEB-3336-tide-drawer/app/components/PrintDateRangeModal.js#L247) and worked backwards up the callback chain to figure out how the `printOpts` arg is determined.

app/pages/dashboard/PatientDrawerContent/useAGPImages/getQueries.js
```diff
@@ -0,0 +1,108 @@
...
+  agpCGM: {
+    bgSource: 'cbg',
+  },
+  settings: {
+    touched: false,
+  },
+  excludedDevices: [],
+};
+
+const getQueries = (
```
henry-tp, line 18: To create this parent function `getQueries`, I started at [this argument in a call to generatePDF()](https://github.com/tidepool-org/blip/blob/WEB-3336-tide-drawer/app/pages/patientdata/patientdata.js#L1061) and worked backwards up the callback chain to figure out how the `queries` arg is determined. 

Working backwards in the callback chain to get to the `queries` arg, I get to these 2 values which are defined by `this.state.bgPrefs` and `this.state.timePrefs`. Because they are held in component state, the process of how they get their value is a little nebulous.

From code reading and testing, it seems like `bgPrefs` represent the `clinic's` preferences for bg readings, while timePrefs represent the _patient's_ device preferences for time. Is this understanding correct?

I _believe_ they are defined [here in componentWillReceiveProps](https://github.com/tidepool-org/blip/blob/WEB-3336-tide-drawer/app/pages/patientdata/patientdata.js#L1694-L1718). Unfortunately, I can't simply copy and paste that code because it is coupled to component lifecycle, so I have written this batch of code to get us `bgPrefs` and `timePrefs`, and I think it works and is accurate. 

Would be awesome if you could help me double check this work to generate the `queries` arg when we get to code review. I'm a little nervous about it.

app/pages/dashboard/PatientDrawerContent/useAGPImages/getOpts.js
```diff
@@ -0,0 +1,106 @@
+import moment from 'moment-timezone';
+import _ from 'lodash';
+import get from 'lodash/get';
+import { utils as vizUtils } from '@tidepool/viz';
+
+const getTimezoneFromTimePrefs = vizUtils.datetime.getTimezoneFromTimePrefs;
+
+const AGP_PERIOD_IN_DAYS = 7; // Temporarily using fixed value for prototype launch
+
+const getOpts = (
```
henry-tp, line 10: To create this function `getOpts`, I started at [this call to onClickPrint()](https://github.com/tidepool-org/blip/blob/WEB-3336-tide-drawer/app/components/PrintDateRangeModal.js#L247) and worked backwards up the callback chain to figure out how the `printOpts` arg is determined.

Would be awesome if you could help me double check this work when we get to code review, I'm a little nervous about it.

app/pages/dashboard/PatientDrawer/useAGPImages/getQueries.js
```diff
@@ -0,0 +1,108 @@
...
+  agpCGM: {
+    bgSource: 'cbg',
+  },
+  settings: {
+    touched: false,
+  },
+  excludedDevices: [],
+};
+
+const getQueries = (
```
henry-tp, line 18: To create this parent function `getQueries`, I started at [this argument in a call to generatePDF()](https://github.com/tidepool-org/blip/blob/WEB-3336-tide-drawer/app/pages/patientdata/patientdata.js#L1061) and worked backwards up the callback chain to figure out how the `queries` arg is determined. 

Working backwards in the callback chain to get to the `queries` arg, I get to these 2 values which are defined by `this.state.bgPrefs` and `this.state.timePrefs`. Because they are held in component state, the process of how they get their value is a little nebulous.

From code reading and testing, it seems like `bgPrefs` represent the `clinic's` preferences for bg readings, while timePrefs represent the _patient's_ device preferences for time. Is this understanding correct?

I _believe_ they are defined [here in componentWillReceiveProps](https://github.com/tidepool-org/blip/blob/WEB-3336-tide-drawer/app/pages/patientdata/patientdata.js#L1694-L1718). Unfortunately, I can't simply copy and paste that code because it is coupled to component lifecycle, so I have written this batch of code to get us `bgPrefs` and `timePrefs`, and I think it works and is accurate. 

Would be awesome if you could help me double check this work to generate the `queries` arg when we get to code review. I'm a little nervous about it.

app/pages/dashboard/PatientDrawer/useAGPImages/getOpts.js
```diff
@@ -0,0 +1,106 @@
+import moment from 'moment-timezone';
+import _ from 'lodash';
+import get from 'lodash/get';
+import { utils as vizUtils } from '@tidepool/viz';
+
+const getTimezoneFromTimePrefs = vizUtils.datetime.getTimezoneFromTimePrefs;
+
+const AGP_PERIOD_IN_DAYS = 7; // Temporarily using fixed value for prototype launch
+
+const getOpts = (
```
henry-tp, line 10: To create this function `getOpts`, I started at [this call to onClickPrint()](https://github.com/tidepool-org/blip/blob/WEB-3336-tide-drawer/app/components/PrintDateRangeModal.js#L247) and worked backwards up the callback chain to figure out how the `printOpts` arg is determined.

Would be awesome if you could help me double check this work when we get to code review, I'm a little nervous about it.

app/pages/dashboard/PatientDrawer/CGMStatistics.js
```diff
@@ -0,0 +1,126 @@
...
+          coefficientOfVariation: { coefficientOfVariation }
+        },
+      }
+    }
+  } = agpCGM;
+
+  const timezoneName = vizUtils.datetime.getTimezoneFromTimePrefs(timePrefs);
+
+  const avgGlucosePrecision = bgUnits === MGDL_UNITS ? 0 : 1;
+  const avgGlucoseTarget    = bgUnits === MGDL_UNITS ? '154' : '8.6';
```
henry-tp, line 77: From what I can tell this value appears to be static and not decided by `bgPrefs`?

app/pages/dashboard/PatientDrawer/MenuBar/CGMClipboardButton.js
```diff
@@ -0,0 +1,89 @@
...
+    }, 1000);
+
+    return () => {
+      clearTimeout(buttonTextEffect);
+    }
+  }, [buttonState])
+
+  const sensorUsage = agpCGM?.data?.current?.stats?.sensorUsage?.sensorUsage || 0;
+
+  const isDisabled = !agpCGM || sensorUsage < 86400000; // minimum 24 hours
```
henry-tp, line 70: I noticed that we still have some statistical data even if the AGP as a whole has insufficient data.

I inferred this value from [these lines in Viz.](https://github.com/tidepool-org/viz/blob/0feca7c7e8b511f90e4abd0bdb9b8c3034b8958c/src/modules/print/utils/AGPUtils.js#L62-L82)

This shows up in the following manner:

image

app/pages/dashboard/PatientDrawer/MenuBar/CGMClipboardButton.js
```diff
@@ -0,0 +1,89 @@
...
+
+  return clipboardText;
+}
+
+const STATE = {
+  DEFAULT: 'DEFAULT',
+  CLICKED: 'CLICKED',
+}
+
+const CGMClipboardButton = ({ patientName, agpCGM }) => {
```
henry-tp, line 54: For this component, I tried my hardest with `<ClipboardButton />` from Viz but I just _could not_ get the callback to work on my machine for some reason, even though the API seems dead simple and all of the _other_ ClipboardButtons on the platform do seem to work. Therefore, I built this one hastily. If you would prefer we go back to the standardized Viz component, I'm happy to do that but I might need your help.

app/pages/dashboard/PatientDrawer/useAgpCGM/buildGenerateAGPImagesFunction.js
```diff
@@ -0,0 +1,64 @@
+/* eslint-disable no-undef */
+
+import _ from 'lodash';
+import { utils as vizUtils } from '@tidepool/viz';
+import Plotly from 'plotly.js-basic-dist-min';
+import * as actions from '../../../../redux/actions';
```
henry-tp, line 6: This function is a rip of [this class method](https://github.com/tidepool-org/blip/blob/a95a74b3443f3f83e29491b779c0a4ea5edad8e5/app/pages/patientdata/patientdata.js#L922-L964) in PatientData.js.

I apologize for using a funky pattern here without checking in with the group, but essentially what I'm doing here is wrapping the main function in 2 layers of functions. The outer layers allow for a pseudo-dependency injection instead of having the function import the dependencies (vizUtils and Plotly) directly. In the test file it makes it much easier to mock vizUtils and Plotly.

app/pages/dashboard/PatientDrawer/useAgpCGM/getOpts.js
```diff
@@ -0,0 +1,105 @@
+import moment from 'moment-timezone';
+import _ from 'lodash';
+import get from 'lodash/get';
+import { utils as vizUtils } from '@tidepool/viz';
+
+const getTimezoneFromTimePrefs = vizUtils.datetime.getTimezoneFromTimePrefs;
+
```
henry-tp, line 7: To create this function `getOpts`, I started at [this call to onClickPrint()](https://github.com/tidepool-org/blip/blob/WEB-3336-tide-drawer/app/components/PrintDateRangeModal.js#L247) and worked backwards up the callback chain to figure out how the `printOpts` arg is determined.

Would be awesome if you could help me double check this work when we get to code review, I'm a little nervous about it.

app/pages/dashboard/PatientDrawer/useAgpCGM/getQueries.js
```diff
@@ -0,0 +1,108 @@
...
+  },
+  agpCGM: {
+    bgSource: 'cbg',
+  },
+  settings: {
+    touched: false,
+  },
+  excludedDevices: [],
+};
+
```
henry-tp, line 17: To create this parent function `getQueries`, I started at [this argument in a call to generatePDF()](https://github.com/tidepool-org/blip/blob/WEB-3336-tide-drawer/app/pages/patientdata/patientdata.js#L1061) and worked backwards up the callback chain to figure out how the `queries` arg is determined. 

Working backwards in the callback chain to get to the `queries` arg, I get to these 2 values which are defined by `this.state.bgPrefs` and `this.state.timePrefs`. Because they are held in component state, the process of how they get their value is a little nebulous.

From code reading and testing, it seems like `bgPrefs` represent the `clinic's` preferences for bg readings, while timePrefs represent the _patient's_ device preferences for time. Is this understanding correct?

I _believe_ they are defined [here in componentWillReceiveProps](https://github.com/tidepool-org/blip/blob/WEB-3336-tide-drawer/app/pages/patientdata/patientdata.js#L1694-L1718). Unfortunately, I can't simply copy and paste that code because it is coupled to component lifecycle, so I have written this batch of code to get us `bgPrefs` and `timePrefs`, and I think it works and is accurate. 

Would be awesome if you could help me double check this work to generate the `queries` arg when we get to code review. I'm a little nervous about it.

app/pages/dashboard/PatientDrawer/useAgpCGM/getQueries.js
```diff
@@ -0,0 +1,108 @@
...
+const getQueries = (
+  data,
+  patient,
+  clinic,
+  opts,
+) => {
```
henry-tp, line 23: To create this parent function `getQueries`, I started at [this argument in a call to generatePDF()](https://github.com/tidepool-org/blip/blob/WEB-3336-tide-drawer/app/pages/patientdata/patientdata.js#L1061) and worked backwards up the callback chain to figure out how the `queries` arg is determined. 

Working backwards in the callback chain to get to the `queries` arg, I get to these 2 values which are defined by `this.state.bgPrefs` and `this.state.timePrefs`. Because they are held in component state, the process of how they get their value is a little nebulous.

From code reading and testing, it seems like `bgPrefs` represent the `clinic's` preferences for bg readings, while timePrefs represent the _patient's_ device preferences for time. Is this understanding correct?

I _believe_ they are defined [here in componentWillReceiveProps](https://github.com/tidepool-org/blip/blob/WEB-3336-tide-drawer/app/pages/patientdata/patientdata.js#L1694-L1718). Unfortunately, I can't simply copy and paste that code because it is coupled to component lifecycle, so I have written this batch of code to get us `bgPrefs` and `timePrefs`, and I think it works and is accurate. 

Would be awesome if you could help me double check this work to generate the `queries` arg when we get to code review. I'm a little nervous about it.

app/pages/dashboard/PatientDrawer/useAgpCGM/getOpts.js
```diff
@@ -0,0 +1,105 @@
...
+const getOpts = (
+  data, // data from redux (state.blip.data)
+  agpPeriodInDays,
+) => {
```
henry-tp, line 11: To create this function `getOpts`, I started at [this call to onClickPrint()](https://github.com/tidepool-org/blip/blob/WEB-3336-tide-drawer/app/components/PrintDateRangeModal.js#L247) and worked backwards up the callback chain to figure out how the `printOpts` arg is determined.

Would be awesome if you could help me double check this work when we get to code review, I'm a little nervous about it.

app/pages/dashboard/PatientDrawer/useAgpCGM/useAgpCGM.js
```diff
@@ -0,0 +1,119 @@
...
+  const hasImagesInState  = !!pdf?.opts?.svgDataURLS;
+  const hasPDFDataInState = !!pdf?.data;
+  const hasPatientInState = !!data?.metaData?.patientId;
+
+  if (hasImagesInState)  return STATUS.SVGS_GENERATED;
+  if (hasPDFDataInState) return STATUS.DATA_PROCESSED;
+  if (hasPatientInState) return STATUS.PATIENT_LOADED;
+
+  return STATUS.STATE_CLEARED;
+}
```
henry-tp, line 49: The decision to use this approach was really tough to make.

I also looked at using `blip.working` as an indicator of each step's success. However, in trying it out, I felt like it is intended for one-way operation of each process and there is no pre-existing mechanism to reset the values in it. Resets are needed because we need to be able to open and close the drawer multiple times for different patients. In addition, because `blip.working` does not store metadata, it was challenging to ensure that it represents the processing state for the _current_ patient instead of a different patient in the case of race conditions and such.

This approach of using relying on the data output is a bit more reliable in that sense, but it is more challenging to read and also to understand the logic of why we're making each check in the order that we are. I'm a little worried about the code being a bit brittle here too.

app/pages/dashboard/PatientDrawer/useAgpCGM/buildGenerateAGPImagesFunction.js
```diff
@@ -0,0 +1,59 @@
...
+
+export const buildGenerateAGPImages = (dispatch) => {
+  const props = {
+    generateAGPImagesSuccess: (images) => dispatch(actions.sync.generateAGPImagesSuccess(images)),
+    generateAGPImagesFailure: (error) => dispatch(actions.sync.generateAGPImagesFailure(error)),
+  }
+
```
henry-tp, line 13: This function is a rip of [this class method](https://github.com/tidepool-org/blip/blob/a95a74b3443f3f83e29491b779c0a4ea5edad8e5/app/pages/patientdata/patientdata.js#L922-L964) in PatientData.js.

app/pages/dashboard/PatientDrawer/useAgpCGM/buildGenerateAGPImages.js
```diff
@@ -0,0 +1,59 @@
+/* eslint-disable no-undef */
+
+import _ from 'lodash';
+import { utils as vizUtils } from '@tidepool/viz';
+import Plotly from 'plotly.js-basic-dist-min';
+import * as actions from '../../../../redux/actions';
+
```
henry-tp, line 7: This function is a rip of [this class method](https://github.com/tidepool-org/blip/blob/a95a74b3443f3f83e29491b779c0a4ea5edad8e5/app/pages/patientdata/patientdata.js#L922-L964) in PatientData.js.

app/pages/dashboard/PatientDrawer/Content.js
```diff
@@ -0,0 +1,113 @@
...
+    </Box>
+  );
+};
+
+const PROTOTYPE_AGP_PERIOD = 7; // Fixed for prototype;
+
+const Content = ({ api, patientId }) => {
+  const { t } = useTranslation();
+  
+  const { status, svgDataURLS, agpCGM } = useAgpCGM(api, patientId, PROTOTYPE_AGP_PERIOD);
```
henry-tp, line 60: So I recognize that component-specific custom hooks are not a common pattern in our platform and in general I'm actually not a big fan of them.

Part of the challenge in duplicating the AGP for this drawer was that in `<PatientData />`, AGP generation is highly coupled to the component render cycle (e.g. by using lifecycle hooks like `componentWillReceiveProps` to kick off each process step), meaning that AGP generation is quite tied to the `PatientData` component itself.

Here, I attempted to abstract out the data fetching logic, creating a layer of separation between AGP generation and component rendering. However, the abstraction would need to be dependent on hooks like `useEffect` to listen to changes in state/props, and thereby that's why I elected to go the custom hook route.

One benefit of this approach is that hypothetically, any component could call `useAgpCGM(patientId, periodInDays)` get an AGP out of it, if we need to start making the AGP available in more places in the future.

app/pages/dashboard/PatientDrawer/Content.js
```diff
@@ -0,0 +1,111 @@
...
+
+      {children}
+    </Box>
+  );
+};
+
+const Content = ({ api, patientId, agpPeriodInDays }) => {
+  const { t } = useTranslation();
+  
+  const { status, svgDataURLS, agpCGM } = useAgpCGM(api, patientId, agpPeriodInDays);
```
henry-tp, line 58: So I recognize that component-specific custom hooks are not a common pattern in our platform and in general I'm actually not a big fan of them.

Part of the challenge in duplicating the AGP for this drawer was that in `<PatientData />`, AGP generation is highly coupled to the component render cycle (e.g. by using lifecycle hooks like `componentWillReceiveProps` to kick off each process step), meaning that AGP generation is quite tied to the `PatientData` component itself.

Here, I attempted to abstract out all of that logic away from the component, creating a layer of separation between AGP generation and component rendering. However, any abstraction here would still be reliant upon hooks like `useEffect` to listen to changes in state/props, and thereby I elected to go the custom hook route.

One benefit of this approach is that hypothetically, any component could call `useAgpCGM(..)` get an AGP out of it, if we need to start making the AGP available in more places in the future.

app/pages/dashboard/PatientDrawer/CGMStatistics.js
clintonium-119: Correct.  This is static.

app/pages/dashboard/PatientDrawer/useAgpCGM/getOpts.js
```diff
@@ -0,0 +1,105 @@
...
+    // daily: 0,
+  };
+
+  const _presetDaysOptions = {
+    agpBGM: [14, 30],
+    agpCGM: [7, 14, 30],
+    // basics: [14, 21, 30, 90],
+    // bgLog: [14, 21, 30, 90],
+    // daily: [14, 21, 30, 90],
+  };
```
clintonium-119, line 58: Can these be removed?

app/pages/dashboard/PatientDrawer/useAgpCGM/getOpts.js
```diff
@@ -0,0 +1,105 @@
...
+  const dates = defaultDates();
+
+  const formatDateEndpoints = ({ startDate, endDate }) => (startDate && endDate ? [
+    startDate.valueOf(),
+    moment.utc(endDate).tz(timezoneName).add(1, 'day').startOf('day').valueOf(),
+  ] : []);
+
+  const opts = {
+    agpCGM:   { disabled: false, endpoints: formatDateEndpoints(dates.agpCGM) },
+    agpBGM:   { disabled: false, endpoints: formatDateEndpoints(dates.agpBGM) },
```
clintonium-119, line 94: Are we actually using agpBGM at this point, or do we only need to generate and define the various options within this file for agpCGM?

app/pages/dashboard/PatientDrawer/useAgpCGM/getOpts.js
clintonium-119: Generally, this looks fine - like you're pulling together all the necessary data as we do elsewhere.  I have a couple of small questions added in-line below, but nothing I'm worried about.

app/pages/dashboard/PatientDrawer/useAgpCGM/getQueries.js
clintonium-119: I'm impressed you were able to work your way through all this.  The approach you took for bg prefs and time prefs is correct.

Re: the queries - I could be mistaken, but I don't think you need the 'latestPumpUpload' metadata, nor do I think excludedDevices will accomplish anything here.  excludedDevices only comes in to play when the user unchecks a device from the devices list below the stats in the sidebar of any data view.  We don't have an equivalent on the Tide Dash.

Same question as elsewhere - Are we rendering the `agpBGM` chart as well?  If not, we could cut it out of this utility (unless you want it here for future use, and it's being disabled default and I just missed that)

clintonium-119 (changes requested):
Amazing work, @henry-tp .  This was a massive first main task for you to take, and you've knocked it out of the park.

I do have a bunch of small 'nits' - mostly code suggestion changes for missing semicolons or missing trailing commas in multiline arrays or objects.  

I could have just mentioned it once, but it's probably easier for me to just add it as a code suggestion where I saw them missing so you can just click through and github will automatically merge them without you having to search for them.

The only things I think we need to change for the v1 are the comments where I noticed that we are not rounding values in a consistent fashion with the AGP print views.  I've pointed out code examples as best I could to existing utilities that get used throughout viz for formatting stat values whereever we currently show them, be it the print views, current text output for the existing data views, and the blip sidebar stats.

It's OK if this is not possible to migrate everything over to the viz utils for v1, but we should do it at some point, since it will ensure consistency and avoid eyebrows raised from potential rounding mismatches.

app/pages/dashboard/PatientDrawer/useAgpCGM/getOpts.js
henry-tp: Addressed here: a1d5375daac53f489521c767864e0a087f6abc7d

app/redux/actions/async.js
henry-tp: Comment added: a1d5375daac53f489521c767864e0a087f6abc7d

app/pages/dashboard/PatientDrawer/useAgpCGM/useAgpCGM.js
henry-tp: Comment added: a1d5375daac53f489521c767864e0a087f6abc7d

app/pages/dashboard/PatientDrawer/PatientDrawer.js
```diff
@@ -0,0 +1,80 @@
...
+const useStyles = makeStyles({
+  paperAnchorRight: { overflow: 'visible' }
+});
+
+const DESKTOP_DRAWER_WIDTH = '1000px';
+const DRAWER_CLOSE_BUTTON_GAP = '70px';
+
+const getAgpPeriodInDays = (period) => {
+  switch(period) {
+    case '1d': // minimum 7-day AGP period, so return 7
```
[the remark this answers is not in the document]
henry-tp: Yeah, good idea. I'll use 14 since that is a standard AGP range. a1d5375daac53f489521c767864e0a087f6abc7d

app/pages/dashboard/PatientDrawer/MenuBar/CGMClipboardButton.js
```diff
@@ -0,0 +1,118 @@
...
+    let buttonTextEffect = setTimeout(() => {
+      setButtonState(STATE.DEFAULT)
+    }, 1000);
+
+    return () => {
+      clearTimeout(buttonTextEffect);
+    }
+  }, [buttonState])
+
+  const sensorUsage = agpCGM?.data?.current?.stats?.sensorUsage?.sensorUsage || 0;
```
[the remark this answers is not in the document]
henry-tp: Ahh, I will double check this.

The good news is that this value here only impacts whether to conditional render the copy button; I am not displaying the value. In the displayed values, it is using `sensorUsageAGP`. So I think this is safe for now, but this should be fixed ASAP. I need to wrap my head around the condition for data sufficiency (see here: https://github.com/tidepool-org/viz/blob/0feca7c7e8b511f90e4abd0bdb9b8c3034b8958c/src/modules/print/utils/AGPUtils.js#L62-L82)

henry-tp:
Hey @clintonium-119 , following our huddle today:

- I have committed the "risk-free" changes to this PR - semicolons, comments, default switch case
a1d5375daac53f489521c767864e0a087f6abc7d
11f1fcb663464d4a5f38fbcdcf01bf0c37a9b5b4

- I will start a new PR to address your comments for this one.

app/pages/dashboard/PatientDrawer/useAgpCGM/getOpts.js
henry-tp: Will address in separate PR - I think I left this in when I initially thought we would be showing the BGM AGP as well

app/themes/colorPalette.js
```diff
@@ -72,6 +72,7 @@ export default {
       '#5A5A5A',
       '#505050',
       '#3E3E3E',
+      '#707070'
```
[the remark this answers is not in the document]
henry-tp: Will address in separate PR - makes sense

app/pages/dashboard/PatientDrawer/CGMStatistics.js
```diff
@@ -0,0 +1,126 @@
...
+
+  const dateRange  = formatDateRange(oldestDatum.time, newestDatum.time, timezoneName);
+  const daySpan    = endpointDays;
+  const cgmActive  = utils.roundToPrecision(sensorUsageAGP, 1);
+  const avgGlucose = utils.roundToPrecision(averageGlucose, avgGlucosePrecision);
+  const gmi        = utils.roundToPrecision(glucoseManagementIndicatorAGP, 1);
+  const cov        = utils.roundToPrecision(coefficientOfVariation, 1);
```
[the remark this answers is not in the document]
henry-tp: Will address in separate PR - sounds good. I need to dig into the code to make sure I use `formatDatum` properly

app/pages/dashboard/PatientDrawer/useAgpCGM/getQueries.js
henry-tp: Will address in separate PR - I think we can remove the BGM AGP, I had initially left it in as I was worried about breaking things, but I think it can be removed

clintonium-119 (approved):
Marking as approved, with the caveat that the feedback items that are unable to be addressed prior to release need to have Jira tasks created and assigned to the next release so that they don't become forgotten prior to this feature becoming generally available (currently, it is only accessible via a feature flag for targeted users).

app/pages/dashboard/PatientDrawer/MenuBar/CGMClipboardButton.js
```diff
@@ -0,0 +1,118 @@
...
+
+  const currentDate = moment().format('MMMM Do, YYYY');
+
+  const dateRange = formatDateRange(oldestDatum.time, newestDatum.time, timezoneName);
+
+  const targetRange  = bgUnits === MGDL_UNITS ? '70-180' : '3.9-10.0';
+  const lowRange     = bgUnits === MGDL_UNITS ? '54-70' : '3.0-3.9';
+  const veryLowRange = bgUnits === MGDL_UNITS ? '<54' : '<3.0';
+  
+  const countsInTarget  = utils.roundToPrecision((counts.target / counts.total) * 100, 0);
```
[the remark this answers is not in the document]
henry-tp: Will address in separate PR

PR #1491, opened 2025-01-02
WEB-3337 - Patient Navbar Uploader Fixes

henry-tp:


app/components/modaloverlay/modaloverlay.less
```diff
@@ -10,7 +10,7 @@
     left: 0;
     right: 0;
     bottom: 0;
-    z-index: 10;
+    z-index: 11;
```
henry-tp, line 13: So this is the issue at hand - the table row and modal overlay both have `z-index: 10`.

image

The z-index for the overlay is set by the `.ModalOverlay-target` class. My issue is that I can't seem to figure out where the z-index for the table row is set.

Increasing the z-index of the overlay in this way seems to be safe to me, but your input would be welcome on whether this is a safe adjustment.

app/components/navpatientheader/index.js
```diff
@@ -53,6 +54,7 @@ const NavPatientHeader = ({
   const handleUpload = () => {
     trackMetric('Clicked Navbar Upload Data');
     setIsUploadOverlayOpen(true);
+    launchCustomProtocol('tidepoolupload://open');
```
henry-tp, line 57: whoops

clintonium-119 (approved):
LTGM :100:

PR #1498, opened 2025-01-09
[WEB-3343, WEB-3344] - Navbar Demographics Not Updating Correctly

henry-tp:
[WEB-3343]
[WEB-3344]

WEB-3343 Description: When an account is claimed, the name,mrn,dob displayed does not update after editing when opened in the clinic patient list.

WEB-3344 Description: when clearing the mrn in the patient list, the mrn displayed when clicking on the patient row does not update to a blank value.

-----

Post-fix:

https://github.com/user-attachments/assets/a6af12a9-9468-46aa-bbcd-359bb41bb2aa

https://github.com/user-attachments/assets/394f8a9b-85f4-4189-bea6-fc4bd8ca0464

[WEB-3344]: https://tidepool.atlassian.net/browse/WEB-3344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[WEB-3343]: https://tidepool.atlassian.net/browse/WEB-3343?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/navpatientheader/index.js
+++ b/app/components/navpatientheader/index.js
@@ -31,7 +31,6 @@ const HeaderContainer = ({ children }) => (
 
 const NavPatientHeader = ({ 
   patient,
-  clinicPatient,
   user, 
   permsOfLoggedInUser,
   trackMetric,
@@ -42,7 +41,7 @@ const NavPatientHeader = ({
   const history = useHistory();
   const [isUploadOverlayOpen, setIsUploadOverlayOpen] = useState(false);
 
-  if (!patient?.profile) return null;
+  if (!patient?.profile?.patient) return null;
 
   const { patientListLink } = getPatientListLink(clinicFlowActive, selectedClinicId, query, patient.userid);
   const { canUpload, canShare } = getPermissions(patient, permsOfLoggedInUser);
@@ -80,7 +79,7 @@ const NavPatientHeader = ({
           ? <>
               <Back onClick={handleBack} />
               <Name patient={patient} />
-              <DemographicInfo patient={patient} clinicPatient={clinicPatient} />
+              <DemographicInfo patient={patient} />
               <ClinicianMenuOptions 
                 onViewData={handleViewData}
                 onViewProfile={handleViewProfile}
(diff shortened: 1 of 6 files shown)


app/pages/app/app.js
```diff
@@ -356,10 +357,12 @@ export class AppComponent extends React.Component {
     } = this.props;
 
     if (!this.isPatientVisibleInNavbar()) return null; // only show on pages with a patient of focus
+
+    const combinedPatient = personUtils.combinedAccountAndClinicPatient(patient, clinicPatient);
```
henry-tp, line 361: I used the same method here as in `PatientInfo` so, while there is some funkyness with how we interpret different data from `patient` vs `clinicPatient`, it should at least be consistent

app/components/navpatientheader/navPatientHeaderHelpers.js
```diff
@@ -17,5 +17,15 @@ export const getPatientListLink = (clinicFlowActive, selectedClinicId, query, pa
     patientListLink = `/dashboard/${query.dashboard}?drawerPatientId=${patientId}`;
   };
 
-  return { patientListLink }
+  return { patientListLink };
+}
+
+export const getDemographicInfo = (patient, clinicPatient) => {
+  const combinedPatient = personUtils.combinedAccountAndClinicPatient(patient, clinicPatient);
```
henry-tp, line 24: The issue was that the form updates `clinicPatient` instead of `patient`.

I used the same method here as in `<PatientInfo />`. So, while there is some funkyness with how we interpret differences in data values from `patient` vs `clinicPatient` when we merge them together, it should at least be consistent with other components in the UI

app/components/navpatientheader/index.js
```diff
@@ -30,7 +30,8 @@ const HeaderContainer = ({ children }) => (
 );
 
 const NavPatientHeader = ({ 
-  patient, 
+  patient,
+  clinicPatient,
```
henry-tp, line 34: I'm tempted to try to grab all of these vales from Redux here instead of passing them as props, as we discussed before. What do you think?

krystophv (approved):
LGTM, nice work! :shipit:

PR #1502, opened 2025-01-17
[WEB-3377] - "Last Reviewed" and "PatientDrawer" flags occassionally missing on TIDE Dashboard

henry-tp:
[WEB-3377]

[WEB-3377]: https://tidepool.atlassian.net/browse/WEB-3377?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

To reliably repro this error on master, a specific QA user on QA2 can be used. You can DM me on slack for their details. 

The issue was that because our fetch for flags is async, they are not always set to `true` on the first render. They are missing from the memo deps array, so they don't trigger a re-render. Adding them to the deps array ensures that a re-render is triggered when they are flipped to `true`.


clintonium-119 (approved):
Looks good.  Nice job debugging this one.  Approved :+1:

PR #1504, opened 2025-01-22
Web 3376 mobile analytics

henry-tp:


app/bootstrap.js
```diff
@@ -46,10 +47,23 @@ export let appContext = {
 };
 
 appContext.trackMetric = (...args) => {
-  let selectedClinicId = appContext.store?.getState()?.blip?.selectedClinicId;
-  if (selectedClinicId) {
-    _.defaultsDeep(args, [, { clinicId: selectedClinicId }]);
-  }
+  const state = appContext.store?.getState();
+
+  const clinicId       = state?.blip?.selectedClinicId || undefined;
```
henry-tp, line 52: My understanding of the original `if` statement was that it prevents `clinicId` from being set to `null` which would cause this issue of having an empty param for `clinicId` in the network request:

image

If that is correct, then setting it to `undefined` should have the same effect of preventing that issue

app/bootstrap.js
```diff
@@ -46,10 +47,24 @@ export let appContext = {
...
+  const loggedInUserId = state?.blip?.loggedInUserId;
+  const user           = state?.blip?.allUsersMap?.[loggedInUserId];
+
+  const clinician = user ? personUtils.isClinicianAccount(user) : undefined;
+  const mobile = utils.isMobile();
+
+  const eventMetadata = {
+    clinicId,
+    clinician,
+    mobile,
```
henry-tp, line 63: KissMetrics doesn't support casing so making this `isMobile` or `isClinician` turns it into `Ismobile` or `Isclinician` in KissMetric's UI. Maybe we could do snake_case?

henry-tp:
image

app/bootstrap.js
```diff
@@ -46,10 +47,24 @@ export let appContext = {
...
-  let selectedClinicId = appContext.store?.getState()?.blip?.selectedClinicId;
-  if (selectedClinicId) {
-    _.defaultsDeep(args, [, { clinicId: selectedClinicId }]);
-  }
+  const state = appContext.store?.getState();
+
+  // Empty values should be undefined, not null, to prevent sending blank query params
+  const clinicId       = state?.blip?.selectedClinicId || undefined;
+  const loggedInUserId = state?.blip?.loggedInUserId;
+  const user           = state?.blip?.allUsersMap?.[loggedInUserId];
```
henry-tp, line 55: by the way Chris, I'm a fan of doing this kind of alignment where possible, but if we're not a fan of this (or any of my other stylistic choices) feel free to tell me off 😋

app/bootstrap.js
```diff
@@ -46,10 +47,24 @@ export let appContext = {
 };
 
 appContext.trackMetric = (...args) => {
-  let selectedClinicId = appContext.store?.getState()?.blip?.selectedClinicId;
-  if (selectedClinicId) {
-    _.defaultsDeep(args, [, { clinicId: selectedClinicId }]);
-  }
```
henry-tp, line 52: My understanding around lodash and `_.defaultsDeep` is weak, but my interpretation is that the `if` condition is needed so that `clinicId` is `undefined` instead of `null`. Let me know if there is another idea behind the `if` statement that I missed.

app/bootstrap.js
krystophv: I think I'm not really actually a fan of this style. Probably prefer to match up with the rest of the codebase (I don't _think_ we do this anywhere)

app/bootstrap.js
krystophv: It was to ensure that there is no property on the resulting object that gets passed to the metric API call if there's not a current selectedClinicId. Having a property (e.g. something that can be gotten with `hasOwnProperty`) who's value is `undefined` is somewhat different than lacking the property entirely. By gating the `defaultsDeep` call, we don't add any properties with `undefined` values. 

I guess another strategy would be to trim the `undefined`s out of the eventMetadata before passing to the `defaultsDeep` call via `_.omitBy(eventMetadata, _.isUndefined)` to prevent passing `undefined` properties through.

app/bootstrap.js
henry-tp: No worries at all!

20026434655e7bcc13f42a2c6cc1d50f6e85b14b

app/bootstrap.js
henry-tp: That works! I can confirm that the undefined values are being successfully omitted out of the object.

2fa6d6d907d30cb9ed5d9d53336599ae630d9b7f

henry-tp: Hey actually, it looks like `_.isNil` captures both `undefined` and `null`, so we don't need to explicitly set empty values to `undefined` first (which is a little clunky IMO)

d91be19949f07b8d4bc656b4efe841527df441e0

app/bootstrap.js
```diff
@@ -46,10 +47,26 @@ export let appContext = {
...
+  const mobile = utils.isMobile();
+
+  let eventMetadata = {
+    clinicId,
+    clinician,
+    mobile,
+  };
+
+  // Empty values should be omitted from the metadata object to prevent sending blank query params
+  _.omitBy(eventMetadata, _.isNil);
```
krystophv, line 66: `_.omitBy` doesn't mutate the target object

henry-tp: Yikes, embarrassing. I had checked the output of `_.omitBy()` with `console.log()` but not that the actual variable `eventMetadata` variable after the change. Thanks for catching this

6f71e003

krystophv (approved):
LGTM 🚀

app/bootstrap.js
krystophv: The only reason I know to check is from having put those sorts of things into production myself :)

PR #1505, opened 2025-01-23
[WEB-3397] - Mobile Responsiveness V1

henry-tp:
[WEB-3397]

[WEB-3397]: https://tidepool.atlassian.net/browse/WEB-3397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/core/utils.js
```diff
@@ -73,7 +73,7 @@ utils.getIn = (obj, props, notFound) => {
 
 utils.isSupportedBrowser = () => {
   var userAgent = navigator.userAgent.toLowerCase();
-  return (userAgent.indexOf('chrome') > -1 && userAgent.indexOf('opr') === -1 && userAgent.indexOf('mobi') === -1);
+  return (userAgent.indexOf('chrome') > -1 && userAgent.indexOf('opr') === -1);
```
henry-tp, line 76: TODO: NTS need to triple check all of the places this method is used and what it impacts

app/components/navbar/index.js
```diff
@@ -13,4 +13,38 @@
...
+  }
+`;
+
+const Navbar = (props) => (
+  <>
+    <DesktopOnly>
+      <DesktopNavbar {...props} />
+    </DesktopOnly>
+    <MobileOnly>
+      <MobileNavbar {...props} />
```
henry-tp, line 45: NTS: remove

app/components/chart/basics.js
```diff
@@ -200,6 +200,12 @@ class Basics extends Component {
       return '';
     }
 
+    const endpointsRange = _.get(props, 'data.data.current.endpoints.range', []);
+
+    if (endpointsRange[0] <= 0 || endpointsRange[1] <= 0) {
+      return '';
+    }
```
henry-tp, line 207: When calling `patientData.setInitialChartView` for a second time, there is a split second where this takes a default value of `0`. It corrects itself immediately on the next render.

app/components/chart/header.js
```diff
@@ -121,8 +123,10 @@ const Header = withTranslation()(class Header extends Component {
       'patient-data-subnav-hidden': this.props.chartType === 'no-data',
     });
 
+    const canSelectDateRange = _.includes(['basics', 'daily'], this.props.chartType);
```
henry-tp, line 126: If you have insight on why only these two chart types are able to launch the chart range modal, I would appreciate being able to know why. Sharon and Bill were unsure as well ([discussion thread](https://tidepoolteam.slack.com/archives/C089XASA7C6/p1738093069585199)). I have duplicated this restriction for the mobile view.

app/components/mediaqueries/index.js
```diff
@@ -0,0 +1,47 @@
...
+ * @param {Number} [breakpoint] Viewport width (in px) above which any children will be hidden
+ */
+export const MobileOnly = styled.div`
+  display: none;
+  @media screen and (max-width: ${props => (props.breakpoint || DEFAULT_MOBILE_BREAKPOINT)}) {
+    display: block;
+  }
+`;
+
+export default { DesktopOnly, MobileOnly };
```
henry-tp, line 47: So I'm hoping this approach to conditional rendering is not too far-fetched for you. For the purposes of conditional rendering things based on screen width, I'm really not a big fan of using event listeners and javascript; I much prefer relying on CSS. As we're moving towards more of a CSS-in-JS approach, I think having a kind of base abstraction like this to handle conditional rendering mobile vs desktop with JSX makes sense. 

One weakness of this approach to keep in mind is that it doesn't actually stop the component from rendering, it just hides it from being visually shown, meaning that the child components will still render (AJAX requests will still fire, etc).

app/components/navbar/DesktopNavbar/DesktopNavbar.js
```diff
@@ -189,10 +180,6 @@ export default withTranslation()(class extends React.Component {
     return personUtils.fullName(this.props.user);
   };
 
-  isSamePersonUserAndPatient = () => {
-    return personUtils.isSame(this.props.user, this.props.patient);
-  };
```
henry-tp, line 194: There were no usages of either `getPatientLink` or `isSamePersonUserAndPatient` in this component. 

Removing this allows us to not require `props.patient` and `props.getUploadUrl` anymore which reduces the props needed to be passed from above.

app/components/navbar/DesktopNavbar/NavigationMenu.js
```diff
@@ -42,34 +43,41 @@ export const NavigationMenu = props => {
...
+  } = useNavigation(api, trackMetric);
+
   const popupState = usePopupState({
     variant: 'popover',
     popupId: 'navigationMenu',
   });
 
   const privateWorkspaceOption = {
-    action: handleSelectWorkspace.bind(null, null),
+    action: () => handleSelectWorkspace(null),
```
henry-tp, line 59: All existing click handlers moved to the `useNavigation` abstraction to prevent discrepancies between different navbars from coming up over time

app/core/navutils.js
```diff
@@ -0,0 +1,161 @@
+import React from 'react';
+import { useSelector, useDispatch } from 'react-redux';
+import { push } from 'connected-react-router';
+import _ from 'lodash';
+import personUtils from './personutils';
+import { useLocation } from 'react-router-dom';
+import launchCustomProtocol from 'custom-protocol-detection';
+import * as actions from '../redux/actions';
+
```
henry-tp, line 9: `getFinalSlug`, `getPermissions`, `getPatientListLink` and `getDemographicInfo` were all just moved here from a different file.

app/pages/app/app.js
```diff
@@ -314,23 +314,18 @@ export class AppComponent extends React.Component {
     if (!_.includes(LOGIN_NAV_ROUTES, this.props.location)) {
       if (this.props.authenticated ||
         (_.get(this.props.fetchingUser, 'inProgress') || this.props.fetchingPatient)) {
-        var patient, getUploadUrl;
-        if (this.isPatientVisibleInNavbar()) {
-          patient = this.props.patient;
-          getUploadUrl = this.props.context.api.getUploadUrl.bind(this.props.context.api);
-        }
```
henry-tp, line 321: The DesktopNavbar no longer requires `patient` or `getUploadUrl` as mentioned in [this comment](https://github.com/tidepool-org/blip/pull/1505/files#r1935000779). However, the MobileNavbar now requires patient information to show the current patient's name.

app/pages/patientdata/navbarcharttypesetter.js
```diff
@@ -0,0 +1,21 @@
+import React, { useEffect } from 'react';
+import _ from 'lodash';
+import { setNavbarChartType } from '../../redux/actions/sync';
+import { useDispatch } from 'react-redux';
+
+const NavbarChartTypeSetter = ({ chartType }) => {
```
henry-tp, line 6: For the record I really don't like having to do this, but React's one-way data binding makes it hard to pass data up the chain and I couldn't come up with a better way

app/pages/patientdata/patientdata.js
```diff
@@ -1739,6 +1758,11 @@ export const PatientDataClass = createReactClass({
           window.patientData = 'No patient data has been loaded yet. Run `window.loadPatientData()` to popuplate this.'
           window.loadPatientData = this.saveDataToDestination.bind(this, 'window');
           window.downloadPatientData = this.saveDataToDestination.bind(this, 'download');
+
+        // A new chart type can be requested by changing queryParams, in which case we need to
+        // fire setInitialChartView again to change the chart rendered
+        } else if (this.props.queryParams.chart !== nextProps.queryParams.chart) {
+          this.setInitialChartView(nextProps);
```
henry-tp, line 1765: My first consequential change in `patientdata.js` and I'm really scared about it. Would appreciate your detailed look to see whether something can go wrong here.

app/pages/patientdata/patientdata.js
```diff
@@ -1739,6 +1758,11 @@ export const PatientDataClass = createReactClass({
           window.patientData = 'No patient data has been loaded yet. Run `window.loadPatientData()` to popuplate this.'
           window.loadPatientData = this.saveDataToDestination.bind(this, 'window');
           window.downloadPatientData = this.saveDataToDestination.bind(this, 'download');
+
```
henry-tp, line 1761: My first consequential change in `patientdata.js` and I'm really scared about it. Would appreciate your detailed look to see whether something can go wrong here.

app/pages/patients/patients.js
```diff
@@ -308,10 +321,6 @@ export let Patients = withTranslation()(class extends React.Component {
   UNSAFE_componentWillMount() {
     this.props.dataWorkerRemoveDataRequest(null, this.props.currentPatientInViewId);
 
-    if (this.props.clearPatientInView) {
-      this.props.clearPatientInView();
-    }
```
henry-tp, line 313: Do you see any issues with this change? 

Bill and Sharon would like to preserve the patient context when the user navigates this page so that you can still keep the navigation options (see below).

https://github.com/user-attachments/assets/669c0074-e3d2-4a5e-b41d-85ad0d571fa8

I would prefer NOT to hold the patient context in the navbar's internal state while the patient context is cleared from the page's state.

I have poked around in as many ways I could think of but I couldn't come up with any issues that arise from removing this call to `clearPatientInView()`

app/routes.js
```diff
@@ -53,6 +53,43 @@ export const requireSupportedBrowser = (next, ...args) => (dispatch) => {
   }
 }
 
+/**
```
henry-tp, line 56: the `requireSupportedBrowser` method is now unused, but i would like to not delete it as we will need to revert back to it. Barring clinicians from accessing the platform in mobile is only a temporary measure.

app/pages/app/app.js
```diff
@@ -314,23 +314,18 @@ export class AppComponent extends React.Component {
...
+
         return (
           <Box className="App-navbar" variant="containers.large" bg="transparent" mb={0} py={2}>
             <Navbar
               user={this.props.user}
               fetchingUser={_.get(this.props.fetchingUser, 'inProgress')}
-              patient={patient}
+              patient={this.props.patient}
```
henry-tp, line 323: Here, I am allowing the Navbar to keep access to the `patient` object even when `isPatientVisibleInNavbar()` is false. I hope you have no issue with this

app/components/navbar/MobileNavbar/Menu.js
```diff
@@ -0,0 +1,203 @@
...
+  justify-content: flex-end;
+  border-top: 1px solid ${colorPalette.primary.gray10};
+
+  :first-child {
+    border-top: none;
+  }
+
+  .icon-custom-svg {
+    filter: brightness(0) saturate(100%) invert(13%) sepia(47%) saturate(861%) hue-rotate(216deg) brightness(93%) contrast(101%);
+  }
```
henry-tp, line 51: This is a hack. This filter property converts a the icon color to `purpleDark`, as the icon color is currently hardcoded in the SVG. I don't believe we've come into situations where we need the same icon but in a different platform.

I attempted a few things before getting here. I'm used to using `currentColor` to pass the color but that does not work with how we're currently rending SVGs in the `Icon` component. In the future we may want to look at a different way of rendering Icon SVGs.

app/components/navpatientheader/MenuOptions/Clinician.js
```diff
@@ -3,16 +3,17 @@ import { Box, Flex } from 'theme-ui';
...
 
 import Button from '../../elements/Button';
 import viewIcon from '../../../core/icons/viewIcon.svg'
 import profileIcon from '../../../core/icons/profileIcon.svg'
 import uploadIcon from '../../../core/icons/uploadIcon.svg'
 
-const ClinicianMenuOptions = ({ 
-  t, 
-  onViewProfile, 
+const ClinicianMenuOptions = ({
```
henry-tp, line 14: I found the configuration that enables automatic removal of trailing whitespace, sorry it took me a little to get this set up.

app/components/navbar/MobileNavbar/Menu.js
```diff
@@ -0,0 +1,203 @@
...
+  .MuiPopover-paper {
+    width: 100%;
+    padding: 0;
+  }
+`;
+
+const Menu = ({ api, trackMetric }) => {
+  const { t } = useTranslation();
+
+  const patient = useSelector(state => selectPatient(state));
```
henry-tp, line 67: If you're okay with this being the first instance of it,, I'm keen to start using this pattern and to start reducing prop drilling as we had discussed.

app/pages/patients/patients.js
clintonium-119: I suppose this is fine, but make sure that there's no way that we can run into stale data issues or network calls for the stored user when switching users - we don't want to accidentally show or fetch some data for alberto (if he's been stored in state) when switching to Cacilia

app/routes.js
clintonium-119: That's fine to leave it for now.

app/components/navbar/MobileNavbar/Menu.js
clintonium-119: Yeah, that's fine.

clintonium-119 (changes requested):
Great work on this.  I have a few suggestions/questions in my comments, but really well done :+1:

app/components/navbar/MobileNavbar/Title.js
```diff
@@ -0,0 +1,114 @@
...
+const TITLE_STATE = {
+  PRIVATE_WORKSPACE: 'PRIVATE_WORKSPACE',
+  ACCOUNT_SETTINGS: 'ACCOUNT_SETTINGS',
+  WELCOME: 'WELCOME',
+  DEVICES: 'DEVICES',
+  SUMMARY_STATS: 'SUMMARY_STATS',
+  SHARE: 'SHARE',
+};
+
+const isEmptyPatientData = (patient, data) => {
```
henry-tp, line 28: These two methods stolen from `patientdata.js:500`

test/unit/components/navbar/MobileNavbar/Menu.test.js
```diff
@@ -0,0 +1,138 @@
...
+      wrapper.find('Button#mobileNavbar_settingsChartButton').simulate('click');
+      expect(handleViewSettingsChart.calledOnce).to.be.true;
+    });
+
+    it('should navigate to share page button is pressed', () => {
+      wrapper.find('Button#mobileNavbar_shareButton').simulate('click');
+      expect(handleShare.calledOnce).to.be.true;
+    });
+
+    it('should navigate to private workspace action when button is pressed', () => {
```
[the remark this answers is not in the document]
henry-tp: Oops. Copypasta 😅 

f9a1d5ed

test/unit/components/navbar/MobileNavbar/Menu.test.js
```diff
@@ -0,0 +1,138 @@
...
+      wrapper.find('Button#mobileNavbar_workspaceButton').simulate('click');
+      expect(handleSelectWorkspace.calledOnceWithExactly(null)).to.be.true;
+    });
+
+    it('should navigate to private workspace action when button is pressed', () => {
```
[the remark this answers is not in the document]
henry-tp: f9a1d5ed

test/unit/components/navbar/MobileNavbar/Menu.test.js
```diff
@@ -0,0 +1,138 @@
...
+      wrapper.find('Button#mobileNavbar_accountSettingsButton').simulate('click');
+      expect(handleViewAccountSettings.calledOnce).to.be.true;
+    });
+
+    it('should navigate to private workspace action when button is pressed', () => {
```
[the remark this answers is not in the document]
henry-tp: f9a1d5ed

app/components/navbar/MobileNavbar/Menu.js
```diff
@@ -0,0 +1,203 @@
...
+              {label}
+            </Button>
+          ))}
+        </Box>
+      </StyledPopover>
+    </Box>
+  );
+};
+
+export default Menu;
```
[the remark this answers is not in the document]
henry-tp: Ahh yikes, have not noticed. I have turned this on in my editor, hopefully shouldn't see it anymore: 

cb40b15884a49dbc199fa372744a42baaf48c58a

app/components/mediaqueries/index.js
```diff
@@ -0,0 +1,50 @@
...
+ * the terms of the associated License, which is identical to the BSD 2-Clause
+ * License as published by the Open Source Initiative at opensource.org.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the License for more details.
+ *
+ * You should have received a copy of the License along with this program; if
+ * not, you can obtain one from Tidepool Project at tidepool.org.
+ */
```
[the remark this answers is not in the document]
henry-tp: Gotcha, makes sense

7e56073e1763451caa4c2ac56adf04fdd9bf6f24

app/components/navbar/MobileNavbar/Menu.js
henry-tp: Sounds good 

7e56073e1763451caa4c2ac56adf04fdd9bf6f24

app/components/navbar/MobileNavbar/Title.js
```diff
@@ -0,0 +1,98 @@
...
+  if (finalSlug === '/data' && chartType === 'settings' && hasDataSources) return TITLE_STATE.WELCOME;
+  if (finalSlug === '/data' && chartType === 'settings') return TITLE_STATE.DEVICES;
+  if (finalSlug === '/data') return TITLE_STATE.SUMMARY_STATS;
+
+  return TITLE_STATE.DEFAULT;
+};
+
+const StyledTitle = styled(Flex)`
+  .icon-custom-svg {
+    filter: brightness(0) saturate(100%) invert(13%) sepia(47%) saturate(861%) hue-rotate(216deg) brightness(93%) contrast(101%);
```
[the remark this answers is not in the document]
henry-tp: 7e56073e1763451caa4c2ac56adf04fdd9bf6f24

app/pages/patients/patients.js
henry-tp: Dang, really good foresight on this. I dug through the code on this and once I was able to wrap my head around some of it, I was able to actually generate a potentially dangerous situation.

What I did not realize was that, when fetching patient data, we actually rely on the `fetchPatientSuccess` action to replace the current `currentPatientInViewId` in redux with a new one. Without the "Private Workspace" page clearing out the `currentPatientInViewId`, it is set to the previous patient's `id` until a `fetchPatientSuccess` is called.

IMO this is not good because we are relying on a visit to the Workspace page to clear out the `currentPatientInViewId`. In the future, if we build a different UI to navigate between patients, the opportunity for this to create a dangerous result is still there.

What I think makes the most sense is to clear out the `currentPatientInViewId` the moment the user requests a new patient. IMO this is the safest way to ensure that any request to fetch the patient results in the current patient not being "active" anymore. What do you think about this approach?

3aa75b8f

henry-tp:
Thanks for the review @clintonium-119 . After your review, there are two remaining things I could use your help on:

1. How to replace the `setInitialChartView()` call and make it work for to derive the needed chart
2. Decision on the best approach to maintain patient context in the nav bar without risking pulling the wrong patient's data

If you have time to pair on this, that would be much appreciated, or we can figure it out however else

clintonium-119 (changes requested):

app/pages/patientdata/patientdata.js
henry-tp: Resolved above in slack discussion

app/components/navbar/MobileNavbar/Menu.js
```diff
@@ -0,0 +1,212 @@
...
+    handleEvent();
+    popupState.close();
+  };
+
+  const { name: patientName } = getDemographicInfo(patient, clinicPatient);
+  const { canShare } = getPermissions(patient, permsOfLoggedInUser);
+
+  const isClinicContext = !!selectedClinicId;
+  const isUserPatient = personUtils.isSame(user, patient);
+  const canViewDevices = isClinicContext || isUserPatient;
```
henry-tp, line 97: @clintonium-119 

I have implemented the changes from our video discussion, first about calling `setInitialChartView` and second to ensure that the Private Workspace still clears the `currentPatientInViewId`. 

Last consequential change I have made is this restriction on who sees a button to view the Device Settings. The `isClinicContext || isUserPatient` condition is a mirror of yours used to render the "Connect a Device" card.

clintonium-119 (approved):
Great job!  LGTM :shipit:

PR #1509, opened 2025-02-04
WEB-2474 - Workspace switches clinic context unexpectedly

henry-tp:


app/pages/patientdata/patientdata.js
```diff
@@ -2210,21 +2210,23 @@ export function getFetchers(dispatchProps, ownProps, stateProps, api, options) {
   // if is clinician user viewing a patient's data with no selected clinic
   // we need to check clinics for patient and then select the relevant clinic
 
-  let clinicToSelect = null;
+  const clinicsWithPatient = [];
+
   _.forEach(stateProps.clinics, (clinic, clinicId) => {
     let patient = _.get(clinic.patients, ownProps.match.params.id, null);
     if (patient) {
-      clinicToSelect = clinicId;
```
henry-tp, line 2217: The main issue is that this does not account for the possibility that a patient could be in multiple clinics of `stateProps.clinics`.

app/pages/patientdata/patientdata.js
```diff
@@ -2210,21 +2210,23 @@ export function getFetchers(dispatchProps, ownProps, stateProps, api, options) {
...
+      clinicsWithPatient.push(clinicId);
     }
   });
 
   if (
     personUtils.isClinicianAccount(stateProps.user) &&
     stateProps.user.userid !== ownProps.match.params.id &&
-    (!stateProps.selectedClinicId || stateProps.selectedClinicId !== clinicToSelect) &&
+    (!stateProps.selectedClinicId || !clinicsWithPatient.includes(stateProps.selectedClinicId)) &&
```
henry-tp, line 2225: The main issue is that the previous condition did account for the possibility that a patient could be in multiple clinics of `stateProps.clinics`. The `stateProps.selectedClinicId !== clinicToSelect` part of the condition meant tha there was only one valid clinic.

Now, if the patient is in _any_ of the `stateProps.clinics`, we don't trigger a clinic auto-selection because we know that every one of the clinics is valid.

app/pages/patientdata/patientdata.js
```diff
@@ -2210,21 +2210,23 @@ export function getFetchers(dispatchProps, ownProps, stateProps, api, options) {
...
     !stateProps.fetchingPatientFromClinic.inProgress
   ) {
-    if (clinicToSelect) {
+    if (clinicsWithPatient.length > 0) {
+      const clinicToSelect = clinicsWithPatient[0];
```
henry-tp, line 2229: So this is arbitrary for now, but there is a further edge case that, given:

- Clinician and Patient are both in 2 different clinics
- Clinician arrives at `patientdata` with no `selectedClinicId` in redux (e.g. arriving from an external hyperlink via their EHR, email, etc)

We don't have enough context in that situation to know "from which clinic" the clinician wants to view the patient.

app/pages/patientdata/patientdata.js
henry-tp: I'd ask you to help me triple check if there are any edge cases you can think of. Based on my interpretation of the code I can't really think of any, as I am only restricting the situations in which we call the `selectClinic` or the api refetch.

krystophv: Nah, I think you're good here 👍🏼.

Would maybe like a unit test or unit test update for this though?

app/pages/patientdata/patientdata.js
krystophv: This isn't any worse than the current behavior, so I don't see any harm in it. If we discover it's a recurring issue in practice, then we can try to solve it more holistically then.

app/pages/patientdata/patientdata.js
henry-tp: Oops, sorry. I've added a test case to ensure that `selectClinic()` does not get called in this case: e5afd12fbb7d431bb436dd7d2a8c723c99baddfe

The test above it seems to handle the positive case alright, so with this test handling the negative case I think iit should be okay.

Also removed the accidental `it.only()` 😅 : 3ee6d99315de7c71b5f9162308cb6eff86975851

krystophv (approved):
Nice work, LGTM :shipit:

PR #1513, opened 2025-02-11
[WEB-3418] - Enable Safari on Mobile

henry-tp:
[WEB-3418]

[WEB-3418]: https://tidepool.atlassian.net/browse/WEB-3418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

- Awaiting partner integration before merge


test/unit/utils/utils.test.js
```diff
@@ -72,6 +72,98 @@ describe('utils', () => {
...
+
+    edgeWin: 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/133.0.0.0 Safari/537.36 Edg/131.0.2903.86',
+    edgeMac: 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/133.0.0.0 Safari/537.36 Edg/131.0.2903.86',
+    edgeAndroid: 'Mozilla/5.0 (Linux; Android 10; HD1913) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/133.0.6943.49 Mobile Safari/537.36 EdgA/131.0.2903.87',
+    edgeIPhone: 'Mozilla/5.0 (iPhone; CPU iPhone OS 17_7_2 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/18.0 EdgiOS/131.2903.92 Mobile/15E148 Safari/605.1.15',
+
+    safariMac: 'Mozilla/5.0 (Macintosh; Intel Mac OS X 14_7_3) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/18.3 Safari/605.1.15',
+    safariIPhone: 'Mozilla/5.0 (iPhone; CPU iPhone OS 17_7_2 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/18.3 Mobile/15E148 Safari/604.1',
+    safariIPad: 'Mozilla/5.0 (iPad; CPU OS 17_7_2 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/18.3 Mobile/15E148 Safari/604.1',
+  };
```
henry-tp, line 95: I'm not sure about the value / longevity of these tests given that they capture only a specific device. But I feel like we should have something.

test/unit/bootstrap.test.js
```diff
@@ -12,7 +12,7 @@ import appContext from '../../app/bootstrap';
 
 describe('appContext', () => {
   before(() => {
-    Object.defineProperty(window.navigator, 'userAgent', { value: 'Mozilla/5.0 .. truncated .. Chrome/131.0.0.0' });
+    Object.defineProperty(window.navigator, 'userAgent', { value: 'Mozilla/5.0 .. truncated .. Chrome/131.0.0.0', configurable: true });
```
henry-tp, line 15: I found that if `configurable` is not set to true, `window.navigator` is prevented from being re-defined by `Object.defineProperty()`. Unfortunately this seems to bleed into other tests as well.

test/unit/bootstrap.test.js
```diff
@@ -12,7 +12,8 @@ import appContext from '../../app/bootstrap';
 
 describe('appContext', () => {
   before(() => {
-    Object.defineProperty(window.navigator, 'userAgent', { value: 'Mozilla/5.0 .. truncated .. Chrome/131.0.0.0' });
+    const userAgent = 'Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/133.0.0.0 Safari/537.36';
+    Object.defineProperty(window.navigator, 'userAgent', { value: userAgent, configurable: true });
```
henry-tp, line 16: They key change here is `configurable: true`. I found that if `configurable` is not set to true, `Object.defineProperty()` cannot be called again with `window.navigator`, so setting `configurable: true` allows us to re-define the property.

Unfortunately, `defineProperty` is called in one file, the effect seems to bleed into other test files as well, as the test runner is not restarting for every new file. That's why I needed to make an adjustment here.

krystophv (approved):
Nice work, LGTM 🚢

PR #1519, opened 2025-02-20
WEB-3442 - Change Routing

henry-tp:
WEB-3442

This PR addresses the bug noted in the ticket. Clicking on dates/datums in Basics, Trends, and BgLog will now redirect to the correct date when you land on the Daily chart.

It also changes the routing scheme so that we can navigate between charts by setting the URL to `/data/trends`, `/data/daily`, etc. 

The changes to the routing scheme handle both a hard refresh (e.g. user punches in `/data/trends` in their browser manually) or an AJAX update to the URL (e.g. if a component elsewhere on the page updates the URL to `/data/trends`, `<PatientData />` will react to the URL change).


app/pages/patientdata/patientdata.js
```diff
@@ -1769,11 +1765,6 @@ export const PatientDataClass = createReactClass({
           window.patientData = 'No patient data has been loaded yet. Run `window.loadPatientData()` to popuplate this.'
           window.loadPatientData = this.saveDataToDestination.bind(this, 'window');
           window.downloadPatientData = this.saveDataToDestination.bind(this, 'download');
-
-        // A new chart type can be requested by changing queryParams, in which case we need to
-        // fire setInitialChartView again to change the chart rendered
-        } else if (this.props.queryParams.chart !== nextProps.queryParams.chart) {
-          this.setInitialChartView(nextProps);
```
henry-tp, line 1776: Now thankfully we don't need this anymore

app/pages/patientdata/patientdata.js
```diff
@@ -1658,15 +1663,6 @@ export const PatientDataClass = createReactClass({
...
 
     this.setState(state, cb);
-
-    // Update chart query param to match current chart
-    const { search, pathname } = this.props.location;
-    const params = new URLSearchParams(search);
-
-    if (params.get('chart') !== chartType) {
-      params.set('chart', chartType);
-      this.props.history.push({ pathname, search: params.toString() });
```
henry-tp, line 1668: This was the issue causing the bug. Broadly speaking, this `history.push()` would cause us to reset the `datetimeLocation`.

app/pages/patientdata/patientdata.js
```diff
@@ -2072,13 +2135,32 @@ export const PatientDataClass = createReactClass({
...
     const bgSource = this.getMetaData('bgSources.current');
     const excludedDevices = this.getMetaData('excludedDevices', undefined, props);
+    let defaultChartTypeForPatient = null;
 
     if (uploads && latestDatum) {
-      // Allow overriding the default chart type via a query param (helps for development);
-      const chartType = this.state.refreshChartType || _.get(
-        props, 'queryParams.chart',
-        this.deriveChartTypeFromLatestData(latestDatum, uploads)
-      );
```
henry-tp, line 2081: `queryParams.chart` should not be used anymore, so no point in this `_.get()` call

app/pages/patientdata/patientdata.js
```diff
@@ -2072,13 +2144,38 @@ export const PatientDataClass = createReactClass({
     const latestDatum = _.last(_.sortBy(_.values(_.get(props.data, 'metaData.latestDatumByType')), ['normalTime']));
     const bgSource = this.getMetaData('bgSources.current');
     const excludedDevices = this.getMetaData('excludedDevices', undefined, props);
+    const pathname = props.location?.pathname;
+
+    let defaultChartTypeForPatient = null;
 
     if (uploads && latestDatum) {
-      // Allow overriding the default chart type via a query param (helps for development);
-      const chartType = this.state.refreshChartType || _.get(
```
henry-tp, line 2078: `queryParams.chart` is not used anymore, so removing the `_.get()` from this call.

app/routes.js
```diff
@@ -459,7 +459,7 @@ export const getRoutes = (appContext) => {
           <Route exact path='/patients/:id/profile' render={routeProps => (<Gate onEnter={boundRequireSupportedBrowserForUserType} key={routeProps.match.path}><PatientProfile {...routeProps} {...props} /></Gate>)} />
           <Route exact path='/patients/:id/share' render={routeProps => (<Gate onEnter={boundRequireAuth} key={routeProps.match.path}><AccessManagement {...routeProps} {...props} /></Gate>)} />
           <Route exact path='/patients/:id/share/invite' render={routeProps => (<Gate onEnter={boundRequireAuth} key={routeProps.match.path}><ShareInvite {...routeProps} {...props} /></Gate>)} />
-          <Route exact path='/patients/:id/data' render={routeProps => (<Gate onEnter={boundRequireSupportedBrowserForUserType} key={routeProps.match.path}><PatientData {...routeProps} {...props} /></Gate>)} />
+          <Route path='/patients/:id/data' render={routeProps => (<Gate onEnter={boundRequireSupportedBrowserForUserType} key={routeProps.match.path}><PatientData {...routeProps} {...props} /></Gate>)} />
```
henry-tp, line 462: So switching the matching from `<Route exact path='' />` to `<Route path='' />` here is a significant detail I'd like to discuss.

Another way I could do this is to pass an array in the following manner:

```jsx
          <Route exact path={[
            '/patients/:id/data',
            '/patients/:id/data/bgLog',
            '/patients/:id/data/basics',
            // ... daily, bgLog, trends
          ]} {...rest} />
```

henry-tp:
@clintonium-119 not sure if it makes more sense here to target the `release` or the `develop` with this PR. It is a critical bugfix and functionality is broken in `release` right now. However, it does also require extensive testing due to the routing changes. It made sense to put the routing changes together in this PR given that the bug was a routing-related issue.

What makes more sense to you?

app/routes.js
clintonium-119: Yeah, we can't have it re-downloading data each time we switch data views.

Did you try adding a path param to handle the view?  This is what I was expecting to see.  Something like:

```
<Route exact path='/patients/:id/data/:chart?' />
```

In this case, the `:chart` param is optional, and you can default it as makes sense in `patientdata.js` based on the data as per usual, and get the requested chart via useParams.  I'm not sure if it would cause remounting -- I'd have to try it out

clintonium-119 (changes requested):
Looks good.  Great job on this.  I did have a question about the route path in `routes.js` to consider prior to approval

app/routes.js
henry-tp: Man, that is a great suggestion, I didn't know why I didn't think of that. I verified that it does NOT cause remounting and therefore no data re-fetch. 

Given that `<PatientData />` is still a class component we can't use `useParams()`, but if I'm not mistaken this will enable functional components down the tree to access the chartType`useParams()`, so it's still a positive change to make.

What do you think of the changes in this commit? d160b244efec10dcd78115088a395a2b8aa63361

clintonium-119:
> @clintonium-119 not sure if it makes more sense here to target the `release` or the `develop` with this PR. It is a critical bugfix and functionality is broken in `release` right now. However, it does also require extensive testing due to the routing changes, so I think it makes more sense to target `develop`. It made sense to put the routing changes together in this PR given that the bug was a routing-related issue.
> 
> What makes more sense to you?

I think this is fine to have the base as develop, but I also think it needs to be merged to the release candidate - it's fine that it will take significant QA.

PR #1522, opened 2025-02-25
[WEB-3448] Banner logic tweaks

clintonium-119:
[WEB-3448]

This is basically providing a couple of important updates:

1. We need a unique `banner.interactionId` for each banner to ensure that, for instance, dismissing a dexcom-specific banner doesn't cause it to not show for abbott down the road, since prior to this, we were just using the bannerId in the interaction keys that were being stored in user prefs, or in session state for clinician users.

2. We need to store the provider name of a just connected banner source in redux state, since we can't reliably determine this the way I'd originally thought, which was supposing that the data source would have an empty `lastImportTime`.  In the case of Dexcom, this is only true for a few milliseconds, and so we weren't able to show the `dataSourceJustConnected` banner.

I also made a couple of small tweaks not included in the Jira ticket that I noticed during development:

1. We should re-fetch patient data sources after a user authorizes a connection, so if it's done via the data source error banner, and the user then navigates somewhere else and sees the data connections list, it will be up to date.

2. Fixed an issue with the navUtils test that was causing the tests to sporadically fail.  Did this by stubbing out `launchCustomProtocol` directly in the place it's called.

[WEB-3448]: https://tidepool.atlassian.net/browse/WEB-3448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/components/datasources/useProviderConnectionPopup.js
```diff
@@ -14,6 +16,10 @@ const useProviderConnectionPopup = ({ popupWatchTimeout = 500 } = {}) => {
   const { set: setToast } = useToasts();
   const [providerConnectionPopup, setProviderConnectionPopup] = useState(null);
   const authorizedDataSource = useSelector(state => state.blip.authorizedDataSource);
+  const justConnectedDataSourceProviderName = useSelector(state => state.blip.justConnectedDataSourceProviderName);
+  const { fetchingDataSources } = useSelector(state => state.blip.working);
```
henry-tp, line 20: Non-blocking: I see that we can get a significant reduction in re-render count by selecting more specifically. To load the patientdata route, I counted 31 renders with destructuring and 25 renders with specific selection, so this may really add up over time. I.e.

```jsx
fetchingDataSources = useSelector(state => state.blip.working.fetchingDataSources)
```

test/unit/utils/navutils.test.js
```diff
@@ -36,7 +36,7 @@ describe('navutils', () => {
 
     const api = { user: { logout: sinon.stub() } };
     const trackMetric = sinon.stub();
-    const launchCustomProtocolStub = sinon.stub(uploadUtils, 'launchCustomProtocol');
+    const launchCustomProtocolStub = sinon.stub();
```
henry-tp, line 39: Comment, no action needed: I'm really surprised and disappointed the original version was flaky

henry-tp (approved):
LGTM, just a few comments but nothing blocking :)

app/components/datasources/useProviderConnectionPopup.js
clintonium-119: This is interesting, and not something I would have expected.  I'll make this change.

test/unit/utils/navutils.test.js
clintonium-119: Curious - was this one of the tests that was causing trouble for you that you'd mentioned wanting to skip recently?

app/components/datasources/useProviderConnectionPopup.js
```diff
@@ -86,6 +98,12 @@ const useProviderConnectionPopup = ({ popupWatchTimeout = 500 } = {}) => {
             message: toastMessages[status],
             variant: toastVariants[status],
           });
+
+          if (status === 'authorized') {
+            dispatch(sync.setJustConnectedDataSourceProviderName(provider?.dataSourceFilter?.providerName));
+          }
```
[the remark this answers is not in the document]
clintonium-119: Good call.  I like that.  Will do.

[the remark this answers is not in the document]
clintonium-119: I had actually initially done this with the thought of trying to avoid adding it to redux, and was going to persist and expose it from tjhis hook which was my original reason for putting it in local state, but decided redux just made more sense, and should have gone back and cleaned this up as you're suggesting.

test/unit/utils/navutils.test.js
henryliang2: No - one of them was an RPM test and one of them was a tier 0300 clinic test. Both of them were resolved with the increased timeout, however. I havent experienced this launchcustomprotocol one failing yet.

clintonium-119: Got it.  I noticed it fail on Travis, and then it happened a few times locally for me.  Hopefully by the end of the year, we have all these intermittent test issues laid to rest with enzyme/mocha

PR #1524, opened 2025-02-28
[WEB-3457]Ensure that we only show toast messages in response to working state updates caused by banner interaction

clintonium-119:
[WEB-3457]

Basically what it says, but I also added some additional for when to consider an action async result to be complete by adding a check against the previous monitored working state.  This extra safeguard is following the pattern we've used in `handleAsyncResult` in a number of other components to prevent extra firing of toast messages

[WEB-3457]: https://tidepool.atlassian.net/browse/WEB-3457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


henry-tp (approved):
Looks good, thank you Clint

PR #1529, opened 2025-03-05
[WEB-3465] Set default chart type based only off of diabetes datums

clintonium-119:
Addresses [WEB-3465]

The issue was that Abbott sends a `cgmSettings` datum that is more recent than the latest `cbg` datum, causing it to default to the Basics view instead of the expected Trends.  I've pared down the list of datum types sent to `deriveChartTypeFromLatestData` to only those that are used to set the default view.

I've also pulled in this viz PR - which could have been handled separately, but it' one less blip PR this way (and I initally thought I'd be fixing this one by adding a device ID in the same section of viz code before I realized the actual issue):
https://github.com/tidepool-org/viz/pull/440

[WEB-3465]: https://tidepool.atlassian.net/browse/WEB-3465?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/pages/patientdata/patientdata.js
+++ b/app/pages/patientdata/patientdata.js
@@ -2147,16 +2147,17 @@ export const PatientDataClass = createReactClass({
   setInitialChartView: function(props = this.props) {
     // Determine default chart type and date from latest data
     const uploads = _.get(props.data, 'data.current.data.upload', []);
-    const latestDatum = _.last(_.sortBy(_.values(_.get(props.data, 'metaData.latestDatumByType')), ['normalTime']));
+    const latestDiabetesDatums = _.filter(_.values(_.get(props.data, 'metaData.latestDatumByType')), d => _.includes(['cbg', 'smbg', 'bolus', 'basal', 'wizard'], d.type));
+    const latestDiabetesDatum = _.last(_.sortBy(latestDiabetesDatums, ['normalTime']));
     const bgSource = this.getMetaData('bgSources.current');
     const excludedDevices = this.getMetaData('excludedDevices', undefined, props);
     const chartTypeFromPath = props.match?.params?.chartType;
 
     let defaultChartTypeForPatient = null;
 
-    if (uploads && latestDatum) {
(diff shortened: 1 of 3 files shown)


henry-tp (approved):
Looks good, thanks Clint!

app/pages/patientdata/patientdata.js
```diff
@@ -2091,18 +2091,18 @@ export const PatientDataClass = createReactClass({
...
       // Ideally, we determine the default view based on the device type
       // so that, for instance, if the latest data type is cgm, but comes from
       // an insulin-pump, we still direct them to the basics view
       const deviceMap = _.keyBy(uploads, 'deviceId');
-      const latestDataDevice = deviceMap[latestData.deviceId];
+      const latestDataDevice = deviceMap[latestDatum.deviceId];
 
       if (latestDataDevice) {
-        const tags = deviceMap[latestData.deviceId].deviceTags;
+        const tags = deviceMap[latestDatum.deviceId].deviceTags;
```
henry-tp, line 2105: I take it here that if `cgmSettings` was the latestDatum then it would have `'cgm'` in the deviceTags

clintonium-119: Actually, not necessarily, and that was the issue causing Abbott-cloud only accounts to fail:

The latest datum was a `cgmSettings`, but because our continuous data set uploads (Dexcom api, LibreView) don't get deviceId's, we weren't getting a match here, and because `cgmSettings` isn't in the list of fallback datum types below, it wouldn't get set to Trends there either.

Now, if, for instance, we had a `cgmSettings` with a deviceID that somehow matched one of our devices by `deviceId` (though, since I think `cgmSettings` is currently only provided by Loop, that is perhaps not possible), then it would default to to the trends view if there was a `cgm` deviceTag, but no `insulin-pump` device tag (there are devices with this combination of tags)

henry-tp (approved):

PR #1532, opened 2025-03-07
WEB-3436 - Modify Settings Dropdown Date Range

henry-tp:
[WEB-3436]

[WEB-3436]: https://tidepool.atlassian.net/browse/WEB-3436?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Part of me wonders if this small UI change is worth all of the additional bloat this causes? I tried to abstract out the bloat but it's still a little much with the API call.


app/components/chart/settings.js
```diff
@@ -60,6 +61,21 @@ function formatDuration(milliseconds) {
...
+  const relevantDatums = datums.filter(datum => {
+    // Filter out loop upload datum ('type' == upload && dataSetType == 'continuous') because it
+    // will have its date defaulted to now() and does not reflect the latest upload time
+    if (datum.type === 'upload' && datum.dataSetType === 'continuous') return false;
+
+    return datum.uploadId === uploadId;
+  });
+
+  return relevantDatums.length ? _.maxBy(relevantDatums, 'normalTime') : null;
+};
```
henry-tp, line 77: - I am making an assumption here that the `uploadId` is how we can find any datums that were batched and uploaded together with a specific `pumpSettings` datum.

- I'm not sure if there are any gotchas other than the Loop Upload datum I should be watching out for

app/components/chart/settings.js
```diff
@@ -60,6 +62,36 @@ function formatDuration(milliseconds) {
...
+
+    if (!uploadId) return;
+
+    const fetchOpts = {
+      uploadId,
+      type: ['cbg', 'smbg', 'basal', 'bolus', 'wizard', 'food', 'pumpSettings', 'upload'].join(','),
+      latest: 1,
+    };
+
+    tidepoolApi.patientData.get(patientId, fetchOpts, (err, latestDatums) => {
```
henry-tp, line 80: Do we have any issue with calling the API in this barebones fashion? From what I can see we don't really do so elsewhere. I didn't want to call `fetchPatientData` because it seems, as a side effect, to touch a lot of properties in Redux that I don't want to affect just to make this little API call. But let me know if there is an alternative preferred option.

app/components/chart/settings.js
```diff
@@ -142,7 +186,19 @@ const Settings = ({
...
       group.forEach((obj, index) => {
-        const previousObj = index > 0 ? group[index - 1] : { normalTime: new Date().getTime() };
+        let previousObj = group[index - 1];
+
+        // For the latest setting option, there is no end date, but we need to display when those pumpSettings
+        // are known to be valid until. We instead use the latest timestamp of ANY datum within the same upload,
+        // assuming that those pumpSettings were at least valid until the latest timestamp of any other data
+        // in the upload it was batched together with.
+        if (index === 0) {
+          setLatestUploadId(obj.uploadId); // triggers a fetch for the latest datums of this upload id
```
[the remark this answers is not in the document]
krystophv: This seems to be an off-base suggestion. The update, while technically in an iteration, is gated by strict equality on the index and is in a useEffect, so it could only happen once. Gonna just hide this one.

app/components/chart/settings.js
krystophv: Given the alternative that I could see being a new branch of the Redux state tree that just maps uploadIds and their latest diabetes datum (which would entail all the boilerplate async actions/reducer/etc) for state that there's not a clear use case for outside of this component, I'm pretty alright having this localized here.

app/components/chart/settings.js
krystophv: `uploadId` is the best candidate to find other data associated with `pumpSettings` at time of upload (or ongoing, in the case of continuous datasets, which will still validly represent the latest `pumpSettings` at the most recent datapoint).

I can't think of any other gotchas off the top of my head.

krystophv (approved):
LGTM 🚀

PR #1535, opened 2025-03-11
[WEB-2367] - New ClinicPatients UI bar for Patient Count and Reset Filters

henry-tp:
[WEB-2367]

Must be on QA1 to use the new API response payload

[WEB-2367]: https://tidepool.atlassian.net/browse/WEB-2367?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/pages/clinicworkspace/ClinicPatients.js
```diff
@@ -3267,9 +3268,24 @@ export const ClinicPatients = (props) => {
...
         <Loader show={loading} overlay={true} />
+
+        <Box
+          sx={{ backgroundColor: colorPalette.primary.bluePrimary00 }}
+          px={2}
+          py={2}
+        >
+          {t('Showing {{ shown }} of {{ total }} patients', {
+            shown: clinic?.fetchedPatientCount,
+            total: clinic?.patientCount,
```
henry-tp, line 3285: NTS MUST RESOLVE: why is this off-by-one??

app/pages/clinicworkspace/ClinicPatients.js
```diff
@@ -3267,9 +3297,33 @@ export const ClinicPatients = (props) => {
...
+          sx={{
+            backgroundColor: colorPalette.primary.bluePrimary00,
+            justifyContent: 'space-between',
+          }}
+        >
+          <Text>
+            {t('Showing {{ shown }} of {{ total }} patients', {
+              shown: clinic?.fetchedPatientCount,
+              total: clinic?.patientCount,
+            })}
```
henry-tp, line 3317: NTS MUST RESOLVE: why is this off-by-one??

Screenshot 2025-03-11 at 15 13 39

app/pages/clinicworkspace/ClinicPatients.js
```diff
@@ -3261,15 +3302,50 @@ export const ClinicPatients = (props) => {
...
+            sx={{
+              backgroundColor: colorPalette.primary.bluePrimary00,
+              borderBottom: '1px solid #D1D6E1',
+              justifyContent: 'space-between',
+            }}
+          >
+            <Text sx={{ fontWeight: 'medium' }}>
+              {t('Showing {{ shown }} of {{ total }} patient accounts', {
+                shown: clinic?.fetchedPatientCount,
+                total: clinic?.patientCount,
```
henry-tp, line 3337: Is it possible for `clinic` to be non-existent? `fetchedPatientCount` is accessed as `clinic?.fetchedPatientCount` elsewhere in the file so not sure if this is something we should be concerned about.

app/pages/clinicworkspace/ClinicPatients.js
```diff
@@ -151,6 +152,71 @@ const editPatientDataConnections = (patient, setSelectedPatient, selectedClinicI
...
+      sx={{
+        backgroundColor: colorPalette.primary.bluePrimary00,
+        borderBottom: '1px solid #D1D6E1',
+        justifyContent: 'space-between',
+      }}
+    >
+      <Text sx={{ fontWeight: 'medium' }}>
+        {t('Showing {{ shown }} of {{ total }} patient accounts', {
+          shown: clinic?.fetchedPatientCount,
+          total: clinic?.patientCount,
```
henry-tp, line 211: Is it possible for `clinic` to be non-existent? `fetchedPatientCount` is accessed as `clinic?.fetchedPatientCount` elsewhere in the file so not sure if this is something we should be concerned about.

henry-tp: Note that in QA environments, there is a bit of an off-by-one:

image

This is due to the addition of `Jill Jellyfish`. My understanding is that this should not really be an issue?

krystophv: There's a demo account for production as well (James Jellyfish) which doesn't count against the clinic's patient count, but _will_ be present in fetched results (and thus the fetched patient count), so this is something we'll need to resolve. I'm not sure what's the best way to go about it though, given that we _are_ fetching and _showing_ one more patient than the clinic patient count.

henry-tp: Okay, I will think on this a bit and see if I can come up with a holistic solution to it 🤔

henry-tp: Okay, I'll think on this a bit and see if I can come up with a holistic solution to it 🤔

henry-tp: Okay @krystophv, I think I have found a rough solution based on this detail. On initial page load, before the user has any chance to set any filters, `fetchedPatientCount` will reflect the value of all users including the Jellyfishes. See here:

Screenshot 2025-03-26 at 15 28 48

Unfortunately, `fetchedPatientCount` decreases as filters are applied, so the idea is to cache the value of `fetchedPatientCount` from the first page load.

This is the fix I ended up with: 9da4fdcfd

Initially, I tried to store it in a react component (3c53db2f7ec026d6f361f4374e9649bc1cc7eee0) but I ran into issues with components mounting/unmounting a user filtered causing the value to be lost. I could work around that by "lifting" the state and drilling the value down through props, but eventually I had to lift it all the way to `ClinicWorkspace.js` and if we're drilling a value that far down, it makes sense to put it in Redux anyway.

P.S. I hate this

app/redux/reducers/misc.js
```diff
@@ -610,7 +610,15 @@ export const clinics = (state = initialState.clinics, action) => {
         return newSet;
       }, {});
       return update(state, {
-        [clinicId]: { $set: { ...state[clinicId], patients: newPatientSet, fetchedPatientCount: count, lastPatientFetchTime: moment.utc().valueOf() } },
+        [clinicId]: {
+          $set: {
+            ...state[clinicId],
+            patients: newPatientSet,
+            fetchedPatientCount: count,
+            initialFetchedPatientCount: state[clinicId].initialFetchedPatientCount || count, // cache the first ever value
```
henry-tp, line 618: Okay @krystophv, I think I have found a rough solution based on this detail: On initial page load, before the user has any chance to set any filters, `fetchedPatientCount` will reflect the value of all users including the Jellyfishes. See here:

Screenshot 2025-03-26 at 15 28 48

Unfortunately, `fetchedPatientCount` decreases as filters are applied, so the idea is to cache the value of `fetchedPatientCount` from the first page load.

This is the fix I ended up with: 9da4fdcfd

Initially, I tried to store it in a react component (3c53db2f7ec026d6f361f4374e9649bc1cc7eee0) but I ran into issues with components mounting/unmounting a user filtered causing the value to be lost. I could work around that by "lifting" the state and drilling the value down through props, but eventually I had to lift it all the way to `ClinicWorkspace.js` and if we're drilling a value that far down, it makes sense to put it in Redux anyway.

P.S. I hate this

henry-tp: Nvm sorry Chris. I stumbled on some cases where this doesn't work (not even edge cases). Bringing it back to in progress to re-consider,

henry-tp: Okay @krystophv, I have to admit I'm a little stuck.

- We could make an additional API call to fetch the patient count reliably, but that seems a little bloaty and expensive just to address this issue of a fake patient. If it's not, we can go that route.

- Other hacks (e.g. `if (patientList.includes('james+jelly') patientCount += 1` don't really work because James + Jill Jellyfish are not guaranteed to be in the list.

- If the main concern is just that the numerator can be bigger than the denominator, I could just throw in a `_.max()`

Unsure what is best here 😞

henry-tp:
Hey @krystophv , I think this is good for another look :) Todd has added the data to the endpoint and deployed it to QA1.

test/unit/components/useProviderConnectionPopup.test.js
```diff
@@ -117,7 +117,7 @@ describe('useProviderConnectionPopup', function () {
       })).to.be.true;
 
       done();
-    }, 100);
+    }, 1000);
```
henry-tp, line 120: breaking locally for me without increasing the wait

krystophv (approved):
Updates LGTM 🚀

PR #1538, opened 2025-03-18
WEB-3494 - Hide specific banners when viewing on a mobile device

henry-tp:
[WEB-3494]

[WEB-3494]: https://tidepool.atlassian.net/browse/WEB-3494?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Admittedly I couldn't quite figure out how to generate these banners organically for testing the exact scenarios they come up, but I did test by modifying some variables and I'm quite certain everything here works as intended.

Also, my thought is that maybe we don't want to put in tests since it's only a temporary measure?

### DIFF
--- a/app/providers/AppBanner/AppBannerProvider.js
+++ b/app/providers/AppBanner/AppBannerProvider.js
@@ -36,7 +36,7 @@ const AppBannerProvider = ({ children }) => {
   const dispatch = useDispatch();
   const { pathname } = useLocation();
   const [formikContext, setFormikContext] = useState({});
-  const [isMobile] = useState(utils.isMobile());
+  const isMobile = useMemo(() => utils.isMobile(), []);
 
   const selectedClinicId = useSelector(state => state.blip.selectedClinicId);
   const clinics = useSelector(state => state.blip.clinics);
--- a/app/providers/AppBanner/AppBannerProvider.js
+++ b/app/providers/AppBanner/AppBannerProvider.js
@@ -36,6 +36,7 @@ const AppBannerProvider = ({ children }) => {
   const dispatch = useDispatch();
   const { pathname } = useLocation();
   const [formikContext, setFormikContext] = useState({});
+  const [isMobile] = useState(utils.isMobile());
 
   const selectedClinicId = useSelector(state => state.blip.selectedClinicId);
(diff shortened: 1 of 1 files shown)


app/providers/AppBanner/AppBannerProvider.js
```diff
@@ -98,8 +99,8 @@ const AppBannerProvider = ({ children }) => {
       bannerArgs: [dispatch, providers[erroredDataSource?.providerName], erroredDataSource],
     },
 
-    uploader: {
-      show: userIsCurrentPatient && dataSources?.length && !userHasPumpData,
+    uploader: { // Temporary: hide on mobile until we have a mobile-friendly profile page
+      show: !utils.isMobile() && userIsCurrentPatient && dataSources?.length && !userHasPumpData,
```
clintonium-119, line 103: Minor nit: Perhaps we should store the utils.isMobile() result in a variable to avoid calling it repeatedly.

clintonium-119 (changes requested):
Just a minor change request for avoiding excessive calls to `utils.isMobile`

app/providers/AppBanner/AppBannerProvider.js
henry-tp: Makes sense, yeah I had thought about that too.

I think `useState`, `useRef`, and `useMemo` all work here, although `useMemo` probably makes the most semantic sense. I ended up using `useState` for simplicity's sake, but let me know if you would prefer an alternative.

f37337c4622e49fe843bec0b1429aebb10dc164d

clintonium-119: I'd agree that `useMemo` is probably better here.  But we're splitting hairs.  Even a `const` would be fine, as we're probably prematurely optimizing here, but useMemo probably makes the most sense.

henry-tp: All good, makes sense 3ff783546d5dba7bf454c70e54035296f489c141

clintonium-119 (approved):
LGTM :fire:

PR #1548, opened 2025-04-01
Update pendoMiddleware.js to include missing subdomain of integration environment

wzeller:
I noticed that logging in to int was not registering in Pendo as the "INT" environment. It appears as if the "int-api" subdomain is missing from the environments defined in pendoMiddleware.js. Adding this URL should fix the issue.


clintonium-119 (approved):

clintonium-119:
Thanks, @wzeller .  Merged.

PR #1555, opened 2025-04-07
Fix missing provider when redirecting after successful Data Connection

henry-tp:
There is an issue with redirecting back on mobile

https://tidepoolteam.slack.com/archives/C07F46SL6QZ/p1743968074258309

### DIFF
--- a/app/components/datasources/useProviderConnectionPopup.js
+++ b/app/components/datasources/useProviderConnectionPopup.js
@@ -36,10 +36,17 @@ const useProviderConnectionPopup = ({ popupWatchTimeout = 500, trackMetric = noo
   const previousJustConnectedDataSourceProviderName = usePrevious(justConnectedDataSourceProviderName);
 
   const trackConnectionMetric = useCallback((status = null) => {
-    const authorizedProvider = find(providers, { id: authorizedDataSource.id});
-    const providerName = authorizedProvider?.dataSourceFilter?.providerName;
     const isMobile = utils.isMobile();
     const action = status ? 'Completed' : 'Started';
+
+    let providerName;
+    if (authorizedDataSource?.id) {
+      const authorizedProvider = find(providers, { id: authorizedDataSource.id });
+      providerName = authorizedProvider?.dataSourceFilter?.providerName;
+    } else {
+      providerName = location?.query?.dataConnectionStatus;
+    }
+
     trackMetric(`${action} provider connection flow`, { providerName, isMobile, status });
   } , [trackMetric, authorizedDataSource]);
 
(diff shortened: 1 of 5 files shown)


app/components/datasources/useProviderConnectionPopup.js
```diff
@@ -36,10 +36,17 @@ const useProviderConnectionPopup = ({ popupWatchTimeout = 500, trackMetric = noo
   const previousJustConnectedDataSourceProviderName = usePrevious(justConnectedDataSourceProviderName);
 
   const trackConnectionMetric = useCallback((status = null) => {
-    const authorizedProvider = find(providers, { id: authorizedDataSource.id});
```
henry-tp, line 39: the issue was `authorizedDataSource` is undefined when redirecting back here on mobile

app/components/datasources/useProviderConnectionPopup.js
```diff
@@ -36,10 +36,17 @@ const useProviderConnectionPopup = ({ popupWatchTimeout = 500, trackMetric = noo
...
-    const providerName = authorizedProvider?.dataSourceFilter?.providerName;
     const isMobile = utils.isMobile();
     const action = status ? 'Completed' : 'Started';
+
+    let providerName;
+    if (authorizedDataSource?.id) {
+      const authorizedProvider = find(providers, { id: authorizedDataSource.id });
+      providerName = authorizedProvider?.dataSourceFilter?.providerName;
+    } else {
+      providerName = location?.query?.dataConnectionStatus;
```
clintonium-119, line 47: Should this be ` providerName = location?.query?.dataConnectionProviderName`?

app/pages/oauth/OAuthConnection.js
```diff
@@ -80,7 +80,12 @@ export const OAuthConnection = (props) => {
     // patientId after the OAuth callback. We'll redirect back to '/patients' which does have
     // access to the patientId, with a flag to open the DataConnections modal back up on load.
     trackMetric('Oauth - Connection - Redirect back to Tidepool App', { providerName, status });
-    dispatch(push(`/patients?justLoggedIn=true&openDataConnectionsModalWithStatus=${status}`));
+
+    let path = '/patients?justLoggedIn=true'
+             + `&dataConnectionStatus=${status}`
+             + `&dataConnectionProviderName=${providerName}`;
+
```
henry-tp, line 87: This is getting a little messy =( but I added a second param here with the provider name so that we can call `trackMetric` properly when we redirect back

clintonium-119 (changes requested):
Looks good, except that the provider name property in the metric seems incorrect to me.

app/components/datasources/useProviderConnectionPopup.js
henry-tp: Thanks for catching that, yikes. 59730b51d2b0abbf1958d952a413d50f747fb4b1

clintonium-119 (commented):
LGTM :100:

PR #1563, opened 2025-04-11
[WEB-3515] - Allow search of clinic members by email

henry-tp:
[WEB-3515]

The original complaint was that invited members were not searchable, but the issue is that the search bar is filtering by `Name` and invited members have a blank `Name` so any search would filter them out. I asked if filter by `Name` + `Email` would be a viable solution and the answer was yes.

[WEB-3515]: https://tidepool.atlassian.net/browse/WEB-3515?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clintonium-119 (approved):
LGTM :medal_sports:

PR #1565, opened 2025-04-14
[WEB-2895] Enable twiist and disable abbott data connections

clintonium-119:
[WEB-2895]


henry-tp (approved):
Looks like it's not enabled on BE yet, but I trust you have that coordinated with Darin

PR #1570, opened 2025-04-16
Update CI to run in docker instead of in Travis CI container

clintonium-119:
Also updated chromedriver, even though it's not strictly necessary, it seems like a good idea.

I replaced the `disable-gpu` flag in chrome headless settings with `enable-automation`.  `disable-gpu` was only ever needed on Windows machines, and has been deprecated for years.

Made a small change to the production docker build stage as well, since it was using the node_modules compiled in the development stage, which weren't installed with the `immutable` (formerly `frozen-lockfile`) flag.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,5 @@
 os: linux
 dist: jammy
-language: node_js
-
-node_js:
-  - "20.8.0"
-
-cache:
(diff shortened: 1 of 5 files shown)


krystophv (approved):
LGTM :shipit:

PR #1578, opened 2025-04-17
Fix connection metrics

henry-tp:


app/core/personutils.js
```diff
@@ -47,7 +47,7 @@ personUtils.isPatient = (person) => {
 
 personUtils.isClinicianAccount = (user) => {
   const userRoles = _.get(user, 'roles', []);
-  return _.intersection(userRoles, ['clinic', 'migrated_clinic', 'clinician']).length > 0 || user?.isClinicMember;
+  return _.intersection(userRoles, ['clinic', 'migrated_clinic', 'clinician']).length > 0 || user?.isClinicMember || false;
```
clintonium-119, line 50:
```suggestion
  return _.intersection(userRoles, ['clinic', 'migrated_clinic', 'clinician']).length > 0 || !!user?.isClinicMember;
```

Could also force a boolean here on undefined `user` by using `!!`.  Your call.

clintonium-119 (approved):
LGTM :+1:   One small code suggestion in comments, but it's entirely optional, as what you've done there works too.

henry-tp:
Hey @clintonium-119 , I added one more effect to the DataConnections modal. It just prevents the query param from hanging around which can cause improper toast messages and trackMetric firings. Let me know if you're good with that.

clintonium-119 (approved):
Changes LGTM :+1:

PR #1587, opened 2025-04-25
[WEB-3650] New clinic create form not redirecting

clintonium-119:
[WEB-3650]

[WEB-3650]: https://tidepool.atlassian.net/browse/WEB-3650?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/pages/clinicdetails/clinicdetails.js
+++ b/app/pages/clinicdetails/clinicdetails.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { useDispatch, useSelector } from 'react-redux';
 import { withTranslation, Trans } from 'react-i18next';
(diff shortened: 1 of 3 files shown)


app/pages/clinicdetails/clinicdetails.js
```diff
@@ -169,14 +173,21 @@ export const ClinicDetails = (props) => {
...
           }
         }
       }
     }
-  }, [clinic, submitting]);
+  }, [
+    action,
+    clinic,
+    submitting,
+    previousSubmitting,
```
krystophv, line 187: `previousSubmitting` is no longer used and can be excluded (though should only change with `submitting`, so it's likely harmless). this is the only reference to the variable outside of declaration, so we could drop it from the file entirely I think.

krystophv (approved):
LGTM 🚀 

can drop `previousSubmitting`, but still giving the approval

app/pages/clinicdetails/clinicdetails.js
clintonium-119: Good call.  Will remove

PR #1593, opened 2025-05-02
Web 3595 setup rtl 2

henry-tp:


jest.config.js
```diff
@@ -0,0 +1,50 @@
...
+  transform: {
+    '^.+\\.(js|jsx|ts|tsx)$': 'babel-jest',
+  },
+  transformIgnorePatterns: [
+    // This pattern ensures that Babel processes @tidepool/viz and all its dependencies.
+    // In the event that a `Jest failed to parse a file` error is encountered and it is
+    // a node_module, the module name can be added here to be processed by Babel.
+
+    'node_modules/(?!(.*\\.mjs$|@tidepool/viz|internmap|react-markdown|vfile|unist-util-stringify-position|unified|bail|is-plain-obj|trough|remark-parse|mdast-util-from-markdown|mdast-util-to-string|mdast-util-to-hast|micromark|decode-named-character-reference|remark-rehype|unist-util-position|trim-lines|unist-util-visit|unist-util-is|unist-util-generated|mdast-util-definitions|property-information|hast-util-whitespace|space-separated-tokens|comma-separated-tokens))',
+  ],
```
henry-tp, line 36: Per discussion from https://tidepoolteam.slack.com/archives/CFSURMMS9/p1746146214613799

Ugly, but does the job.

jest.config.js
```diff
@@ -0,0 +1,50 @@
...
+    'd3-scale': '<rootDir>/node_modules/d3-scale/dist/d3-scale.min.js',
+    'd3-shape': '<rootDir>/node_modules/d3-shape/dist/d3-shape.min.js',
+    'd3-time': '<rootDir>/node_modules/d3-time/dist/d3-time.min.js',
+    'd3-interpolate': '<rootDir>/node_modules/d3-interpolate/dist/d3-interpolate.min.js',
+    'd3-color': '<rootDir>/node_modules/d3-color/dist/d3-color.min.js',
+    'd3-path': '<rootDir>/node_modules/d3-path/dist/d3-path.min.js',
+    'd3-ease': '<rootDir>/node_modules/d3-ease/dist/d3-ease.min.js',
+    'd3-timer': '<rootDir>/node_modules/d3-timer/dist/d3-timer.min.js',
+    'd3-voronoi': '<rootDir>/node_modules/d3-voronoi/dist/d3-voronoi.min.js',
+    'd3-time-format': '<rootDir>/node_modules/d3-time-format/dist/d3-time-format.min.js',
```
henry-tp, line 20: d3 provides pre-transpiled versions so we can just use those instead of adding all of these packages to `transformIgnorePatterns`

jest.config.js
```diff
@@ -0,0 +1,50 @@
...
+
+    // Mock problematic ES modules
+    'property-information': '<rootDir>/__mocks__/property-information.js',
```
henry-tp, line 23: This is the weirdest one. If I put it into `transformIgnorePatterns` to transpile, I get this:

image

`ReferenceError: boolean is not defined`

I looked into the files in `node_modules` and couldn't pin down the issue.

So I just mocked it here, I really think it will be inconsequential.

app/components/elements/Banner.js
```diff
@@ -99,7 +99,7 @@ Banner.propTypes = {
   ...FlexProps,
   actionText: PropTypes.string,
   dismissable: PropTypes.bool,
-  label: PropTypes.string.isRequired,
+  label: PropTypes.string,
```
henry-tp, line 102: It was just throwing an ugly warning :) but we use Banners without labels in a few places so I think it's not required

__mocks__/property-information.js
```diff
@@ -0,0 +1,25 @@
+/* global jest */
+
+// Mock for property-information
```
henry-tp, line 3: Explanation: https://github.com/tidepool-org/blip/pull/1593/files#r2072217192

test/unit/redux/actions/async.test.js
```diff
@@ -434,9 +434,11 @@ describe('Actions', () => {
 
     describe('acceptTerms', () => {
       it('should trigger ACCEPT_TERMS_SUCCESS and it should call acceptTerms once for a successful request', () => {
-        let acceptedDate = new Date();
+        const now = new Date();
+
+        let acceptedDate = now;
         let loggedInUserId = 500;
-        let termsData = { termsAccepted: new Date() };
+        let termsData = { termsAccepted: now };
```
henry-tp, line 441: These tests were just being really flaky for me on Travis

package.json
```diff
@@ -7,7 +7,9 @@
   "version": "1.85.0-web-3029-dep-updates.2",
   "private": true,
   "scripts": {
-    "test": "TZ=UTC NODE_ENV=test NODE_OPTIONS='--max-old-space-size=4096' yarn karma start",
+    "test": "TZ=UTC NODE_ENV=test yarn jest __tests__ --verbose; TZ=UTC NODE_ENV=test NODE_OPTIONS='--max-old-space-size=4096' yarn karma start",
```
henry-tp, line 10: I changed it to `;` instead of `&&` to ensure that the karma/enzyme suite always runs even if the jest/RTL suite fails

package.json
```diff
@@ -7,7 +7,9 @@
   "version": "1.85.0-web-3029-dep-updates.2",
   "private": true,
   "scripts": {
-    "test": "TZ=UTC NODE_ENV=test NODE_OPTIONS='--max-old-space-size=4096' yarn karma start",
+    "test": "TZ=UTC NODE_ENV=test yarn jest __tests__ --verbose; JEST_EXIT_CODE=$?; TZ=UTC NODE_ENV=test NODE_OPTIONS='--max-old-space-size=4096' yarn karma start; KARMA_EXIT_CODE=$?; [ $JEST_EXIT_CODE -eq 0 ] && [ $KARMA_EXIT_CODE -eq 0 ]",
```
henry-tp, line 10: Okay, I won't lie, I AI'd my way through this command, but it makes logical sense to me, and I've verified each scenario. Please have a look at the command to see if I've missed any cases.

The idea is that I needed to address a few scenarios that using a simpler command wouldn't allow us to address:

`yarn jest && yarn karma start`: if the jest suite fails, we don't run the karma suite. Ideally, we should always run both test suites.

`yarn jest; yarn karma start`: both test suites will run, but the success of both suites will be based on the success of karma. So if the jest suite fails but the karma suite succeeds, Travis will still consider it a pass. Not good.

So here, I'm always running both test suites, and we're returning Exit status 0 only if both suites succeed.

package.json
```diff
@@ -7,7 +7,9 @@
   "version": "1.85.0-web-3029-dep-updates.2",
   "private": true,
   "scripts": {
-    "test": "TZ=UTC NODE_ENV=test NODE_OPTIONS='--max-old-space-size=4096' yarn karma start",
+    "test": "TZ=UTC NODE_ENV=test yarn jest --verbose --runInBand; JEST_EXIT_CODE=$?; TZ=UTC NODE_ENV=test NODE_OPTIONS='--max-old-space-size=4096' yarn karma start; KARMA_EXIT_CODE=$?; [ $JEST_EXIT_CODE -eq 0 ] && [ $KARMA_EXIT_CODE -eq 0 ]",
```
henry-tp, line 10: The idea is that I needed to address a few scenarios that using a simpler command wouldn't allow us to address:

`yarn jest && yarn karma start`: if the jest suite fails, we don't run the karma suite. Ideally, we should always run both test suites.

`yarn jest; yarn karma start`: both test suites will run, but the success of both suites will be ONLY based on the success of karma. So if the jest suite fails but the karma suite succeeds, Travis will still consider it a pass. Not good.

So here, I'm always running both test suites, and we're returning Exit status 0 only if both suites succeed.

Won't lie, I AI'd my way through this command, but it makes logical sense to me based on what I can find online. I've manually verified that Travis will A) still run the Karma suite if the Jest suite fails and B) fail the entire run if the Jest suite fails but the Karma suite passes. Please let me know if I've missed any cases.

jest.config.js
```diff
@@ -0,0 +1,47 @@
...
+    'd3-voronoi': '<rootDir>/node_modules/d3-voronoi/dist/d3-voronoi.min.js',
+    'd3-time-format': '<rootDir>/node_modules/d3-time-format/dist/d3-time-format.min.js',
+  },
+  setupFilesAfterEnv: ['<rootDir>/jest.setup.js'],
+  testMatch: [
+    '**/__tests__/**/*.[jt]s?(x)',
+  ],
+  transform: {
+    '^.+\\.(js|jsx|ts|tsx)$': [
+      '@swc/jest',
```
henry-tp, line 28: SWC and ESBuild seem comparable. I also had it work with a transformer called `esbuild-jest`: 099a9a0a0d041aca7c58a4dc0d2a9408cdbf8396

I have no strong preference, I think performance is about the same, but the `@swc/jest` npm package seems to have better support

package.json
henry-tp: Per `https://jestjs.io/docs/troubleshooting#tests-are-extremely-slow-on-docker-andor-continuous-integration-ci-server`

For performance-enhancing flags, there are two options as I can find - `--runInBand` or `--maxWorkers=4`. They are opposed to eachother, as `--runInBand` forces all tests to run sequentially with no worker spawning, while `--maxWorkers=4` is meant to allow parallelism but prevent Jest from spawning lots workers which our Travis setup can't really handle.

It seems like it's unknown at this point as to which would be faster as our test suite grows. I think maybe we can revisit this once we start really adding tests if performance becomes an issue.

krystophv: Can we get away with using `concurrently` here? I don't know (and admit I haven't checked) what the performance implications might be. It also might interleave logging messages in a way that's not ideal, but seems like it would solve the "all subsequent processes need to exit 0" issue.

henry-tp: I just gave it a shot. One immediate thing it's doing is mixing in Jest output with Karma output in the terminal, which is problematic for debugging:

image

So in my eyes I think it's a no

henry-tp: I just gave it a shot with this command:

```bash
"concurrently \"TZ=UTC NODE_ENV=test yarn jest --verbose --runInBand\" \"TZ=UTC NODE_ENV=test NODE_OPTIONS='--max-old-space-size=4096' yarn karma start\""
```

As you suggested, it's mixing in Jest output with Karma output in the terminal, which is problematic for debugging:

image

So I'd be open to an alternative but I'm not a fan of that fact personally. I just looked up packages and theres one called `npm-run-all` but it's not something we already have, and I'd prefer not to add another package just for this. Open to other suggestions

henry-tp: Okay, so after testing with 60 files:

`concurrently` seems to mostly complete at the same time as a script that runs them sequentially. However, with my testing setup, I ran into some build failures given the number of tests I was running:
https://github.com/tidepool-org/blip/pull/1593/checks?check_run_id=41824132798
https://app.travis-ci.com/github/tidepool-org/blip/builds/275119578

Looking at the logs, I can't say for sure what the issue, I am assuming it's a resource or timeout issue. However, this scenario _may_ not be realistic, as at that point the whole CI suite took 35 minutes to run anyway.

So using `concurrently` seems to work for now, but I'd like to keep the sequential `test` script saved under a different alias in case we run into any issues.

.eslintrc
```diff
@@ -40,5 +40,14 @@
...
   "plugins": [
     "react"
-  ]
+  ],
+  "overrides": [
+		{
+			"files": ["**/__tests__/**/*.test.[jt]s?(x)"],
+			"extends": ["plugin:testing-library/react"],
+      "rules": {
+        "testing-library/prefer-presence-queries": "off"
```
henry-tp, line 49: https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/prefer-presence-queries.md

I really don't think I'd be a fan of this rule. The `getBy` queries error instead of returning null, so you have to use `getBy` for truthy assertions and `queryBy` for falsy assertions, which is a weird thing to have to remember. Idk.

krystophv: The [justification I've seen for this](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-query-variants-for-anything-except-checking-for-non-existence) is that the test failure will be much _nicer_. We might also want to pick up [jest-dom](https://github.com/testing-library/eslint-plugin-jest-dom) rules as well (per the same blog post's suggestion)

henry-tp: Okay, I've added the `jest-dom` rules as well. If it's okay with you, I'll leave `testing-library/prefer-presence-queries` to `warn` instead of `error`

package.json
henry-tp: cc @clintonium-119 just so this conversation doesn't get lost, I think the above is important discussion when you take a look at this :)

__tests__/unit/app/components/clinic/WorkspaceSwitcher.test.js
```diff
@@ -0,0 +1,100 @@
...
+            <Route path='/'>
+              <WorkspaceSwitcher trackMetric={trackMetric} api={api} />
+            </Route>
+          </Switch>
+        </MemoryRouter>
+      </Provider>
+    );
+
+    // Workspace switcher should be labelled with currently selected clinic name
+    const dropdownButton = screen.getByRole('button', { id: 'workspace-switcher-button' });
```
krystophv, line 76: `getByRole` doesn't have an `id` option, I don't think: https://testing-library.com/docs/queries/byrole/#api

jest.setup.js
```diff
@@ -0,0 +1,19 @@
+/* global jest */
+
+// Add any global setup for Jest tests here
+require('@babel/polyfill');
```
krystophv, line 4: Given that `@babel/polyfill` has been deprecated for _a while_ (https://babeljs.io/docs/babel-polyfill) is this an opportunity to update to the newer `import "core-js/stable";`?

__tests__/unit/app/components/clinic/WorkspaceSwitcher.test.js
henry-tp: Ahh, you're right. In fact there isn't even anything with an id of `workspace-switcher-button`, don't know how that slipped through =P looks like the query param was just ignored.

6fa2be99bce45f85331344aa14ab9b3cfc22b486

jest.setup.js
henry-tp: Replaced with `core-js/stable` 3a59d7f67475b7301dace20851b7faf8126e600d

package.json
```diff
@@ -64,7 +68,13 @@
     "@storybook/cli": "7.5.0",
     "@storybook/react": "7.5.0",
     "@storybook/react-webpack5": "7.5.0",
+    "@swc/core": "^1.11.24",
+    "@swc/jest": "^0.2.38",
+    "@testing-library/dom": "^10.4.0",
+    "@testing-library/jest-dom": "^6.6.3",
+    "@testing-library/react": "^12.1.5",
     "@testing-library/react-hooks": "8.0.1",
+    "@testing-library/user-event": "^14.6.1",
```
krystophv, line 77: We always try to specify exact dependency versions. These will need to be locked down. 🔒  (along with the eslint and jest deps lower down)

krystophv (approved):
LGTM 🚀

PR #1595, opened 2025-05-05
WEB-2367 - Design Feedback 1

henry-tp:
WEB-2367

Feedback from Design:

"I totally failed to define this in the Figma but would it be possible to hide the blue bar when no filters/search are applied? There are a few reasons for this: ...`


clintonium-119 (approved):
LGTM :+1:

PR #1601, opened 2025-05-09
[WEB-3662] - Step 1 - Suggest Tags

henry-tp:
[WEB-3662]

[WEB-3662]: https://tidepool.atlassian.net/browse/WEB-3662?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/components/clinic/PatientForm/PatientForm.js
```diff
@@ -350,55 +363,24 @@ export const PatientForm = (props) => {
...
-                  onClick: tagId => {
-                    setFieldValue('tags', [...values.tags, tagId]);
-                  },
-                }}
-              />
-            </Box>
-          )}
+        <Box ref={tagSectionRef} mb={3}>
+          <MediumTitle mb={2} sx={{ fontWeight: 'bold', fontSize: 2 }}>{t('Tags')}</MediumTitle>
+
```
henry-tp, line 375: The logic specifically for the tag selection combobox is starting to get a little dense so IMO it's time to move it into its own concern. I will do the same for sites

app/components/clinic/PatientForm/PatientForm.js
```diff
@@ -14,25 +14,23 @@ import map from 'lodash/map';
...
-import * as actions from '../../redux/actions';
-import TextInput from '../../components/elements/TextInput';
-import { TagList } from '../../components/elements/Tag';
-import { useToasts } from '../../providers/ToastProvider';
-import { getCommonFormikFieldProps } from '../../core/forms';
-import { useInitialFocusedInput, useIsFirstRender, usePrevious } from '../../core/hooks';
-import { dateRegex, patientSchema as validationSchema } from '../../core/clinicUtils';
-import { accountInfoFromClinicPatient } from '../../core/personutils';
-import { Body0 } from '../../components/elements/FontStyles';
-import { borders, colors } from '../../themes/baseTheme';
```
henry-tp, line 35: just cleaning up / deleting unused imports and declarations

test/unit/pages/ClinicPatients.test.js
```diff
@@ -2750,116 +2769,6 @@ describe('ClinicPatients', () => {
             sinon.assert.calledWith(defaultProps.trackMetric, 'Clinic - Population Health - Assign patient tag confirm', sinon.match({ clinicId: 'clinicID123' }));
           });
         });
-
-        it('should allow updating tags for a patient', done => {
```
henry-tp, line 2754: Test in RTL

app/components/clinic/PatientForm/PatientForm.js
```diff
@@ -88,50 +91,9 @@ export const PatientForm = (props) => {
   const previousFetchingPatientsForClinic = usePrevious(fetchingPatientsForClinic);
   const previousFetchOptions = usePrevious(patientFetchOptions);
   const initialFocusedInputRef = useInitialFocusedInput();
+  const tagSectionRef = useRef(null);
 
-  const dexcomConnectStateUI = {
```
henry-tp, line 92: Doesn't look like this is being used anywhere, and I can safely remove?

jest.setup.js
```diff
@@ -17,3 +17,5 @@ global.__TEST__ = true;
 global.__PROD__ = false;
 global.__I18N_ENABLED__ = 'false';
 global.__DEV_TOOLS__ = false;
+
+window.getComputedStyle = () => ({ getPropertyValue: () => undefined });
```
henry-tp, line 23: I think this is fine for us because we never really test the styles (nor should we IMO). There is a property of RTL that it does allow you to really test styles in a more powerful way than with enzyme - For example, an element with style properties of  `opacity: 0`  or `visibility: hidden` will fail the .toBeVisible() check. But I think we’re pretty consistent at using conditional rendering if it really matters if something is showing or hidden. So I think this is fine.

__tests__/unit/components/clinic/PatientForm/SelectTags.test.js
```diff
@@ -0,0 +1,102 @@
...
+        <MemoryRouter initialEntries={['/clinic-workspace']}>
+          <Switch>
+            <Route path='/clinic-workspace'>
+              <SelectTags {...testProps} />
+            </Route>
+          </Switch>
+        </MemoryRouter>
+      </Provider>
+    );
+
```
henry-tp, line 66: There is a school of thought that components like this don't need to be tested, because I am effectively just testing `React-Select` rather than any kind of useful application logic. But, ehh.

__tests__/unit/components/donateform.test.js
```diff
@@ -37,7 +37,7 @@ describe('DonateForm', ()  => {
     expect(saveButton).toHaveTextContent('Save');
 
     // Select dropdown should be disabled
-    expect(screen.queryByRole('combobox')).not.toBeInTheDocument();
+    expect(screen.queryByRole('combobox')).toBeDisabled();
```
henry-tp, line 40: A consequence of replacing the `getComputedStyle` function change, is that it is now still "in the document" because the "disabled" style no longer applies. But the attribute it still there, so we can use `toBeDisabled()`, which is probably actually a better check. Just stuff to keep in mind as we head into the future.

app/components/inputgroup/inputgroup.less
```diff
@@ -64,9 +64,8 @@
   border: 0;
 
   .Select__control {
-    min-height: 40px;
+    min-height: 38px;
```
henry-tp, line 67: Brings it in line with our other inputs which are `38px`

app/pages/clinicworkspace/useClinicPatientsFilters.js
```diff
@@ -0,0 +1,26 @@
...
+  const loggedInUserId = useSelector((state) => state.blip.loggedInUserId);
+
+  const activeFiltersStorageKey = `activePatientFilters/${loggedInUserId}/${selectedClinicId}`;
+  const [activeFilters, setActiveFilters] = useLocalStorage(activeFiltersStorageKey, defaultFilterState, true);
+
+  return [
+    activeFilters,
+    setActiveFilters,
+  ];
+};
```
henry-tp, line 24: This _may_ or may not be a useful abstraction. The reason I wanted to implement it was to allow a number of different components to access the activeFilters from localStorage. The Sites dropdown will also need access to it.

I'm open to removing it and just calling `useLocalStorage(activeFiltersStorageKey, defaultFilterState, true);` multiple times if you would prefer. My preference is to not store/declare the `activeFiltersStorageKey` or `defaultFilterState` in multiple places.

package.json
```diff
@@ -7,8 +7,8 @@
   "version": "1.85.2",
   "private": true,
   "scripts": {
-    "test": "concurrently \"TZ=UTC NODE_ENV=test yarn jest --verbose --runInBand\" \"TZ=UTC NODE_ENV=test NODE_OPTIONS='--max-old-space-size=4096' yarn karma start\" --group --success=all",
-    "test:sequential": "TZ=UTC NODE_ENV=test yarn jest --verbose --runInBand; JEST_EXIT_CODE=$?; TZ=UTC NODE_ENV=test NODE_OPTIONS='--max-old-space-size=4096' yarn karma start; KARMA_EXIT_CODE=$?; [ $JEST_EXIT_CODE -eq 0 ] && [ $KARMA_EXIT_CODE -eq 0 ]",
+    "test": "TZ=UTC NODE_ENV=test yarn jest --verbose --runInBand; JEST_EXIT_CODE=$?; TZ=UTC NODE_ENV=test NODE_OPTIONS='--max-old-space-size=4096' yarn karma start; KARMA_EXIT_CODE=$?; [ $JEST_EXIT_CODE -eq 0 ] && [ $KARMA_EXIT_CODE -eq 0 ]",
```
henry-tp, line 10: @krystophv I had to revert to running them sequentially in order to get the tests to run in a reasonable time. See [this test run](https://app.travis-ci.com/github/tidepool-org/blip/builds/275478882) (you can CMD+F `Test Suites:` which will get you to the Jest Test results). You'll see that running sequentially, the tests for `ClinicPatients.js` still took 10+ seconds. When I run them concurrently with Karma I think it takes like 25+ seconds. My tests IMO are written quite simply and idiomatically. I'm hoping this will improve as we deconstruct our larger components into smaller ones over time.

app/components/clinic/PatientForm/PatientForm.js
clintonium-119: I don't think we actually use the keys at all.  Should be able to remove this completely

app/components/clinic/PatientForm/PatientForm.js
```diff
@@ -9,30 +9,35 @@ import get from 'lodash/get';
...
+import { getCommonFormikFieldProps } from '../../../core/forms';
+import { useInitialFocusedInput, usePrevious } from '../../../core/hooks';
+import { dateRegex, patientSchema as validationSchema } from '../../../core/clinicUtils';
+import { accountInfoFromClinicPatient } from '../../../core/personutils';
+import { Body0 } from '../../../components/elements/FontStyles';
+import { MediumTitle } from '../../../components/elements/FontStyles';
+
+import SelectTags from './SelectTags';
+
+export const DEXCOM_CONNECTION_STATES = [
```
clintonium-119: I don't think we need these at all.  I just missed deleting some of the dexcom stuff when we moved over to the multi-cloud provider UI.

Can remove this.

app/components/clinic/PatientForm/SelectTags.js
```diff
@@ -0,0 +1,113 @@
...
+      closeMenuOnSelect={false}
+      minMenuHeight={selectMenuHeight}
+      maxMenuHeight={selectMenuHeight}
+      filterOption={createFilter({ stringify: opt => opt.label })}
+      isMulti
+      isClearable
+    />
+  );
+};
+
```
clintonium-119, line 113: I think we should define `SelectTags.propTypes`, and perhaps `SelectTags.defaultProps` as it helps with code editor intel.

app/components/clinic/PatientForm/SelectTags.js
```diff
@@ -0,0 +1,113 @@
...
+  // tags currently applied as filters on the clinic patient dashboard.
+  const [suggested, nonSuggested] = partition(options, opt => activeFilters?.patientTags?.includes(opt.value));
+
+  return [
+    { options: suggested, label: t('Suggested - based on current dashboard filters') },
+    { options: nonSuggested, label: '' },
+  ];
+};
+
+export const selectElementStyleOverrides = {
```
clintonium-119, line 37: Is there a need to export this const?

app/pages/clinicworkspace/useClinicPatientsFilters.js
clintonium-119: I'm good with this approach.

__tests__/unit/components/clinic/PatientForm/SelectTags.test.js
clintonium-119: Yeah.  Agree it's best to avoid generally testing external dependancy operations, but it's fine here, IMO, as you are verifying other things such as the sorting.

Plus, since we're pretty new to RTL here, it's helpful to get our eyes familiar with various testing paradigms that we may not yet be aware of.

clintonium-119: Although, that said, it looks like you export `buildSelectOptions` where that occurs, which I'd assumed was for unit testing it, but it seems to be missing here.  Perhaps You could just unit test that, and do minimal verification for the UI stuff.

app/components/clinic/PatientForm/SelectTags.js
henry-tp: oh, I'm using it in the `<SelectSites />` dropdown as well, so I thought I could just share it between the two

__tests__/unit/components/clinic/PatientForm/SelectTags.test.js
henry-tp: Okay, I have added a unit test for buildSelectOptions, just to verify some sorting differences and the differences when doing suggested vs non-suggested

app/components/clinic/PatientForm/PatientForm.js
henry-tp: Ahh, it wasn't being consumed. Removed fully

app/components/clinic/PatientForm/SelectTags.js
henry-tp: Oh, I'm actually using the same styling for the `<SelectSites />` component, so I imported it there (in a future PR)

app/components/clinic/PatientForm/SelectTags.js
henry-tp: Added PropTypes and defaultProps :)

__tests__/unit/app/pages/clinicworkspace/ClinicPatients.test.js
```diff
@@ -356,6 +356,7 @@ describe('ClinicPatients', ()  => {
             // Click the Edit Tags icon for a patient. The Dialog for Edit Patient Details should open.
             expect(screen.queryByText('Edit Patient Details')).not.toBeInTheDocument();
             await userEvent.click(screen.getAllByTestId('edit-tags-icon')[0]); // Open patient2
+            await userEvent.click(screen.getAllByTestId('edit-tags-icon')[0]); // Open patient2
```
clintonium-119, line 359: Curious why this line was duplicated

PR #1603, opened 2025-05-12
[WEB-3663] Fix JS error for mobile users when initial data view is bgLog or daily

clintonium-119:
[WEB-3663] 

A JS error was occurring due to various calls within the bgLog or daily data views for mobile users due to the chart not being initialized for mobile users, but lifecycle methods still calling various chart methods.  This was noticed by Dave when his production account, which had SMBG as the latest data, was loading the bgLog view by default and hitting the error state.

With this fixed, I noticed that we were still loading the bgLog view by default if the latest datum was an SMBG, so I've updated the default view to always be 'basics' for mobile users when not specified via a URL param.  This way, the user will see the full list of stats available on the basics view, instead of the SMBG-only stats shown on the BG Log view

[WEB-3663]: https://tidepool.atlassian.net/browse/WEB-3663?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/chart/bgLog.js
+++ b/app/components/chart/bgLog.js
@@ -94,7 +94,7 @@ class BgLogChart extends Component {
 
   unmountChart = () => {
     this.log('Unmounting...');
-    if (this.chart) this.chart.destroy();
+    if (this.chart) this.chart?.destroy();
   };
 
   remountChart = (updates = {}) => {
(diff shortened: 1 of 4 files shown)


henry-tp (approved):

henryliang2:
Sorry I missed these. They didnt come up in my manual testing.

PR #1608, opened 2025-05-14
[WEB-3670] Add cbgFilter query parameter to backend API calls

darinkrauss:
- Add cbgFilter query parameter to backend API calls
- https://tidepool.atlassian.net/browse/WEB-3670


clintonium-119 (approved):
LGTM :100:

PR #1628, opened 2025-06-04
Release 1.85.2 to master

clintonium-119:
Release 1.85.2 to master


krystophv:
Note: 2 transient false test failure due to UTC timezone issues are showing in the Travis build and are acknowledged and acceptable

PR #1634, opened 2025-06-08
[WEB-3568] - Step 2 - Enable Create Sites

henry-tp:
[WEB-3568]

Contains work from #1601; That PR should be reviewed first

https://github.com/user-attachments/assets/3e707b62-03c8-4575-9d11-46a3b51a06e5

[WEB-3568]: https://tidepool.atlassian.net/browse/WEB-3568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/core/clinicUtils.js
```diff
@@ -373,11 +375,17 @@ export const clinicSchema = yup.object().shape({
   timezone: yup.string(),
 });
 
+export const clinicSiteSchema = yup.object().shape({
+  name: yup.string()
+    .max(20, t('Site name max length is ${max} characters'))
+    .matches(/^[\p{L}\p{N}_+><-]{1}[\p{L}\p{N}\s_+><-]*$/u, t('Allowed special characters: - _ + > <'))
+});
```
henry-tp, line 382: Check: Validation/limitations?

app/redux/actions/async.js
```diff
@@ -2662,6 +2662,39 @@ export function sendPatientDataProviderConnectRequest(api, clinicId, patientId,
...
+    dispatch(sync.createClinicSiteRequest());
+
+    api.clinics.createClinicSite(clinicId, siteToCreate, (err, site) => {
+      if (err) {
+        let message = ErrorMessages.ERR_CREATING_CLINIC_PATIENT_TAG; // TODO: CHANGE TO APPROPRIATE
+
+        if (err.status === 422) {
+          message = ErrorMessages.ERR_CREATING_CLINIC_PATIENT_TAG_MAX_EXCEEDED; // TODO: CHANGE TO APPROPRIATE
+        } else if (err.status === 409) {
+          message = ErrorMessages.ERR_CREATING_CLINIC_PATIENT_TAG_DUPLICATE; // TODO: CHANGE TO APPROPRIATE
```
henry-tp: NTS: Must create error message classes

test/unit/pages/ClinicPatients.test.js
```diff
@@ -1812,24 +1812,6 @@ describe('ClinicPatients', () => {
             sinon.assert.calledWith(defaultProps.trackMetric, 'Clinic - Population Health - Edit clinic tags open', sinon.match({ clinicId: 'clinicID123', source: 'Filter menu' }));
           });
 
-          it('should allow adding a clinic patient tag', done => {
```
henry-tp, line 1839: Rewritten in RTL

app/pages/clinicworkspace/ClinicPatients.js
```diff
@@ -1687,6 +1716,165 @@ export const ClinicPatients = (props) => {
...
+                          </Text>
+                          { sortedSiteFilterOptions.length > 0 &&
+                            <Text sx={{ display: 'block', color: colors.gray50, fontSize: 0, fontStyle: 'italic', maxWidth: '208px', whiteSpace: 'wrap', lineHeight: 1 }}>
+                              {t('Any patient with one or more of the sites you select below will be shown.')}
+                            </Text>
+                          }
+                        </Box>
+
+                        { // Render a list of checkboxes
+                          sortedSiteFilterOptions.map(({ id, label }) => {
```
henry-tp, line 1790: PENDING ISSUE: When we create a new site the response body is

```json
{ "name": "New Site" }
```

This leads to the `id` being `undefined` and causes a key error here as multiple checkboxes will have the key `clinic-site-filter-option-undefined`.

I have asked Eric to add the `id` to the response, hoping he has no problem with that. If we don't get that then I will need to make a really annoying refetch to get the new key.

https://tidepoolteam.slack.com/archives/C08G8SCUTDK/p1749511076242879

app/redux/reducers/misc.js
```diff
@@ -944,6 +944,17 @@ export const clinics = (state = initialState.clinics, action) => {
...
     }
+    case types.CREATE_CLINIC_SITE_SUCCESS: {
+      const {
+        clinicId,
+        clinicSites,
+        newSite,
+      } = action.payload;
+
+      return update(state, {
+        [clinicId]: { sites: { $set: [...clinicSites, newSite] } },
```
henry-tp: This is a deviation from `tags`, where we get the entire list of clinic tags as an API response. For the Sites API it is only responding with the new tag. I'm not a big fan of modifying response bodies like this, so I have asked Eric to see if it's a big ask to return a full list of clinic sites after a CRUD operation. Hoping he says it's okay, but if not I'll have to add in refetching and do some modification of state this way.

https://tidepoolteam.slack.com/archives/C08G8SCUTDK/p1749511076242879

app/pages/clinicworkspace/ClinicPatients.js
```diff
@@ -1687,6 +1717,165 @@ export const ClinicPatients = (props) => {
...
+                      icon={KeyboardArrowDownRoundedIcon}
+                      iconLabel="Filter by clinic sites"
+                      sx={{ fontSize: 0, lineHeight: 1.3 }}
+                    >
+                      <Flex sx={{ alignItems: 'center', gap: 1 }}>
+                        {showTideDashboard && !clinic?.sites?.length && <Icon
+                          variant="static"
+                          icon={InfoOutlinedIcon}
+                          sx={{ fontSize: '14px' }}
+                        />}
```
henry-tp: Copied over from Tags, but I'm unsure of what meaning or purpose this icon has, given that it is only rendered when TIDE Dashboard is available, and I'm not sure if this is relevant for sites. This is the icon:

image

app/pages/clinicworkspace/ClinicPatients.js
```diff
@@ -1845,6 +2005,36 @@ export const ClinicPatients = (props) => {
                     }
                   </Popover>
 
+                  <Button
```
henry-tp, line 2078: The ordering has changed per PROD-250. New order is. This button is just cut/pasted from ~line 1660

| Data Recency | Sites | Tags | % Time In Range | % CGM Use | 
|---|---|---|---|---|

__tests__/unit/app/pages/clinicworkspace/ClinicPatients.test.js
```diff
@@ -337,20 +361,78 @@ describe('ClinicPatients', ()  => {
...
+            await userEvent.click(screen.getByRole('button', { name: /Tags/ }));
+            await userEvent.click(screen.getByRole('button', { name: /Edit Tags/ }));
+
+            // Type in a new tag "Delta" into the textbox and click add
+            const newTag = screen.getByRole('textbox');
+            await userEvent.click(newTag);
+            await userEvent.paste('Tag Delta');
+            await userEvent.click(screen.getByRole('button', { name: /Add/ }));
+
+            await waitFor(() => expect(defaultProps.api.clinics.createClinicSite).toHaveBeenCalled());
```
clintonium-119: Should this be `createClinicPatientTag` here?  Assuming that's the case, it may be prudent to `mockClear` both 
`defaultProps.api.clinics.createClinicSite` and `defaultProps.api.clinics.createClinicPatientTag` in the beforeEach block to avoid false positives due to mock history carrying over from previous tests.

```suggestion
            await waitFor(() => expect(defaultProps.api.clinics.createClinicPatientTag).toHaveBeenCalled());
```

app/pages/clinicworkspace/ClinicPatients.js
clintonium-119: The design intent was to call attention to the location where the user could add tags due to a requirement that there be at least be one tag available for TIDE dash configuration.

I don't think we need it here for Sites.

app/pages/clinicworkspace/ClinicPatients.js
```diff
@@ -2521,7 +2878,7 @@ export const ClinicPatients = (props) => {
               )}
             </Formik>
 
-            { patientTagsFilterOptions.length > 0 &&
+            { clinicSitesFilterOptions.length > 0 &&
```
clintonium-119: Was this meant to be changed here?  If not, should be reverted back to
```suggestion
            { patientTagsFilterOptions.length > 0 &&
```

app/redux/actions/async.js
```diff
@@ -2662,6 +2662,39 @@ export function sendPatientDataProviderConnectRequest(api, clinicId, patientId,
   };
 }
 
+/**
+ * Create a site for a clinic
+ *
+ * @param {Object} api - an instance of the API wrapper
+ * @param {String} clinicId - Id of the clinic
+ * @param {Object} clinicSites - all existing clinic sites
```
clintonium-119: These params defs are not matching the current function signature

__tests__/unit/app/pages/clinicworkspace/ClinicPatients.test.js
henry-tp: Yikes, you are correct. It did indeed fail after adding the `.mockClear()`. I've added the mockClear(s) now and fixed the test.

app/pages/clinicworkspace/ClinicPatients.js
henry-tp: Sounds good, removed

app/pages/clinicworkspace/ClinicPatients.js
henry-tp: Yikes, you are correct. Fixed

app/redux/actions/async.js
henry-tp: Oops, I had originally created the function and then had to modify it. Fixed

PR #1635, opened 2025-06-08
[WEB-3719] - Fix KissMetrics when sorting table

henry-tp:
[WEB-3719]

Some trackmetric calls are being made with `undefined` as part of the string.

[WEB-3719]: https://tidepool.atlassian.net/browse/WEB-3719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/pages/clinicworkspace/ClinicPatients.js
+++ b/app/pages/clinicworkspace/ClinicPatients.js
@@ -38,7 +38,7 @@ import SearchIcon from '@material-ui/icons/Search';
 import VisibilityOffOutlinedIcon from '@material-ui/icons/VisibilityOffOutlined';
 import VisibilityOutlinedIcon from '@material-ui/icons/VisibilityOutlined';
 import ArrowUpwardIcon from '@material-ui/icons/ArrowUpward';
-import { components as vizComponents, utils as vizUtils } from '@tidepool/viz';
+import { components as vizComponents, utils as vizUtils, colors as vizColors } from '@tidepool/viz';
 import sundial from 'sundial';
 import ScrollToTop from 'react-scroll-to-top';
 import styled from '@emotion/styled';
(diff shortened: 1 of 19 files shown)


app/pages/clinicworkspace/ClinicPatients.js
```diff
@@ -1277,6 +1277,11 @@ export const ClinicPatients = (props) => {
 
       const sortColumnLabels = {
         fullName: 'Patient details',
+        lastData: 'Data recency',
+        glucoseManagementIndicator: 'GMI',
+        averageGlucoseMmol: 'Average glucose',
+        timeInVeryLowRecords: 'Time in very low',
+        timeInVeryHighRecords: 'Time in very high',
```
henry-tp, line 1284: Looking at this commit: https://github.com/tidepool-org/blip/commit/021ebd2c5da9a73d3d1bf4b5105e88458e519108

It _seems like_ this call to `trackMetric()` was added and then new columns were added since, and we just didn't update the definitions. After digging for a while, I don't think it's because there are any deeper-rooted reasons why the definitions are misaligned.

test/unit/pages/ClinicPatients.test.js
```diff
@@ -1591,62 +1591,86 @@ describe('ClinicPatients', () => {
           const patientHeader = table.find('#peopleTable-header-fullName .MuiTableSortLabel-root').at(0);
 
           defaultProps.api.clinics.getPatientsForClinic.resetHistory();
+          defaultProps.trackMetric.resetHistory();
```
henry-tp, line 1594: I would rewrite the tests in RTL right now, but I've got a big honkin' feature branch for sites going that sets up the `ClinicPatients.test.js` file in for RTL but it's not ready yet. I'd like to avoid conflict for now.

krystophv (approved):
LGTM :shipit:

PR #1649, opened 2025-06-17
[WEB-3753] - Migrate Tests for ClinicPatients.js

henry-tp:
[WEB-3753]

Most tests have been translated 1:1. Where possible, I have tried to combine test cases and reduce calls to `render()` because it is so expensive and non-performant. This has been recognized as an issue by basically every team that has written an article about their transition to RTL :(

If you run the tests you may also notice a LARGE NUMBER of the following warnings:

```
    Warning: An update to SelectTags inside a test was not wrapped in act(...).

    When testing, code that causes React state updates should be wrapped into act(...):

    act(() => {
      /* fire events that update state */
    });
    /* assert on the output */
```

I have not been able to work my way around these warnings. Here is a resource I've found helpful for understanding these warnings: https://kentcdodds.com/blog/fix-the-not-wrapped-in-act-warning

I think the primary issue is that we have a lot of side-effect based re-rendering in this page (and our application in general). So side-effect triggers are happening after the expectations we set forth. These warnings are only warnings, and based on my reading, they are just kind of saying "Hey, there is still stuff changing after your `expect()` call". The methods used in the video above^ to resolve the act call I'm not really comfortable with doing, especially the ones involving modifying the actual component structure. So I think we can leave them.

[WEB-3753]: https://tidepool.atlassian.net/browse/WEB-3753?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


henry-tp:
Targetting release-1.89.0 just to resolve conflicts, not to merge into release branch. DO NOT MERGE into release 1.89.0

PR #1671, opened 2025-07-11
WEB-3693 - Reduce duration of "inviteJustSent" state to 3 seconds

henry-tp:
[WEB-3693]

Sharon pointed out this ticket as a small one that potentially I could help with. In actuality though, I struggled quite a bit on how to implement this given the existing architecture. I realized it would not work to simply replace the `minutesAgo` check with a `secondsAgo` check, since we only re-render every minute. I didn't want us to be re-rendering every second just to make this work, so I looked to find an alternative method based on internal state.

If you had something (simpler) in mind feel free to reject this solution.

https://github.com/user-attachments/assets/ce4af223-c86c-44b4-be74-eb7fe844ad22

[WEB-3693]: https://tidepool.atlassian.net/browse/WEB-3693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/datasources/DataConnections.js
+++ b/app/components/datasources/DataConnections.js
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useState } from 'react';
+import React, { useMemo, useCallback, useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { useDispatch, useSelector } from 'react-redux';
 import ErrorRoundedIcon from '@material-ui/icons/ErrorRounded';
@@ -171,13 +171,11 @@ export const getConnectStateUI = (patient, isLoggedInUser, providerName) => {
     ]);
 
   let timeAgo;
-  let inviteJustSent;
 
   if (mostRecentConnectionUpdateTime) {
     const { daysAgo, daysText, hoursAgo, hoursText, minutesAgo, minutesText } = formatTimeAgo(mostRecentConnectionUpdateTime);
     timeAgo = daysText;
     if (daysAgo < 1)  timeAgo = hoursAgo < 1 ? minutesText : hoursText;
-    if (!isLoggedInUser && minutesAgo < 1) inviteJustSent = true;
   }
 
   let patientConnectedMessage;
@@ -220,7 +218,6 @@ export const getConnectStateUI = (patient, isLoggedInUser, providerName) => {
       icon: null,
       message: t('Invite sent {{timeAgo}}', { timeAgo }),
       text: t('Connection Pending'),
-      inviteJustSent,
     },
     pendingReconnect: {
       color: colors.grays[5],
@@ -228,7 +225,6 @@ export const getConnectStateUI = (patient, isLoggedInUser, providerName) => {
       icon: null,
       message: t('Invite sent {{timeAgo}}', { timeAgo }),
       text: t('Invite Sent'),
-      inviteJustSent,
     },
     pendingExpired: {
       color: colors.feedback.warning,
@@ -269,7 +265,7 @@ export const getConnectStateUI = (patient, isLoggedInUser, providerName) => {
   }
 };
 
-export const getDataConnectionProps = (patient, isLoggedInUser, selectedClinicId, setActiveHandler) => reduce(activeProviders, (result, providerName) => {
+export const getDataConnectionProps = (patient, isLoggedInUser, selectedClinicId, setActiveHandler, justInvitedProviders) => reduce(activeProviders, (result, providerName) => {
   result[providerName] = {};
 
   let connectState;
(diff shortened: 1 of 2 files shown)


app/components/datasources/DataConnections.js
```diff
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, useState } from 'react';
+import React, { useMemo, useCallback, useEffect, useState } from 'react';
```
henry-tp-public, line 1: test

app/components/datasources/DataConnections.js
```diff
@@ -329,6 +329,42 @@ export const getDataConnectionProps = (patient, isLoggedInUser, selectedClinicId
   return result;
 }, {});
 
+export const useJustInvitedProviders = () => {
+  // initialState in the form of { twiist: false, dexcom: false, etc: false }
+  const initialState = useMemo(() =>{
+    return Object.fromEntries(activeProviders.map(s => [s, false]));
+  }, [activeProviders]);
```
henry-tp-public, line 336: Stable reference, `activeProviders` is not. Missing from dependencies or wrongly included, it may be. Import this constant or ensure stable it remains, you must. If constant `activeProviders` is, remove from dependencies you should.

app/components/datasources/DataConnections.js
```diff
@@ -329,6 +329,42 @@ export const getDataConnectionProps = (patient, isLoggedInUser, selectedClinicId
...
+    const hasJustInvitedProviders = Object.values(justInvitedProviders).some(Boolean);
+
+    if (!hasJustInvitedProviders) return;
+
+    const timeoutId = setTimeout(() => {
+      setJustInvitedProviders(initialState);
+    }, 3000);
+
+    return () => clearTimeout(timeoutId);
+  }, [justInvitedProviders, initialState]);
```
henry-tp-public, line 351: Dangerous path to infinite re-renders, this dependency array leads. When `justInvitedProviders` changes, new `initialState` created it may be, causing effect to run again. Remove `initialState` from dependencies, you should. Only `justInvitedProviders` needed here, it is.

app/components/datasources/DataConnections.js
```diff
@@ -329,6 +329,42 @@ export const getDataConnectionProps = (patient, isLoggedInUser, selectedClinicId
...
+
+  // Automatically resets the status to false after waiting 3000ms
+  useEffect(() => {
+    const hasJustInvitedProviders = Object.values(justInvitedProviders).some(Boolean);
+
+    if (!hasJustInvitedProviders) return;
+
+    const timeoutId = setTimeout(() => {
+      setJustInvitedProviders(initialState);
+    }, 3000);
```
henry-tp-public, line 348: Magic numbers, the path to the dark side they are. Extract to named constant, you should:
```javascript
const INVITE_JUST_SENT_DURATION = 3000;
```
Clear and maintainable, this makes the code.

test/unit/components/datasources/DataConnections.test.js
```diff
@@ -615,6 +618,18 @@ describe('DataConnections', () => {
     });
 
     describe('data connection invite just sent', () => {
+      beforeEach(() => {
+        DataConnections.__Rewire__('useJustInvitedProviders', sinon.stub().returns({
+          justInvitedProviders: { dexcom: true, twiist: true },
+          handleProviderInviteSent: sinon.stub(),
+          isInvitatationHandler: sinon.stub(),
```
henry-tp-public, line 625: Spelled incorrectly, "isInvitatationHandler" is. Match the actual function name "isInvitationHandler", it must. Failed, your tests will, if this typo remains.

app/components/datasources/DataConnections.js
```diff
@@ -157,7 +157,7 @@ export function getProviderHandlers(patient, selectedClinicId, provider) {
   }
 };
 
-export const getConnectStateUI = (patient, isLoggedInUser, providerName) => {
+export const getConnectStateUI = (patient, isLoggedInUser, providerName, justInvitedProviders) => {
```
henry-tp-public, line 160: Breaking change to exported function signature, this is. Document well or consider backward compatibility, you must. Affect other parts of codebase, this change might.

henry-tp-public (dismissed):
Strong with the Force, this solution is, young developer. Clever approach to avoid re-rendering every second, you have taken. Wisdom in your implementation, I sense.

However, several paths to improvement, still there are:

**The Good:**
- Clean separation of concerns, achieved you have
- Proper cleanup of timeouts, implemented it is  
- Avoid frequent re-renders, wise this approach is
- Tests updated appropriately, responsible you are

**Areas for improvement:**
- Infinite re-render potential in useEffect dependencies
- Magic numbers instead of named constants
- Typo in test mock function name
- Breaking changes to exported function signatures

Address these concerns, you should. Then stronger and more maintainable, your code will become. The path of the Jedi coder, patience and attention to detail it requires.

May the Force be with your refactoring! 🌟

henry-tp-public (changes requested):
_

henry-tp:
@clintonium-119 please ignore for now. Discussed with Sharon and we may want a different approach.

PR #1690, opened 2025-07-22
WEB-3570 - Fix Regression due to Site Id missing

henry-tp:
[WEB-3570]

There is a regression due to the `id` not being sent when trying to update a clinic site

image

I'm absolutely confident it was working before. I think there may have been a backend change. The `siteId` is already in the URL, which is why it wasn't needed previously. However, the requirement to have the `id` in the payload matches the [PR Preview API doc](https://developer.tidepool.org/TidepoolApi/pr-preview/pr-159/#tag/Clinics/operation/UpdateSite), so we'll modify the frontend to meet the docs.

[WEB-3570]: https://tidepool.atlassian.net/browse/WEB-3570?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clintonium-119 (approved):
LGTM :100:

PR #1692, opened 2025-07-22
[WEB-3757] Fix data refresh issues for BG Log

clintonium-119:
[WEB-3757]

The primary issue was that the datetime query param could be outside of the range of the fetched data, and it was not properly handled post-refetch.  I ended up trying a few approaches, but decided that the best one was to remove the datetime param after refetch, which allowed the data view to simply fall back to the most recent data time, which is really what we want upon refresh.

I also noticed and fixed a secondary issue that was introduced with our device settings history update - namely that a refresh on the settings view only fetches the pump history.  I added a check within handleRouteChangeEvent to see if the settings view was previously refreshed, and we are now changing to one of the other data views.  In this case, I'm skipping the typical route switch handlers and performing a refresh on the newly-navigated-to chartType, which will fetch the appropriate data and set the view to the most recent data time as desired. 

[WEB-3757]: https://tidepool.atlassian.net/browse/WEB-3757?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/pages/patientdata/patientdata.js
+++ b/app/pages/patientdata/patientdata.js
@@ -1189,45 +1189,57 @@ export const PatientDataClass = createReactClass({
     const chartTypeFromPath = nextProps.match?.params?.chartType;
     const targetDate = utils.parseDatetimeParamToInteger(nextProps.queryParams?.datetime);
 
-    switch(true) {
-      // If the chart is explicitly specified in the URL, we switch to that chart type.
-      case chartTypeFromPath === 'settings':
-        this.handleSwitchToSettings();
-        break;
-      case chartTypeFromPath === 'basics':
-        this.handleSwitchToBasics();
-        break;
-      case chartTypeFromPath === 'daily':
-        this.handleSwitchToDaily(targetDate);
-        break;
-      case chartTypeFromPath === 'trends':
-        this.handleSwitchToTrends(targetDate);
-        break;
-      case chartTypeFromPath === 'bgLog':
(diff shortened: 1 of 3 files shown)


app/pages/patientdata/patientdata.js
```diff
@@ -1189,45 +1189,62 @@ export const PatientDataClass = createReactClass({
...
-        break;
+    // If the chart was previously refreshed on settings, we need to refetch the data, since the refresh
+    // would have only fetched the pump settings history, and not all the data we need to render other charts.
+    const needsDataRefetch = this.state.refreshChartType === 'settings' && _.includes([
+      'basics',
+      'daily',
+      'trends',
+      'bgLog',
+    ], chartTypeFromPath);
+
```
henry-tp, line 1200: Optional Suggestion: Just to reduce nesting, might recommend doing here:

```jsx
if (needsDataRefetch) {
  this.setState()
  return;
}

switch(true) {
  // ...
}
```

as opposed to an `else` block

henry-tp (approved):
Approved with optional comment :)

app/pages/patientdata/patientdata.js
clintonium-119: Yeah, I like that better too.  Will do.

PR #1708, opened 2025-07-28
WEB-3569 - Bug fixes from Design Review

henry-tp:


app/components/clinic/PatientForm/PatientForm.js
```diff
@@ -43,8 +43,8 @@ export function getFormValues(source, clinicPatientTags, clinicSites) {
 }
 
 export function emptyValuesFilter(value, key) {
-  // We want to allow sending an empty `tags` array. Otherwise, strip empty fields from payload.
-  return !includes(['tags'], key) && isEmpty(value);
+  // We want to allow sending an empty `tags` and `sites` arrays. Otherwise, strip empty fields from payload.
+  return !includes(['tags', 'sites'], key) && isEmpty(value);
```
henry-tp, line 47: Bug prevented empty sites array from being sent

app/pages/clinicworkspace/ClinicPatients.js
```diff
@@ -2124,7 +2124,7 @@ export const ClinicPatients = (props) => {
                         }
 
                         { // Display an option to filter for patients with zero tags
-                          sortedSiteFilterOptions.length > 0 &&
+                          sortedTagFilterOptions.length > 0 &&
```
henry-tp, line 2153: Yikes 🤐

app/components/clinic/PatientForm/SelectSites.js
```diff
@@ -53,7 +53,7 @@ const SelectSites = ({
   };
 
   // Suggest sites only if user is viewing ClinicPatients list (where Filters are used)
-  const shouldSuggestSites = pathname === '/clinic-workspace';
+  const shouldSuggestSites = pathname?.includes('/clinic-workspace');
```
henry-tp, line 56: prevented this from being true when viewing `/clinic-workspace/patients`

app/redux/actions/async.js
```diff
@@ -1119,7 +1119,7 @@ export function fetchPatientData(api, options, id) {
             const fetchFromTime = latestDiabetesDatumTime || serverTime;
             const fetchToTime = latestDatumTime || serverTime;
 
-            options.startDate = moment.utc(fetchFromTime).subtract(30, 'days').startOf('day').toISOString();
+            options.startDate ||= moment.utc(fetchFromTime).subtract(30, 'days').startOf('day').toISOString();
```
henry-tp: This is within the `if (options.initial) { ... }` block. 

if `initial: true` is set, then options.startDate would previously always be set to 30 days.

So this change allows setting a custom start date if caller wants when `initial: true` is set in the opts.

Please have a think to see if this could break anything. I grepped around for `fetchPatientData` and I think it's safe?

The alternative is that I could set `initial: false` in my use of the AGP for TIDE Dashboard. But I'm not sure of the issues with that - there is so much going on in `options.initial`. So I thought it would be safer to just allow setting a custom start date when `initial: true` is set

test/unit/redux/actions/async.test.js
```diff
@@ -4210,6 +4210,14 @@ describe('Actions', () => {
             ...initialState,
           }, router: { location: { pathname: `data/${patientId}` } } });
 
+          const options = {
+            // startDate: undefined,
+            // endDate: undefined,
+            returnData: false,
+            useCache: true,
+            initial: true,
+          };
```
henry-tp: Explanation: A few lines below, `async.fetchPatientData(api, options, patientId)` is being called. That `options` argument is defined by the `beforeEach()` block at the top:

```jsx
        options = {
          startDate: '2018-01-01T00:00:00.000Z',
          endDate: '2018-01-30T00:00:00.000Z',
          returnData: false,
          useCache: true,
          initial: true,
        }
```

Prior to this PR, the code in `async.fetchPatientData()` was simply ignoring the fact that `startDate` was specified in the in the options, overriding it with the default (30 days). Now, with the change to using `||=` in the code, it is actually registering and using the `startDate` in the options instead of ignoring/overriding it, which is why this test is breaking. 

By setting the dates to `undefined` allows the default to once again take over.

I would expect that this is wanted behaviour (NOT ignoring the startDate if it is specified in the options) but let me know if this causes problems.

app/pages/dashboard/PatientDrawer/useAgpCGM/useAgpCGM.js
```diff
@@ -51,7 +52,20 @@ const inferLastCompletedStep = (patientId, data, pdf) => {
...
+  let daysToFetch = agpPeriodInDays * 2; // Fetch enough data for current AND past-period AGP (ie. double)
+  daysToFetch += safetyFactorDays;       // Request a few extra days in case of timezone mismatch.
+
+  return {
+    initial: true,
+    startDate: moment.utc().subtract(daysToFetch, 'days').toISOString(),
+    forceDataWorkerAddDataRequest: true,
+    useCache: false,
+  };
+};
```
henry-tp, line 69: The bug is that in the 30 day AGP, the "Changes Since" component would only consist of 2 days of data. The 7- and 14-day versions had no issues. I originally thought that this was because test patient accounts only populated the last 32-ish days, but I was mistaken.

image

The key issue was that I missed, in my initial implementation, that we make a call to `this.fetchAdditionalData()` when we are trying to generate an AGP from a previous time period: 

https://github.com/tidepool-org/blip/blob/develop/app/pages/patientdata/patientdata.js#L419-L429

It turns out I was relying on the default of querying for the `last 30 days of data` when calling `actions.async.fetchPatientData()`: 

https://github.com/tidepool-org/blip/blob/develop/app/redux/actions/async.js#L1122

This is why 7- and 14-day AGP worked (requiring the last 14 and 28 days of data respectively) but the 30-day AGP doesn't (requiring the last 60 days of data)

The fix to this is to query for the necessary amount of data. It works now:

image

The great thing is that I can create the past-period AGP using the patient data print, and verify that the numbers do line up, which is a relief

image

app/components/clinic/PatientForm/PatientForm.js
```diff
@@ -117,7 +117,7 @@ export const PatientForm = (props) => {
           clinic: {
             handler: 'acceptPatientInvitation',
             args: () => [selectedClinicId, invite.key, invite.creatorId, omitBy(
-              pick(getFormValues(values, clinicPatientTags, clinicSites), ['mrn', 'birthDate', 'fullName', 'tags']),
+              pick(getFormValues(values, clinicPatientTags, clinicSites), ['mrn', 'birthDate', 'fullName', 'tags', 'sites']),
```
henry-tp, line 120: bug prevented sites from being set when accepting invite

app/redux/actions/async.js
krystophv: This will probably be harmless, in general, but I'm wondering what issues you were running into that you needed to use `initial: true` to resolve for the additional data fetching for the PDF generation.

henry-tp: Honestly, I was unsure either way, whether to have `initial` be `true` or `false`. 

In truth, the pre-fetch for `serverTime` seemed quite odd to me and I was kind of unsure about what prior problems would have led us to need to do that. I looked at [WEB-810](https://tidepool.atlassian.net/browse/WEB-810) and [WEB-803](https://tidepool.atlassian.net/browse/WEB-803) and was even more confused. I was worried about missing some edge cases or that logic being "written in blood" so to speak. For that reason, I set `initial` to be true, since opening the drawer would always be the first time we fetch data for a patient, it made sense.

After double-checking with Claude about the difference between initial being `true` or `false`, I don't think there are any gotchas and I'm not terribly convinced that the server time pre-fetch would be a major point of contention. https://claude.ai/share/190b083a-4f6a-4ccd-9ed0-b1e6092271cc

Given this, let's revert this change (to prevent any other regressions due to switching `=` to `||=`) and make it so that useAgpCGM calls `fetchPatientData` with `initial: false`. I added an `endDate` as it is necessary for calls with `initial: false`

a61667d03

henry-tp: Now that I really think about it, I don't _think_ that future datums would impact the AGP since the AGP Util excludes data outside the defined endpoints. So that shouldn't be a concern

test/unit/redux/actions/async.test.js
henry-tp: Reverting back changes that caused this new object to be neeed

PR #1715, opened 2025-07-30
[WEB-3771] Informed consent flow updates

clintonium-119:
[WEB-3771]

Related PR: https://github.com/tidepool-org/platform-client/pull/207

Main PR details:
 - moves data donation sectionin new patient flow to it's own 2-stage page
   - new UI for selecting supported non-profits
   - requires consent document to be viewed and agreed with, and consent record to be stored via new API enpoints 
   - Text in UI differs by account type and age in numerous places (helpful to refer to [figma](https://www.figma.com/design/CTkPBVhZkDaD3YZGDW0pjR/Informed-Consent-V1?node-id=4901-1426&t=fEGHIb27oy21cFjr-0) to see this)
 - replaces the old method of tracking data donation accounts via shared accounts, and it's associated redux state, with new API endpoints and redux state for consent documents and records 
 - new donation UI also added to patient profile page
 - donate banner now redirects to patient profile page and scrolls to appropriate section

[WEB-3771]: https://tidepool.atlassian.net/browse/WEB-3771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/unit/app/providers/AppBanner/AppBannerProvider.test.js
+++ b/test/unit/app/providers/AppBanner/AppBannerProvider.test.js
@@ -323,12 +323,7 @@ describe('AppBannerProvider', () => {
         data: {
           metaData: { size: 1, devices: [] },
         },
-        consentRecords: {
-          [DATA_DONATION_CONSENT_TYPE]: {
-            status: 'revoked',
-            metadata: { supportedOrganizations: [] },
-          },
-        },
+        consentRecords: {},
       },
     };
 
--- a/test/unit/pages/ClinicianPatients.test.js
+++ b/test/unit/pages/ClinicianPatients.test.js
@@ -153,11 +153,11 @@ describe('ClinicianPatients', () => {
 
   context('patients hidden', () => {
     beforeEach(() => {
-      const initialState = { 
-        blip: { 
+      const initialState = {
+        blip: {
           ...hasPatientsState.blip,
           patientListFilters: { isPatientListVisible: false, patientListSearchTextInput: '' }
-        } 
+        }
       }
 
       store = mockStore(initialState);
@@ -173,7 +173,7 @@ describe('ClinicianPatients', () => {
       defaultProps.trackMetric.resetHistory();
     });
 
-    it('should render a button that toggles patients to be visible', () => { 
+    it('should render a button that toggles patients to be visible', () => {
       wrapper.find('.peopletable-names-showall').hostNodes().simulate('click');
       expect(store.getActions()).to.eql([{ type: 'SET_IS_PATIENT_LIST_VISIBLE', payload: { isVisible: true } }])
     })
--- a/test/unit/pages/patientprofile.test.js
+++ b/test/unit/pages/patientprofile.test.js
@@ -47,13 +47,11 @@ describe('PatientProfile', () => {
       expect(result[0]()).to.equal('fetchPatient');
       expect(result[1]).to.be.a('function');
       expect(result[1]()).to.equal('fetchAssociatedAccounts');
-      expect(result[2]).to.be.a('function');
-      expect(result[2]()).to.equal('fetchUserConsentRecords');
     });
 
     it('should only add the associated accounts, consents, and pending invites fetchers if fetches are not already in progress or completed', () => {
       const standardResult = getFetchers(dispatchProps, ownProps, stateProps, api);
-      expect(standardResult.length).to.equal(3);
+      expect(standardResult.length).to.equal(2);
 
       const inProgressResult = getFetchers(dispatchProps, ownProps, {
         user: { userid: '12345' },
@@ -88,7 +86,7 @@ describe('PatientProfile', () => {
 
     it('should only add the associated accounts fetcher when viewing the profile of the logged-in user', () => {
       const standardResult = getFetchers(dispatchProps, ownProps, stateProps, api);
-      expect(standardResult.length).to.equal(3);
+      expect(standardResult.length).to.equal(2);
 
       const loggedInUserResult = getFetchers(dispatchProps, {
         match: {
@@ -106,12 +104,12 @@ describe('PatientProfile', () => {
         },
       }, dispatchProps, api);
 
-      expect(loggedInUserResult.length).to.equal(3);
+      expect(loggedInUserResult.length).to.equal(2);
     });
 
     it('should only fetch the clinic patient if selectedClinicId is set, and the patient permissions are unavailable', () => {
       const standardResult = getFetchers(dispatchProps, ownProps, stateProps, api);
-      expect(standardResult.length).to.equal(3);
+      expect(standardResult.length).to.equal(2);
 
       const clinicUserResult = getFetchers(dispatchProps, {
         match: {
@@ -123,7 +121,7 @@ describe('PatientProfile', () => {
         clinics: { clinic123: { patients: { '12345': { name: 'Jackie', permissions: { foo: 'bar' } } } } }
       }, dispatchProps, api);
 
-      expect(clinicUserResult.length).to.equal(3);
+      expect(clinicUserResult.length).to.equal(2);
 
       const clinicUserMissingPermissionsResult = getFetchers(dispatchProps, {
         match: {
@@ -135,9 +133,9 @@ describe('PatientProfile', () => {
         clinics: { clinic123: { patients: { '12345': { name: 'Jackie' } } } }
       }, dispatchProps, api);
 
-      expect(clinicUserMissingPermissionsResult.length).to.equal(4);
-      expect(clinicUserMissingPermissionsResult[3]).to.be.a('function');
-      expect(clinicUserMissingPermissionsResult[3]()).to.equal('fetchPatientFromClinic');
+      expect(clinicUserMissingPermissionsResult.length).to.equal(3);
+      expect(clinicUserMissingPermissionsResult[2]).to.be.a('function');
+      expect(clinicUserMissingPermissionsResult[2]()).to.equal('fetchPatientFromClinic');
     });
   });
 
--- a/test/unit/redux/reducers/consentRecords.test.js
+++ b/test/unit/redux/reducers/consentRecords.test.js
@@ -12,15 +12,24 @@ var expect = chai.expect;
 
 describe('consentRecords', () => {
   describe('fetchUserConsentRecordsSuccess', () => {
-    it('should set latest consentRecord by type to state', () => {
+    it('should set latest consentRecord by type to state if status is active', () => {
       const initialStateForTest = {};
       const consentType = 'consentType123';
-      const records = { data: [{ id: 'record123' }] };
+      const records = { data: [{ id: 'record123', status: 'active' }] };
       const action = actions.sync.fetchUserConsentRecordsSuccess(consentType, records);
       const state = reducer(initialStateForTest, action);
       expect(state).to.eql({
-        consentType123: { id: 'record123' },
+        consentType123: { id: 'record123', status: 'active' },
       });
+    });
+
+    it('should not set latest consentRecord by type to state if status is revoked', () => {
+      const initialStateForTest = {};
+      const consentType = 'consentType123';
+      const records = { data: [{ id: 'record123', status: 'revoked' }] };
+      const action = actions.sync.fetchUserConsentRecordsSuccess(consentType, records);
+      const state = reducer(initialStateForTest, action);
+      expect(state).to.eql({});
     });
   });
 


__tests__/unit/app/pages/patientnew/dataDonation.test.js
```diff
@@ -0,0 +1,215 @@
+/* global jest, test, expect, describe, it */
```
clintonium-119, line 1: I've kept the testing fairly high-level here, and on the patient profile page, just making sure that the expected elements are present in the 2 stages (non-consented and consented states), and that any buttons or links do what they need to do (opening dialogs, redirecting).

The main logic of the DonationForm, creating and revoking consents for the various account types and ages, will be tested more thoroughly in it's own test file.

app/components/elements/SlideShow/SlideShow.js
```diff
@@ -0,0 +1,262 @@
+import React, { useEffect } from 'react';
+import PropTypes from 'prop-types';
+import { Box, BoxProps, Flex, Image } from 'theme-ui';
+import map from 'lodash/map';
+import { useTranslation } from 'react-i18next';
+import { useSnapCarousel } from 'react-snap-carousel';
```
clintonium-119, line 6: There were a couple other carousel libraries that met our needs of being lightweight, responsive, accessible, and supporting css-based snappoint dragging, while being a bit higher level with predefined components, but this was the only one that supported React 16.

This one works well though.  Just a bit more verbose/low-level than I'd wanted.

app/pages/patient/DataDonationConsentDialog.js
```diff
@@ -0,0 +1,362 @@
...
+import i18next from '../../core/language';
+import { colors, shadows } from '../../themes/baseTheme';
+import personUtils from '../../core/personutils';
+import Pill from '../../components/elements/Pill';
+
+import { DATA_DONATION_CONSENT_TYPE } from '../../core/constants';
+
+const t = i18next.t.bind(i18next);
+
+export const getConsentText = memoize((accountType, patientAgeGroup, patientName, caregiverName, consentDate) => {
```
clintonium-119, line 34: This is the main method to create the various bits of text content based on account type and patient age.  I memoized it since it does get called from a few places, and the arguments rarely change.

app/pages/patient/DataDonationConsentDialog.js
```diff
@@ -0,0 +1,362 @@
...
+        document.getElementById('consentDocumentText')?.scrollTo({ top: 0 });
+        setScrolledToBottom(false);
+      } else {
+        onConfirm({ ...values, name: caregiverName });
+      }
+    },
+  });
+
+  useEffect(() => {
+    if (isString(formikContext.values.name)) setCaregiverName(formikContext.values.name);
```
clintonium-119: This is set to state so it can be used to update the caregiver name in the consent signature for personal accounts of youth age, as it gets provided by a text input dynamically for this scenario, since it won't be available in the stored account settings (we only collect caregiver name for 'caregiver' accounts, not personal ones.

app/pages/patient/DataDonationConsentDialog.js
```diff
@@ -0,0 +1,362 @@
...
+  const consentDescription = consentText[`${formSteps[currentConsentStep]}ConsentDescription`];
+  const consentInputLabel = consentText[`${formSteps[currentConsentStep]}ConsentInputLabel`];
+  const consentReviewMessage = consentText[`${formSteps[currentConsentStep]}ConsentReviewMessage`];
+  const consentSignature = consentText[`${formSteps[currentConsentStep]}ConsentSignature`];
+  const consentNameInputLabel = consentText[`${formSteps[currentConsentStep]}ConsentNameInputLabel`];
+
+  const handleConsentDocumentScroll = (e) => {
+    const { scrollTop, scrollHeight, clientHeight } = e.target;
+    // Use a small threshold (5px) to account for rendering differences
+    if (scrollTop + clientHeight >= scrollHeight - 5) {
```
clintonium-119, line 161: User must scroll through consent document before they can check the consent checkbox.

app/pages/patient/DataDonationConsentDialog.js
```diff
@@ -0,0 +1,362 @@
...
+}, (accountType, patientAgeGroup, patientName, caregiverName, consentDate) =>
+  `${accountType}-${patientAgeGroup}-${patientName}-${caregiverName}-${consentDate}`
+);
+
+export const DataDonationConsentDialog = (props) => {
+  const { t, onClose, onConfirm, open, accountType, patientAgeGroup, patientName, caregiverName: caregiverNameProp, consentDate } = props;
+  const patientAssentRequired = patientAgeGroup === 'youth';
+  const formSteps = patientAssentRequired ? ['primary', 'secondary'] : ['primary'];
+  const [currentConsentStep, setCurrentConsentStep] = React.useState(0);
+  const [scrolledToBottom, setScrolledToBottom] = React.useState(false);
```
henry-tp: I think we can remove this piece of state

app/pages/patient/DataDonationConsentDialog.js
```diff
@@ -0,0 +1,362 @@
...
+
+          {consentNameInputLabel && (
+            <Box mb={3}>
+              <TextInput
+                {...getCommonFormikFieldProps('name', formikContext)}
+                label={consentNameInputLabel}
+                placeholder={t('Name')}
+                variant="condensed"
+                width="100%"
+              />
```
henry-tp, line 280: Can we eliminate the need to have a `useEffect`  by instead setting the `caregiverName` directly from here?

```jsx
            <Box mb={3}>
              <TextInput
                {...getCommonFormikFieldProps('name', formikContext)}
                onChange={(e) => {
                  if (!e?.target) return;

                  formikContext.handleChange(e);
                  setCaregiverName(e.target.value);
                }}
                label={consentNameInputLabel}
                placeholder={t('Name')}
                variant="condensed"
                width="100%"
              />
            </Box>
```

app/pages/patient/DataDonationForm.js
```diff
@@ -0,0 +1,371 @@
...
+  }, [loggedInUserId]); // eslint-disable-line react-hooks/exhaustive-deps
+
+  useEffect(() => {
+    if (currentConsent?.status === 'active' && currentForm === formSteps.dataDonationConsent) {
+      setCurrentForm(formSteps.supportedOrganizations);
+    }
+    if (currentConsent?.status !== 'active' && currentForm === formSteps.supportedOrganizations) {
+      setCurrentForm(formSteps.dataDonationConsent);
+    }
+  }, [currentConsent]); // eslint-disable-line react-hooks/exhaustive-deps
```
henry-tp: It seems like `currentForm` is Derived State (derivable from `currentConsent`). If so, could we eliminate this effect and also the `currentForm` piece of state?

So that would look like:

```jsx
// remove the useState() and useEffect(), then

const currentForm = (
  currentConsent?.status === 'active'
    ? formSteps.supportedOrganizations
    : formSteps.dataDonationConsent
  );
```

app/pages/patient/DataDonationForm.js
```diff
@@ -0,0 +1,371 @@
...
+  useEffect(() => {
+    if (!formikContext.touched.supportedOrganizations && currentConsent?.metadata?.supportedOrganizations) {
+      const supportedOrganizations = currentConsent?.metadata?.supportedOrganizations.join(',');
+
+      if (isEmpty(formikContext.values.supportedOrganizations) && !isEmpty(supportedOrganizations)) {
+        formikContext.setFieldValue('supportedOrganizations', supportedOrganizations);
+        formikContext.setFieldTouched('supportedOrganizations', true);
+      }
+    }
+  }, [currentConsent?.metadata?.supportedOrganizations]); // eslint-disable-line react-hooks/exhaustive-deps
```
henry-tp, line 183: [For future consideration, no action required]

This is a pretty standard pattern in our codebase. One reason I'm not the biggest fan of using effects for "catching" API call responses is because it can be hard to trace the cause and effect of when an Effect is expected to fire. In this case, this is expected to fire in response to the `fetchUserConsentRecords` response, but for me, it's a little challenging to figure out just by reading the function.

What I'd like to propose is moving us in a direction of tying the API call to the state setters that should be triggered as a result of that call succeeding. Using our existing patterns with callbacks, that might look like

```jsx
// in app/redux/actions/async.js

export function fetchUserConsentRecords(api, consentType, onSuccess = noop) {
  return (dispatch, getState) => {
    const { blip: { loggedInUserId } } = getState();

    dispatch(sync.fetchUserConsentRecordsRequest());

    api.consent.getUserConsentRecords(loggedInUserId, consentType, (err, records) => {
      if (err) {
        dispatch(sync.fetchUserConsentRecordsFailure (
          createActionError(ErrorMessages.ERR_FETCHING_USER_CONSENT_RECORDS, err), err
        ));
      } else {
        dispatch(sync.fetchUserConsentRecordsSuccess(consentType, records));

/* -> */ onSuccess(consentType, records) /* <- */

      }
    });
  };
}

// in app/pages/patient/DataDonationForm.js

const handleFetchUserConsentRecordsSuccess = () => {
  if (!formikContext.touched.supportedOrganizations && currentConsent?.metadata?.supportedOrganizations) {
    const supportedOrganizations = currentConsent?.metadata?.supportedOrganizations.join(',');

    if (isEmpty(formikContext.values.supportedOrganizations) && !isEmpty(supportedOrganizations)) {
      formikContext.setFieldValue('supportedOrganizations', supportedOrganizations);
      formikContext.setFieldTouched('supportedOrganizations', true);
    }
  }
};

useEffect(() => {
  if (loggedInUserId) {
    dispatch(
      actions.async.fetchUserConsentRecords(api, DATA_DONATION_CONSENT_TYPE, handleFetchUserConsentRecordsSuccess)
    );
  }
}, [loggedInUserId]); // eslint-disable-line react-hooks/exhaustive-deps

```

This would help us avoid needing constructs like `workingState` + `handleAsyncResult`, and in big files it would help make the code more parse-able.

If it were _fully_ up to me, I'd really like to have the async actions return promises, allowing the fetch function to return the data directly.

This is just for consideration, maybe we can talk about it in our frontend huddle ;)

app/redux/reducers/misc.js
```diff
@@ -589,6 +557,46 @@ export const devices = (state = initialState.devices, action) => {
...
+export const consentRecords = (state = initialState.consentRecords, action) => {
+  switch (action.type) {
+    case types.FETCH_USER_CONSENT_RECORDS_SUCCESS: {
+      const consentType = _.get(action.payload, 'consentType');
+      const latestRecord = _.get(action.payload, 'records.data.0');
+      return update(state, { $set: { [consentType]: latestRecord } });
+    }
+
+    case types.CREATE_USER_CONSENT_RECORD_SUCCESS: {
+      const createdRecord = _.get(action.payload, 'createdRecord');
```
henry-tp, line 587: https://github.com/user-attachments/assets/47cc6400-fa7e-4476-972e-1a55960b8dc2

A bit of an edge case, but we may need to clear out the FE consent state on creation in case there are stale ones stuck in frontend state.

app/pages/patientnew/dataDonation.js
```diff
@@ -0,0 +1,96 @@
...
+
+  useEffect(() => {
+    if (trackMetric) {
+      trackMetric('Viewed Profile Create');
+    }
+  }, []);
+
+  const loggedInUserId = useSelector((state) => state.blip.loggedInUserId);
+  const [submitting, setSubmitting] = useState(false);
+  const [currentForm, setCurrentForm] = useState(formSteps.dataDonationConsent);
```
henry-tp: There is an edge case here to be aware of. Steps to repro are:

- Create new patient account
- Get to `supportedOrganizations` form by creating a `consent`
- Hard refresh the page
- Re-start at `dataDonationConsent`. That's fine, fill it out again
- Network call fails, not able to view `supportedOrganizations`

https://github.com/user-attachments/assets/cf170e8f-9337-4d29-837e-fb6341ca94a8

I think the issue is that `consents` and `consentRecords` are out of sync in Redux - `consents` is fetched but `consentRecords` is not fetched.

I think that we should fetch `consentRecords` anywhere we fetch `consents`. Maybe we could do both within `<DataDonationForm />`.

This way, this `PatientNewDataDonation` component could also derive the `currentForm` value from redux, ie. 

```
state.blip.consentRecords[DATA_DONATION_CONSENT_TYPE]?.status === 'active' ? formSteps.supportedOrganizations : formSteps.dataDonationConsent
``` 

instead of relying on state to keep track of the form position. This would help prevent synchronization errors like this. So the DataDonationForm 

what do you think?

app/routes.js
```diff
@@ -454,8 +455,9 @@ export const getRoutes = (appContext) => {
           <Route path='/profile' render={routeProps => (<Gate onEnter={boundRequireAuth} key={routeProps.match.path}><UserProfile {...routeProps} {...props} /></Gate>)} />
           <Route exact path='/patients' render={routeProps => (<Gate onEnter={boundRequireAuth} key={routeProps.match.path}><Patients {...routeProps} {...props} /></Gate>)} />
           <Route exact path='/patients/new' render={routeProps => (<Gate onEnter={boundRequireAuthAndNoPatient} key={routeProps.match.path}><PatientNew {...routeProps} {...props} /></Gate>)} />
-          <Route exact path='/prescriptions/new' render={routeProps => (<Gate onEnter={boundRequireAuth} key={routeProps.match.path}><PrescriptionForm {...routeProps} {...props} /></Gate>)} />}
-          <Route exact path='/prescriptions/:id' render={routeProps => (<Gate onEnter={boundRequireAuth} key={routeProps.match.path}><PrescriptionForm {...routeProps} {...props} /></Gate>)} />}
```
henry-tp, line 458: [No Action Needed] Huh, that's weird. Glad it didn't blow up our application before

henry-tp (changes requested):
Some comments

app/pages/patient/DataDonationForm.js
```diff
@@ -0,0 +1,371 @@
...
+  const isYouth = patientAge < 18;
+  const isChild = patientAge < 13;
+  const accountType = personUtils.patientIsOtherPerson(user) ? 'caregiver' : 'personal';
+  const patientAgeGroup = isChild ? 'child' : (isYouth ? 'youth' : 'adult');
+  const patientName = personUtils.patientFullName(user);
+  const caregiverName = accountType === 'caregiver' ? personUtils.fullName(user) : undefined;
+  const { [DATA_DONATION_CONSENT_TYPE]: consentDocument } = useSelector((state) => state.blip.consents);
+  const currentConsent = useSelector(state => state.blip.consentRecords[DATA_DONATION_CONSENT_TYPE]);
+  const fallbackConsentDate = formContext === formContexts.newPatient ? moment().format('MMMM D, YYYY') : null;
+  const consentDate = currentConsent?.grantTime ? moment(currentConsent.grantTime).format('MMMM D, YYYY') : fallbackConsentDate;
```
henry-tp, line 80: This comes up as `null` for me a lot.

For example, create a new patient, immediately navigate to the `Profile` page, click `Yes, I'm interested`

image

Should we default to current date?

app/pages/patient/DataDonationConsentDialog.js
clintonium-119: Good catch.  Removing

app/pages/patient/DataDonationForm.js
clintonium-119: Yeah, that's a good idea.  Will do.

app/pages/patient/DataDonationForm.js
clintonium-119: Good topic for discussion at our "next huddle" :laughing:  (we really should set some time aside for those)

I'd need to think about the ramifications of adopting this approach a bit more, vs, say adopting redux toolkit.  Our current method is cumbersome, to be sure, and it would be good to move onto something that provides more clarity (and way less boilerplate)

app/pages/patient/DataDonationConsentDialog.js
clintonium-119: Yep.  Love this change.  Good catch.

app/routes.js
clintonium-119: I think it's all good.  Just strange how GH choosed to show the diff.  The prescription routes are still intact.

app/pages/patientnew/dataDonation.js
clintonium-119: Good call.  I actually noticed the same edge case before signing off yesterday, so I appreciate you diving in and coming up with a solution. 

I also realized that I no longer needed the updated formik context passed from the `DataDonationForm` component into the dataDonation page (there was some cruft from the old approach, a checkbox, that I stripped out when I was revisiting the code to write tests, but hadn't noticed that I no longer needed the formikContext in here at all.

I've added the fetch for `consentRecords` (locally, not yet pushed, but will be) to `DataDonationForm` as suggested, and removed it from the patientprofile.js wrapper.  I'm still fetching it from the `patientData` view, though, since we need them fetched to determine banner display, and they likely will not have loaded the `DataDonationForm` component when they land on that view.

app/redux/reducers/misc.js
clintonium-119: Good catch.  I'm now calling `formikContext.resetForm` after consent revocation.  I could do it after creation, but really, once the consent has been revoked, the form context should be reset (even though the supported organizations field will not be seen again until after a new consent creation)

app/pages/patient/DataDonationForm.js
clintonium-119: Interesting.  This null scenario went unnoticed for me because we get revoked consents sent as well as active ones when we fetch the user's consent records, and so I was typically (unknowingly) seeing the grantTime of the revoked consent records.

I've altered the redux store to only store active consent records.  Now I'm seeing the issue each time, and I'll come up with a fix.

There was a point during development when I had a reason to have a null value as a potential fallback (it was addressing some other edge case, but I can't recall what it was).  There was a lot of code shift once I actually was able to work with the real API endpoints, and so it may not be needed at all.

Originally, I always used the current date as a fallback, and hopefully, I can revert to that.

clintonium-119: Ah yes, I remember it now.  Up until the very end of development, I was having to support the legacy account-sharing method for data donation accounts.  For previous shares like this, there was not a consent date available, and we needed a null fallback.

When the decision was made to migrate all legacy shares to the new consent record model on the backend, I had cleaned up most of the supporting code, but missed this aspect.

So we're fine to go back to always defaulting to the current date.

stories/SlideShow.stories.js
```diff
@@ -0,0 +1,67 @@
+import React from 'react';
+import { Box, ThemeUIProvider } from 'theme-ui';
+
+import baseTheme from '../app/themes/baseTheme';
+import ChangeMindImage from '../app/components/elements/SlideShow/Images/changeMindImage.svg';
+import FuelingNextGenImage from '../app/components/elements/SlideShow/Images/fuelingNextGenImage.svg';
```
henry-tp, line 6: I think these SVGs can be optimized

app/redux/reducers/misc.js
```diff
@@ -574,7 +574,13 @@ export const consentRecords = (state = initialState.consentRecords, action) => {
     case types.FETCH_USER_CONSENT_RECORDS_SUCCESS: {
       const consentType = _.get(action.payload, 'consentType');
       const latestRecord = _.get(action.payload, 'records.data.0');
-      return update(state, { $set: { [consentType]: latestRecord } });
+
+      // Only store active records so we don't have to litter our application with checks to see if the current consent record is active
+      if (latestRecord.status === 'active') {
```
henry-tp: On new patient creation, I hit a bug here because `latestRecord` is `undefined`, just need a null check

henry-tp (approved):
Looks great!! Well done on this feature

Approved, but please fix:

- One critical error and needing a null check
- Tests passing of course

app/redux/reducers/misc.js
clintonium-119: Thanks.  Fixed.

stories/SlideShow.stories.js
clintonium-119: Done

PR #1720, opened 2025-08-04
WEB-3877 - Loading issue on TIDE Dashboard

henry-tp:
WEB-3877

Fixes a loading bug in TIDE Dashboard

See https://tidepoolteam.slack.com/archives/CF59ADWLB/p1754325316767959?thread_ts=1753897758.073649&cid=CF59ADWLB for issue discussion

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "@testing-library/react": "12.1.5",
     "@testing-library/react-hooks": "8.0.1",
     "@testing-library/user-event": "14.6.1",
-    "@tidepool/viz": "1.50.0-rc.1",
+    "@tidepool/viz": "1.50.0-rc.2",
     "async": "2.6.4",
     "autoprefixer": "10.4.16",
     "babel-core": "7.0.0-bridge.0",
--- a/yarn.lock
+++ b/yarn.lock
@@ -6083,9 +6083,9 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@tidepool/viz@npm:1.50.0-rc.1":
-  version: 1.50.0-rc.1
-  resolution: "@tidepool/viz@npm:1.50.0-rc.1"
+"@tidepool/viz@npm:1.50.0-rc.2":
+  version: 1.50.0-rc.2
+  resolution: "@tidepool/viz@npm:1.50.0-rc.2"
   dependencies:
     bluebird: 3.7.2
     bows: 1.7.2
@@ -6146,7 +6146,7 @@ __metadata:
     react-dom: 16.x
     react-redux: 8.x
     redux: 4.x
-  checksum: 546909935849c114a23a92a71ce8cab570cd93be8b8f07956e8765231499e9ceae9f2d30567e572f5b3e1bfb4358c441b01b1765717ea6426c8c2904312eb741
+  checksum: 6da5dd1c38751d22af0e3f6c7a330a108751fff606e0449ee0c59a8cf477f970c1c02d798ef62d67b84c520435d006c3f9f6637702ac1010de9dc0a9e23757e6
   languageName: node
   linkType: hard
 
@@ -8335,7 +8335,7 @@ __metadata:
     "@testing-library/react": 12.1.5
     "@testing-library/react-hooks": 8.0.1
     "@testing-library/user-event": 14.6.1
-    "@tidepool/viz": 1.50.0-rc.1
+    "@tidepool/viz": 1.50.0-rc.2
     async: 2.6.4
     autoprefixer: 10.4.16
     babel-core: 7.0.0-bridge.0


app/pages/dashboard/TideDashboard.js
```diff
@@ -937,6 +937,8 @@ export const TideDashboard = (props) => {
   }
 
   useEffect(() => {
+    if (!showTideDashboard) return;
+
```
henry-tp, line 943: See [slack discussion](https://tidepoolteam.slack.com/archives/CF59ADWLB/p1754325316767959?thread_ts=1753897758.073649&cid=CF59ADWLB) of issue

So the issue is that we were usually fetching TIDE dashboard patients twice on every dashboard load. The cause was that `showTideDashboard` initializes to `false`, and then it changes to `true`, causing this effect to fire twice and thereby the API call to fire twice. You can see the duplicate redux action for `FETCH_TIDE_DASHBOARD_PATIENTS_SUCCESS` here:

image

I believe there were some race conditions that caused, in some circumstances, the some working states to get stuck

You can see it in how we would occassionally load the data, then it would go back to loading state, causing this "double load" phenomenon

https://github.com/user-attachments/assets/58810157-c1b5-43dc-b06f-0f00a700b42a

app/redux/actions/async.js
```diff
@@ -2933,6 +2933,7 @@ export function deleteClinicPatientTag(api, clinicId, patientTagId) {
       } else {
         if (results.config) results.config.lastData = options.lastData;
         dispatch(sync.fetchTideDashboardPatientsSuccess(results));
+        onSuccess();
```
henry-tp, line 2936: There was a small issue where `handleAsyncResult` did not always call `setLoading(false)`

app/pages/dashboard/TideDashboard.js
```diff
@@ -937,17 +939,21 @@ export const TideDashboard = (props) => {
...
 
   useEffect(() => {
+    if (!isTideDashboardEnabled) return;
+
     if (validateTideConfig(localConfig?.[localConfigKey], patientTags)) {
       fetchDashboardPatients();
     } else {
       setShowTideDashboardConfigDialog(true);
     }
+  }, [isTideDashboardEnabled]);
```
henry-tp, line 949: According to [this commit](https://github.com/tidepool-org/blip/commit/b4502957d64aadd66b4a9e3484da106107f7f42e), we had issues before with trying to fetch patient before ldContext is loaded. So this variable `isTideDashboardEnabled` allows us to ensure that A) ldContext is loaded and B) the logic of the effect is only run once.

app/pages/dashboard/TideDashboard.js
```diff
@@ -937,17 +939,21 @@ export const TideDashboard = (props) => {
...
 
+  useEffect(() => {
     // Always clear stored dashboard results upon unmount to avoid flashing stale results upon remount
     return () => {
       dispatch(actions.sync.clearTideDashboardPatients());
-    }
-  }, [showTideDashboard]);
+    };
+  }, []);
```
henry-tp, line 956: moved this cleanup to its own effect to ensure that it is not impacted by the early return I introduced earlier

app/pages/dashboard/TideDashboard.js
```diff
@@ -796,6 +796,8 @@ export const TideDashboard = (props) => {
   const ldClient = useLDClient();
   const ldContext = ldClient.getContext();
 
+  const isTideDashboardEnabled = ldContext?.clinic?.tier && (clinic?.entitlements?.tideDashboard || showTideDashboard);
```
clintonium-119: I'm wondering about this.  Would this change still work to fix the issue?
```suggestion
  const isTideDashboardEnabled = (ldContext?.clinic?.tier && showTideDashboard) || clinic?.entitlements?.tideDashboard;
```

My thinking is that this way, if we have trouble connecting to LD, but the clinic tier provides an entitlement that enables the dashboard, we would still be able to show the dashboard.

Otherwise, in the case of LD network errors, no users could see the dashboard, even if their clinic tier (or clinic entitlements, if we eventually set these directly on the clinic objects on the backend as has been talked about) would allow for it.

Thoughs?

clintonium-119 (commented):
Nice job digging into this.  I do have a question in the comments about `isTideDashboardEnabled`, though.

test/unit/pages/TideDashboard.test.js
```diff
@@ -400,6 +400,10 @@ describe('TideDashboard', () => {
       TideDashboard.__Rewire__('useLocalStorage', useLocalStorageRewire(mockedLocalStorage));
       TideDashboardConfigForm.__Rewire__('useLocalStorage', useLocalStorageRewire(mockedLocalStorage));
 
+      TideDashboard.__Rewire__('useLDClient', sinon.stub().returns(new LDClientMock({ clinic : {
+        tier: 'tier0300',
+      }})));
```
henry-tp, line 405: This shouldn't matter. These 3 tests were failing because the TIDE Dashboard Config Dialog wasn't being shown now that I have my early return. However, it only makes sense to show that config dialog if the clinic _should_ have access to TIDE dashboard anyway.

PR #1729, opened 2025-08-15
WEB-3832 - Notify Banner

henry-tp:


app/pages/patientdata/patientdata.js
```diff
@@ -2354,7 +2354,9 @@ export function getFetchers(dispatchProps, ownProps, stateProps, api, options) {
     fetchers.push(dispatchProps.fetchPendingSentInvites.bind(null, api));
   }
 
-  if (stateProps.isUserPatient && !stateProps.fetchingClinicsForPatient.inProgress && !stateProps.fetchingClinicsForPatient.completed) {
+  const isUserPatient = ownProps.match.params.id === stateProps.user?.userid;
```
henry-tp, line 2357: I believe a small bug was introduced in #1597

When a patient first logs in, at the moment this `getFetchers` function is run, the `currentPatientInViewId` and `patient` variables in redux are `null`, and the `fetchClinicsForPatient` function is never called even when it should be.

You can verify this by opening up network tab FIRST and then logging in as the patient, noticing that we never make a request to the `/patients/:patientId/clinics` endpoint.

image

It's tricky to catch because after you are already logged in as the patient, even if you do a hard browser refresh, we WILL make a request to the endpoint. 

image

So this change ensures that we always ping the endpoint, even on first login

clintonium-119: Good catch

clintonium-119: And, funny enough, I just ran into this when trying to fetch consent info for the currently logged-in patient on this view, so it was nice to know the solution right away :)

henry-tp:
Closing for #1730

henry-tp:
Switching to solution in #1730

PR #1730, opened 2025-08-22
WEB-3832 - Notify Banner V2

henry-tp:
[WEB-3832]

[WEB-3832]: https://tidepool.atlassian.net/browse/WEB-3832?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/pages/patientdata/patientdata.js
```diff
@@ -2354,7 +2354,9 @@ export function getFetchers(dispatchProps, ownProps, stateProps, api, options) {
     fetchers.push(dispatchProps.fetchPendingSentInvites.bind(null, api));
   }
 
-  if (stateProps.isUserPatient && !stateProps.fetchingClinicsForPatient.inProgress && !stateProps.fetchingClinicsForPatient.completed) {
+  const isUserPatient = ownProps.match.params.id === stateProps.user?.userid;
```
henry-tp: See comment: https://github.com/tidepool-org/blip/pull/1729#discussion_r2288801529

__tests__/unit/app/pages/patient/patientsettings.test.js
```diff
@@ -0,0 +1,232 @@
+/* global jest */
+/* global expect */
+/* global describe */
+/* global afterEach */
+/* global it */
+/* global beforeEach */
+
```
henry-tp, line 7: migrated test file. Scrapped some older tests testing component internal state.

app/pages/patient/ClinicsUsingAltRangeNotifications.js
```diff
@@ -0,0 +1,130 @@
...
+    }
+  }, []);
+
+  if (!isUserPatient) return null;
+
+  const handleDismiss = (clinicId) => {
+    const dismissedAt = moment.utc().toISOString();
+    const prefUpdate = { [getDismissedAltRangeNotificationKey(clinicId)]: dismissedAt };
+
+    dispatch(actions.async.updatePreferences(api, loggedInUserId, prefUpdate));
```
henry-tp, line 102: This is the JSON structure:

image

It's not great to have field values like that, but I needed to do something like this in order to allow the banner to pop up again when there is a new clinic. I had previously tried a different method where I used nested objects, but I don't want to complicate the schema of this preferences object too much. So this way we can keep a flat structure.

app/providers/AppBanner/AppBannerProvider.js
```diff
@@ -52,6 +54,15 @@ const AppBannerProvider = ({ children }) => {
   const isCustodialPatient = has(clinicPatient?.permissions, 'custodian');
   const userHasDiabetesType = !!loggedInUser?.profile?.patient?.diagnosisType;
 
+  const newClinicsUsingNonStandardRange = Object.values(clinics || {})
+    // We filter out any clinics that we already dismissed this banner for - we only show this
+    // banner if a NEW clinic has been added that has set a non-standard range. Thus this
+    // banner can re-appear if a NEW clinic is invited that sets a non-standard range.
+    .filter(clinic => !loggedInUser?.preferences?.[getDismissedAltRangeBannerKey(clinic.id)])
+    .filter(clinic => isRangeWithNonStandardTarget(clinic.patients?.[currentPatientInViewId]?.glycemicRanges));
```
henry-tp, line 62: I know that I could technically combine these chained calls to `.filter()`, but I wanted to keep it as as 2 separate and intuitive steps when reading

app/providers/AppBanner/appBanners.js
```diff
@@ -349,4 +349,28 @@ export const appBanners = [
...
+    variant: 'info',
+    priority: 12,
+    context: ['patient'],
+    paths: [pathRegexes.patientData],
+    getProps: (dispatch, loggedInUserId, clinicName = '') => ({
+      interactionId: 'ClinicUsingAltRange',
+      label: t(`${clinicName} is using a non-standard target range to view your data`),
+      message: t(`${clinicName} is using a non-standard target range to view your data`),
+      show: {
+        metric: 'Banner displayed Send Verification',
```
krystophv: The `show` metric text needs _adjustment_ :)

app/redux/actions/async.js
```diff
@@ -13,9 +13,13 @@ import personUtils from '../../core/personutils';
 import { keycloak } from '../../keycloak';
 import { push } from 'connected-react-router';
 import { worker } from '.';
+import { utils as vizUtils } from '@tidepool/viz';
+const { GLYCEMIC_RANGE } = vizUtils.constants;
```
krystophv: I think these two imports aren't used?

__tests__/unit/app/pages/patient/ClinicsUsingAltRangeNotifications.test.js
```diff
@@ -0,0 +1,160 @@
...
+        },
+        working: {
+          fetchingClinicsForPatient: { completed: true },
+        },
+      },
+    };
+
+    let store;
+
+    it('renders a dismissable notification for each clinic with non-standard ranges', async () => {
```
krystophv:
```suggestion
    it('renders a dismissible notification for each clinic with non-standard ranges', async () => {
```

app/providers/AppBanner/AppBannerProvider.js
```diff
@@ -52,6 +54,15 @@ const AppBannerProvider = ({ children }) => {
   const isCustodialPatient = has(clinicPatient?.permissions, 'custodian');
   const userHasDiabetesType = !!loggedInUser?.profile?.patient?.diagnosisType;
 
+  const newClinicsUsingNonStandardRange = Object.values(clinics || {})
```
krystophv: With the `|| {}` added on line 44, this one seems redundant?

__tests__/unit/app/pages/patient/ClinicsUsingAltRangeNotifications.test.js
```diff
@@ -0,0 +1,160 @@
...
+        working: {
+          fetchingClinicsForPatient: { completed: true },
+        },
+      },
+    };
+
+    it('renders nothing', () => {
+      store = mockStore(state);
+
+      const { container} = render(
```
krystophv:
```suggestion
      const { container } = render(
```

krystophv (changes requested):
Handful of little comments/suggestions, but broadly looking good!

__tests__/unit/app/pages/patient/ClinicsUsingAltRangeNotifications.test.js
henry-tp: dea1e2805

__tests__/unit/app/pages/patient/ClinicsUsingAltRangeNotifications.test.js
henry-tp: Ahh, the suffix switcheroo that always gets me

dea1e2805

app/providers/AppBanner/appBanners.js
henry-tp: Copypasta problems 😅 
https://github.com/tidepool-org/blip/commit/dea1e28059aeeae80b25e5be19087edb80cc94d7

app/redux/actions/async.js
henry-tp: You are correct

https://github.com/tidepool-org/blip/commit/dea1e28059aeeae80b25e5be19087edb80cc94d7

app/providers/AppBanner/AppBannerProvider.js
henry-tp: Yeah, you're right 

dea1e2805

I've been _trying_ not to be overly cautious (e.g. by littering our codebase with safe navigation and fallback values when they are really not necessary). 

At the same time, I have had some challenges in adapting to one aspect of our codebase which is that we commonly leave fields `undefined` or unset instead of initializing them to `null`. Or another example would be initializing a field like `{ patientTags: null }` instead of `{ patientTags: [] }` (that one bit me last week 🤣).

But yeah, do know that I am trying not to be overly safe to pollute our codebase, I am aware that doing stuff like this makes it harder to infer whether the object actually exists or not downstream. However, in the code that I write, you will see me often preferring `null` over `undefined` or unset, even when `undefined` technically makes more sense, as I find it makes it much easier to interpret what objects a field _can_ have.

krystophv: I _do_ yearn for a fully-typed future. I think we generally lean into `undefined` or unset because of uncertainty about how a `null` value for a property gets interpreted by the backend and into the database. Pretty sure that `undefined`/unset _don't_ get sent at all where explicit `null`, I'm not _entirely sure about_.

PR #1717, opened 2025-07-30
[WEB-3869] - Refetch Sites and Tags after CRUD operations

henry-tp:
https://github.com/tidepool-org/platform-client/pull/205

- [x] Make frontend internally modify state after `createClinicSiteSuccess` and `createClinicPatientTagSuccess`
- [x] Tests
- [x] Update `platform-client` to latest release candidate
- [x] modify `platform-client` dependency in this PR after publishing

Waiting for:

- [x] Patient Tags should have field names in lowercase instead of titlecase after POST and PUT
- [x] Should get the Site back after PUT

PR #1732, opened 2025-08-25
WEB-3869 - JS Error on updating redux state after API call

henry-tp:


app/redux/reducers/misc.js
```diff
@@ -935,7 +935,13 @@ export const clinics = (state = initialState.clinics, action) => {
...
       const { clinicId, patientTag } = action.payload;
 
-      return update(state, { [clinicId]: { patientTags: { $push: [patientTag] } } });
+      return update(state, {
+        [clinicId]: {
+          patientTags: !!state[clinicId]?.patientTags
+            ? { $push: [patientTag] }
+            : { $set: [patientTag] },
+        },
+      });
```
henry-tp, line 944: In #1717 the architecture was changed such that we modify the redux state after the API response comes back by pushing the new tag/site into state manually. 

I did not correctly handle a scenario where tags/sites are `null` or `undefined` which is the case when a clinic has no tags/sites (I had expected it to be `[]`). 

This causes a JS error when adding the very _first_ tag or site.

PR #1744, opened 2025-09-05
WEB-3937 - Settings options vary

henry-tp:


app/pages/patientdata/patientdata.js
```diff
@@ -2213,6 +2213,15 @@ export const PatientDataClass = createReactClass({
...
       }, false);
 
+      if (chartType === 'settings') {
+        this.fetchAdditionalData({
+          returnData: false,
+          showLoading: true,
+          noDates: true,
+          type: 'pumpSettings,upload',
+        });
+      }
```
henry-tp, line 2223: This latter part is needed only if the settings page is the initial screen (i.e. user navigates directly to `/patients/:id/data/settings`).

PR #1750, opened 2025-09-15
WEB-3890 - TIDAL Dashboard

henry-tp:


app/pages/dashboard/TideDashboard.js
```diff
@@ -169,7 +224,7 @@ const SortPopover = React.memo(props => {
     section,
     sections,
     selectedClinicId,
-    setSections,
+    setSections = noop,
```
henry-tp, line 243: I removed the `sections` as a piece of state and made it just a variable, since it's currently directly derivable from the flags. This component is unused and has been for a while, so implementing it would probably require significant modification. I'll just put this noop here and if we need to use the component again we can adjust.

app/core/clinicUtils.js
```diff
@@ -453,14 +453,16 @@ export const patientSchema = config => {
...
+
 export const tideDashboardConfigSchema = yup.object().shape({
   period: yup
     .string()
     .oneOf(map(summaryPeriodOptions, 'value'))
     .required(t('Please select a duration period')),
   lastData: yup
     .number()
-    .oneOf(map(lastDataFilterOptions, 'value'))
+    .oneOf(map(tideDashboardLastDataFilterOptions, 'value'))
```
henry-tp, line 465: After this is a merged, any clinic that already has 14 days and 30 days set will see this config modal opened and be prompted to select one of the remaining options. Tested and is working 👍

test/unit/pages/ClinicPatients.test.js
```diff
@@ -2843,28 +2845,11 @@ describe('ClinicPatients', () => {
...
 
-            tags.at(0).hostNodes().simulate('click');
-            tags.at(2).hostNodes().simulate('click');
-
-            // Tags should now be selected
-            expect(selectedTags()).to.have.lengthOf(2);
-            expect(selectedTags().at(0).text()).to.equal('>test tag 1');
-            expect(selectedTags().at(1).text()).to.equal('ttest tag 3');
+            // Select 2 tags from select menu
+            wrapper.find(SelectTags).props().onChange(['tag1', 'tag3']);
```
henry-tp, line 2852: I searched far and wide but was not able to find a better way to simulate the selection process idiomatically with enzyme (the underlying components within `react-select` are kind of special). I think this is fine tbh

PR #1754, opened 2025-09-22
[WEB-3680] CIQ pump shutdown events

clintonium-119:
[WEB-3680] 

This is similar to the alarms handling we introduced a few releases ago, with a few notes:

1. Events can be found on any data type (as the tandem uploader driver will add the `pump-shutdown` annotation to the most recent datum post-shutdown, regardless of type), so we need to fetch pumpSettings datums as well on the daily view to make sure that we are able to visualise the events on that datum type as well.
2. We currently only have one event type, pump shutdown (and only rendering for CIQ is in scope currently), but I've modelled the code in a similar fashion to alarms, since we expect to have more event types to render down the road.
3. I needed to allow processing basals with a duration of zero in case the event gets tacked onto such a datum (as was the case in a test datum I came across)
4. Events currently are not added to the Print views

Related PRs:
https://github.com/tidepool-org/tideline/pull/499
https://github.com/tidepool-org/viz/pull/541

[WEB-3680]: https://tidepool.atlassian.net/browse/WEB-3680?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


krystophv (approved):
LGTM :shipit:

PR #1758, opened 2025-09-25
[WEB-3891] Patient limit updates

clintonium-119:
[WEB-3891]

UI updates are to now show patients remaining on the plan limits, some small copy changes, an image swap on the patientLimitReached tooltip, and the removal of a progress bar showing how much of the limit was filled within the clinic workspace header.

Most changes stem from a API responses have changes when fetching patient counts and patient count limit settings.

Patient count changed from a single numeric `patientCount` value to an object with `plan`, `demo`, and `total` properties.

Patient count settings' `hardLimit.patientCount` is now `hardLimit.plan`.

As well, we no longer mutate the patient count manually when removing or adding patients, or accepting patient invites within a clinic.  Each of these operations triggers a call to get fresh patient count objects from the server to ensure correct counts.  This was needed since we 

[WEB-3891]: https://tidepool.atlassian.net/browse/WEB-3891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/redux/actions/async.js
```diff
@@ -3161,15 +3156,15 @@ export function selectClinic(api, clinicId) {
 
         if (values.clinicPatientCount) {
           dispatch(sync.fetchClinicPatientCountSuccess(clinicId, values.clinicPatientCount));
-          selectedClinic.patientCount = values.clinicPatientCount?.patientCount;
+          selectedClinic.patientCount = values.clinicPatientCount;
```
henry-tp: [optional style preference] my only qualm is that any variable named `somethingCount` I expect to be of type Number, so I would have preferred this to take an alternate name to help with readability. not a big deal

henry-tp (approved):
LGTM 🚀

app/redux/actions/async.js
clintonium-119: I could change it to `patientCounts`

clintonium-119: I've changed it.  Just pinging for a quick re-review on the name change, just since it touches a lot of files.

clintonium-119: I've added a few more renames - it wouldn't make sense to rename this property without also pluralizing all of the other related variable names

PR #1760, opened 2025-09-29
Web 3981 picker dates

henry-tp:


app/components/chart/trends.js
```diff
@@ -347,11 +353,9 @@ const Trends = withTranslation()(class Trends extends PureComponent {
...
 
   isAtMostRecent() {
-    const mostRecentCeiling = getLocalizedCeiling(
-      this.props.mostRecentDatetimeLocation,
-      _.get(this.props, 'data.timePrefs', {})
-    ).toISOString();
-    return _.get(this.refs, 'chart.state.dateDomain.end') >= mostRecentCeiling;
+    const { mostRecentDatetimeLocation } = this.props;
+
+    return _.get(this.refs, 'chart.state.dateDomain.end') >= mostRecentDatetimeLocation;
```
henry-tp, line 358: I see no reason that the `mostRecentCeiling` needed to be calculated before. If the end of the current viewing window is past the latest data point, then we are definitely looking at the most recent data

app/components/ChartDateRangeModal.js
```diff
@@ -136,6 +147,25 @@ export const ChartDateRangeModal = (props) => {
...
+    const mostRecentDatumMoment = moment.utc(mostRecentDatumDate).tz(timezoneName);
+    const adjustedDates = setDateRangeToExtents(newDates);
+
+    // If the date selected contains the mostRecentDatum, we want to exclude the time
+    // between the mostRecentDatum and the end of that day. We also adjust the start
+    // time so that the period is a multiple of 24 hours.
+    if (mostRecentDatumMoment?.isBefore(adjustedDates?.endDate)) {
+      const hourCeiling = getLocalizedHourCeiling(mostRecentDatumDate, timePrefs);
+      const endDate = moment.utc(hourCeiling).tz(timezoneName);
+      const startDate = adjustedDates?.startDate?.hour(endDate.hour());
```
clintonium-119: In cases where a timezone change happens (say, a 1 hour shift) between the start and end dates, this would end up representing one more or one less hour than the expected multiple of 24 hours. 

If I'm thinking about this correctly (I may not be - correct me if I'm wrong), on a selection of 3 days in the spring here, if the change happened within the range, we would have 71 actual hours hrs represented.  Likewise, in the fall, we would have 73 hours.

Perhaps the startDate should be calculated by moment functions that take this into account?

app/components/PrintDateRangeModal.js
```diff
@@ -198,12 +227,44 @@ export const PrintDateRangeModal = (props) => {
...
+    const mostRecentDatumMoment = moment.utc(mostRecentDatumDates[chartType]).tz(timezoneName);
+    const adjustedDates = setDateRangeToExtents(newDates);
+
+    // If the date selected contains the mostRecentDatum, we want to exclude the time
+    // between the mostRecentDatum and the end of that day. We also adjust the start
+    // time so that the period is a multiple of 24 hours.
+    if (mostRecentDatumMoment?.isBefore(adjustedDates?.endDate)) {
+      const hourCeiling = getLocalizedHourCeiling(mostRecentDatumDates[chartType], timePrefs);
+      const endDate = moment.utc(hourCeiling).tz(timezoneName);
+      const startDate = adjustedDates?.startDate.hour(endDate.hour());
```
clintonium-119: Same concern here as in ChartDateRangeModal - are we accounting for DST changeovers when we simply set the endDate hour as the startDate hour?

app/components/chart/trends.js
```diff
@@ -142,7 +148,7 @@ const Trends = withTranslation()(class Trends extends PureComponent {
     const timezone = getTimezoneFromTimePrefs(timePrefs);
 
     // endpoint is exclusive, so need to subtract a day
-    const end = moment(datetimeLocationEndpoints[1]).tz(timezone).subtract(1, 'day');
+    const end = moment(datetimeLocationEndpoints[1]).tz(timezone);
```
clintonium-119, line 150: If you're sure that we never need to subtract the day, we can remove the comment above, but I'm wondering why we no longer subtract the day as below.

app/components/ChartDateRangeModal.js
henry-tp: You are correct. I've modified it with an additional step to subtract a 24-hour multiple, which should prevent this issue. 9087fb2a04746f62bdccc974e4e6c4d80ec711a8

app/components/PrintDateRangeModal.js
henry-tp: As with other comment: 9087fb2a04746f62bdccc974e4e6c4d80ec711a8

app/components/chart/trends.js
henry-tp: Yes! This was a tricky problem I had to deal with to ensure that the time stamps we render and the data we grab are all in alignment. This call is within the `getTitle` method so its output represents the time range shown in the middle of this blue bar:

image

image

I wrote an explanation out but found it was hard to convey the intentions and reasoning. I fed my explanation through Claude and asked it to improve my explanation, so I apologize that the next part sounds like AI (it is.) 

---

## The Original Problem

Previously, our date range handling used an "exclusive endpoint" approach that was causing confusion and requiring complex transformations throughout the codebase. Here's how it worked:

**Example: User selects "Oct 6 - Oct 10" in the DatePicker**

1. **Top-level state would be set to:** `Oct 6 00:00:00 - Oct 11 00:00:00`
   - Notice the end date is actually Oct 11, not Oct 10
   - This was done to ensure we captured all data through the last second of Oct 10

2. **In the `getTitle` method, we had to subtract 1 day:**
   ```javascript
   const end = moment(datetimeLocationEndpoints[1]).tz(timezone).subtract(1, 'day');
   ```
   - This would transform `Oct 11 00:00:00` back to `Oct 10 00:00:00` for display
   - So the title would correctly show "Oct 6 - Oct 10"

## Why This Approach Was Problematic

This "add a day, then subtract a day" pattern worked for full-day ranges, but broke down when dealing with intraday time ranges or custom time periods:

**Example Problem Case:** User selects a custom range from "Oct 12 09:00 AM - Oct 17 09:00 AM"

1. **The exact range of data we want to query:** "Oct 12 09:00 AM - Oct 17 09:00 AM"
2. **What the old logic title logic would show:** "Oct 12 09:00 AM - Oct 16 09:00 AM" (after subtracting 1 day)
3. **This was completely wrong** - it didn't represent the actual data range being queried

The core issue was that we were applying a "whole day" transformation (subtract 1 day) to all date ranges, regardless of whether they were actually whole-day ranges or specific time periods.

## The New Solution

I've implemented a cleaner approach that eliminates this back-and-forth transformation:

### 1. Changed how we set the top-level state for whole-day ranges

**Old approach:**
```javascript
// When user selects "Oct 6 - Oct 10"
const formatDateEndpoints = dates => ([
    moment.utc(dates.startDate).tz(timezoneName).startOf('day').valueOf(),
    moment.utc(dates.endDate).tz(timezoneName).add(1, 'day').startOf('day').valueOf(),
]);
// Result: Oct 6 00:00:00 - Oct 11 00:00:00
```

**New approach:**
```javascript
// When user selects "Oct 6 - Oct 10"
const formatDateEndpoints = dates => ([
    moment.utc(dates.startDate).tz(timezoneName).valueOf(),
    moment.utc(dates.endDate).tz(timezoneName).valueOf(),
]);
// Result: Oct 6 00:00:00 - Oct 10 23:59:59.999
```

The key difference: Instead of adding a day and using an exclusive endpoint, we now set the end time to the very last millisecond of the intended day (`23:59:59.999`).

### 2. Removed the subtract-1-day logic from getTitle

Since our timestamps now accurately represent the actual date range (no artificial "+1 day"), we can display them directly without any transformation:

```javascript
// New code - no subtraction needed
const end = moment(datetimeLocationEndpoints[1]).tz(timezone);
```

## Benefits of This Approach

1. **Consistency:** The same logic works for both whole-day ranges and intraday ranges
2. **Clarity:** What you see in the title is exactly what's stored in the state
3. **Simplicity:** No more "add a day, then subtract a day" transformations
4. **Accuracy:** Custom time ranges display correctly without artificial date manipulation

## Detailed Examples

**Whole-day range: "Oct 6 - Oct 10"**
- Old state: `Oct 6 00:00:00 - Oct 11 00:00:00` → subtract 1 day for display → "Oct 6 - Oct 10" ✓
- New state: `Oct 6 00:00:00 - Oct 10 23:59:59.999` → no transformation → "Oct 6 - Oct 10" ✓

**Intraday range: "Oct 12 09:00 AM - Oct 17 09:00 AM"**
- Old state: `Oct 12 09:00:00 - Oct 17 09:00:00` → subtract 1 day for display → "Oct 12 09:00 AM - Oct 16 09:00 AM" ✗
- New state: `Oct 12 09:00:00 - Oct 17 09:00:00` → no transformation → "Oct 12 09:00 AM - Oct 17 09:00 AM" ✓

henry-tp:
@clintonium-119 are you able to take a look at this again? I'll have to bump up viz to make sure it's got the latest `develop` changes as well

clintonium-119 (dismissed):
Changes LGTM :100:

PR #1840, opened 2026-01-13
WEB-4275 - EHR C2C from Email

henry-tp:
[WEB-4275]

[WEB-4275]: https://tidepool.atlassian.net/browse/WEB-4275?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Manual testing of this requires use of QA1 which the backend changes are currently deployed on.


test/unit/pages/OAuthConnection.test.js
```diff
@@ -1,413 +0,0 @@
-import React from 'react';
```
henry-tp, line 1: Replaced with jest version

app/components/clinic/PatientForm/PatientForm.js
```diff
@@ -98,8 +101,7 @@ export const PatientForm = (props) => {
 
   const formikContext = useFormik({
     initialValues: getFormValues(patient, clinicPatientTags, clinicSites),
-    initialStatus: { showDataConnectionsModalNext: false },
```
henry-tp, line 101: no need for this status anymore

henry-tp:
@krystophv this is ready for review again

app/pages/verificationwithpassword/verificationwithpassword.js
```diff
@@ -1,209 +1,275 @@
...
     return validationErrors;
   };
 
-  handleInputChange = (attributes) => {
-    var key = attributes.name;
-    var value = attributes.value;
-    if (!key) {
-      return;
-    }
+  const handleSkip = () => window?.close();
```
krystophv, line 138: since this page is navigated to and not opened by a script, does `window.close()` actually work here?

henry-tp: Surprisingly yes (I was surprised too)


https://github.com/user-attachments/assets/0a2a21de-63ad-4f45-a2d6-aa5eb06a0e92

app/pages/verificationwithc2c/VerificationWithC2C.js
```diff
@@ -0,0 +1,153 @@
...
+      setToast({
+        message: t('An error occurred. Your connection invite may be expired or already used. Create an account to check your connection status.'),
+        variant: 'danger',
+      });
+
+      return;
+    }
+
+    const popupUrl = getDataConnectionPopupUrl(providerName, restrictedToken);
+    dispatch(actions.sync.connectDataSourceSuccess(providerId, popupUrl));
```
krystophv, line 106: (If this is traced correctly) this would chain into `useProviderConnectionPopup` that will try and `fetchDataSources` on an unauthenticated session. I think this harmlessly 401s? Maybe we should gate that fetch on an authenticated session though?

app/pages/verificationwithpassword/verificationwithpassword.js
```diff
@@ -1,209 +1,275 @@
...
+        <Box sx={styleProps.notificationContainer}>
+          <Notification notification={notification || propsNotification}/>
+        </Box>
+      </Container>
+    </Box>
+  );
 };
 
-export default connect(mapStateToProps, mapDispatchToProps, mergeProps)(VerificationWithPassword);
+export default VerificationWithPassword;
```
krystophv, line 271: Good translation. I think we should try and maintain the `propTypes`, though.

__tests__/unit/app/pages/OAuthConnection.test.js
```diff
@@ -1,50 +1,550 @@
...
+          </Route>
+        </Switch>
+      </MemoryRouter>
+    </Provider>
+  );
+
+  describe('dexcom authorized', () => {
+    describe('when desktop non-custodial signup', () => {
+      utils.isMobile.mockReturnValue(true);
+
```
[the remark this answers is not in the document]
krystophv: @henry-tp will this negatively affect the tests?

app/pages/verificationwithc2c/VerificationWithC2C.js
henry-tp: Interesting. I probably missed this (`useProviderConnectionPopup` can be a bit to dig through). I just manually went through the full flow and it didn't try to fire that network request for me ... I'm not sure if it's because `useRedirectOnC2CSuccess` fires first and just gets there faster or what the reason is. Regardless, this is a good catch. 

I've gated it on `!!user`. Again, I can't actually get the bug to be produced but it seems to make sense. to put it there What do you think of this gate condition?

95cd29d24
3f89431782329cf6f4c12b4e749ae5b0e3a87a4d

app/pages/verificationwithpassword/verificationwithpassword.js
henry-tp: TY. The design was completely overhauled (unrecognizable from before) so it made sense to rewrite using functional components IMO. Was hoping you wouldn't mind :)

Added propTypes back: 81aeb2333

app/pages/verificationwithpassword/verificationwithpassword.js
```diff
@@ -1,209 +1,275 @@
-import PropTypes from 'prop-types';
-import React from 'react';
-import { connect } from 'react-redux';
-import { withTranslation } from 'react-i18next';
-import { bindActionCreators } from 'redux';
+import React, { useEffect, useState } from 'react';
```
krystophv: looks like `useEffect`, `Button`, `queryParams`, and `setToast` aren't actually used anywhere

app/pages/verificationwithpassword/verificationwithpassword.js
```diff
@@ -1,209 +1,275 @@
...
+    color: vizColors.blue50,
+    my: 2,
+  },
+  subtitleContainer: {
+    fontSize: 1,
+    display: 'flex',
+    justifyContent: 'center',
+    color: vizColors.blue50,
+    my: 2,
+  },
```
krystophv: not huge, but these styles aren't used here nor the same over in `VerificationWithC2C`

app/pages/verificationwithpassword/verificationwithpassword.js
krystophv: TIL

app/pages/verificationwithc2c/VerificationWithC2C.js
krystophv: lgtm

app/pages/verificationwithpassword/verificationwithpassword.js
henry-tp: Thanks for catching these. My editor isn't showing unused imports like it used to. Gotta fix it 

ecae72ad9

app/pages/verificationwithpassword/verificationwithpassword.js
henry-tp: As above 

ecae72ad9

app/pages/verificationwithpassword/verificationwithpassword.js
krystophv: I only recently realized my install of oxc was scoped to my nvm node version, which got updated semi-recently across the frontend repos, so my "global" npm packages were less "global" than intended. Need to put it on the list of things to revisit re: our eslint config when we get to it.

krystophv (approved):
LGTM 🚀

PR #1862, opened 2026-02-09
[WEB-4390] - Install RTK

henry-tp:
[WEB-4390]

platform-client: https://github.com/tidepool-org/platform-client/pull/224

I know it's +600 lines but It's not that bad, just renaming a bunch of test cases to use the 'thunk' named export (breaking change) 

You can't really test this manually, so here is a branch you can checkout if you want to see it in action: 
#1865

[WEB-4390]: https://tidepool.atlassian.net/browse/WEB-4390?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/redux/api/baseApi.js
```diff
@@ -0,0 +1,51 @@
+import { createApi, fetchBaseQuery, retry } from '@reduxjs/toolkit/query/react';
+import config from '../../config';
+import { keycloak } from '../../keycloak';
+import { v4 as uuidv4 } from 'uuid';
+import { RTK_QUERY_REQUEST_ERROR, RTK_QUERY_REQUEST_SUCCESS } from '../constants/actionTypes';
+
+import tidepoolApi from '../../core/api';
+
+const fallbackSessionTrace = uuidv4();
```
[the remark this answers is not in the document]
henry-tp: Won't do: The goal is to maintain the same session trace token for all requests

app/redux/api/baseApi.js
```diff
@@ -0,0 +1,55 @@
...
+  // To ensure that all requests from Blip have the same trace-session token, we inject
+  // the trace-session token that is generated by `platform-client`. In the future, when
+  // `platform-client` is no longer being used, we can remove this dependency and use
+  // only a token generated by Blip
+  return tidepoolApi?.tidepool?.getSessionTrace() || fallbackSessionTrace;
+};
+
+const getSessionToken = () => {
+  return keycloak?.token || '';
+};
```
henry-tp, line 22: The abstraction of `getSessionToken()` and `getSessionTrace()` may be a pointless exercise, but I wanted to create some level of separation such that the query call isn't calling on keycloak or tidepoolApi directly. A functional-style dependency-injection, if you will

app/redux/api/baseApi.js
```diff
@@ -0,0 +1,51 @@
...
+
+const baseQueryWithDispatch = async (args, api, extraOptions) => {
+  const result = await baseQuery(args, api, extraOptions);
+
+  // Notify Middleware
+  if (result.error) {
+    api.dispatch({ type: RTK_QUERY_REQUEST_ERROR, error: result.error });
+  } else {
+    api.dispatch({ type: RTK_QUERY_REQUEST_SUCCESS, data: result.data });
+  }
```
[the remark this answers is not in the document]
henry-tp: `RTK_QUERY_REQUEST_ERROR` is indeed being consumed by keycloak middleware

app/redux/store/configureStore.prod.js
```diff
@@ -35,12 +35,14 @@ import pendoMiddleware from '../utils/pendoMiddleware';
...
 import { keycloak, keycloakMiddleware } from '../../keycloak';
 import config from '../../config';
+import { RTKQueryApi } from '../api/baseApi';
 
 export const history = qhistory(createBrowserHistory(), stringify, parse);
 
 const reducer = combineReducers({
   blip: reducers,
   router: connectRouter(history),
+  [RTKQueryApi.reducerPath]: RTKQueryApi.reducer,
```
henry-tp, line 45: This creates a new top-level redux slice called `api`, so it's neatly separated from all existing `blip` data which is nice.

package.json
```diff
@@ -199,7 +200,7 @@
     "terser-webpack-plugin": "5.3.9",
     "theme-ui": "0.16.1",
     "tideline": "1.38.0",
-    "tidepool-platform-client": "0.64.0",
+    "tidepool-platform-client": "0.65.0-web-4390-install-rtk.1",
```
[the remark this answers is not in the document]
henry-tp: yes

yarn.lock
```diff
@@ -3568,6 +3644,30 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@open-draft/deferred-promise@npm:^2.2.0":
```
henry-tp: Essentially all of this stuff is from `msw`

jest.config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-  testEnvironment: 'jsdom',
+  testEnvironment: 'jest-fixed-jsdom',
```
henry-tp, line 3: 🤦🤦🤦  https://mswjs.io/docs/faq/#requestresponsetextencoder-is-not-defined-jest

yarn.lock
```diff
@@ -2813,6 +2813,68 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@inquirer/ansi@npm:^1.0.2":
```
henry-tp, line 2816: Essentially all of this stuff is from `msw`

app/redux/api/baseApi.js
```diff
@@ -0,0 +1,47 @@
...
+  // only a token generated by Blip
+  return tidepoolApi?.tidepool?.getSessionTrace() || fallbackSessionTrace;
+};
+
+const getSessionToken = () => {
+  return keycloak?.token || '';
+};
+
+const apiHost = config.API_HOST;
+const apiVersion = 'v1';
```
henry-tp, line 25: I notice that we hardcode `v1` in platform-client everywhere instead of putting it in `config` or something. I was thinking about putting it in `config` or at some higher level, but it seems like something we are intentionally not doing that

app/redux/api/baseApi.js
```diff
@@ -0,0 +1,47 @@
...
+  baseUrl: baseUrl,
+  prepareHeaders: (headers) => {
+    headers.set('x-tidepool-trace-session', getSessionTrace());
+    headers.set('x-tidepool-session-token', getSessionToken());
+
+    return headers;
+  },
+});
+
+export const RTKQueryApi = createApi({
```
henry-tp, line 39: bad name?

A conventional name might be `baseApi` or `emptySplitApi` but now we have two `api` objects in blip.

jest.setup.js
```diff
@@ -18,6 +18,10 @@ global.__PROD__ = false;
 global.__I18N_ENABLED__ = 'false';
 global.__DEV_TOOLS__ = false;
 
+window.config = {
```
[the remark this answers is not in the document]
henry-tp: I think it's fine as-is for now

jest.config.js
henry-tp: We could potentially extend it ourselves instead (it's not a lot of code) if we don't want to rely yet another package.

app/redux/store/configureStore.prod.js
clintonium-119: Nice 👍

jest.config.js
clintonium-119: I think it's fine to pull this in.  Doing it ourselves gives us the same result with likely more maintenance.

jest.setup.js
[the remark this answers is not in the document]
clintonium-119: Yeah, the whole window.config init makes me scratch my head every now and again, but agree it's fine.

app/redux/api/baseApi.js
clintonium-119: I'd lean towards `baseApi` to match the filename.  Any good code editor intellisense should allow us to easily find it's definition and what it does.

You're fine to leave the name as-is too.  I'm fine either way.

app/redux/api/baseApi.js
clintonium-119: I like this abstraction.

app/redux/api/baseApi.js
clintonium-119: It's good here for now.  Who knows if we ever will see a v2, but it's good to split the version out at any rate.

clintonium-119 (dismissed):
Everything looks great to me.  Nice clean work on this :)
LGTM 🚢

app/redux/api/baseApi.js
```diff
@@ -0,0 +1,47 @@
...
+  reducerPath: 'api',
+  baseQuery: retry(
```
krystophv, line 41: We might find that adding retries to _all_ endpoints might be undesirable if the calls aren't idempotent and we may want to move the unauth'd retry logic into the base query and actually `await` the keycloak updateToken attempt so they don't end up interleaved, but those improvements can be handled as we actually start to put this to use, I think.

krystophv (approved):
LGTM 🚢  
Excellent foundational work!

PR #1894, opened 2026-03-24
[WEB-4476] Prevent stale stats after panning daily/bgLog charts

clintonium-119:
## Problem
See [WEB-4476]
Two related bugs caused incorrect stats after using the pan buttons on the daily and bgLog chart views.  Note: I was only able to trigger the race condition on the daily chart, but it made sense to me to update the bgLog chart in like manner since it could theoretically happen at some point down the road there too.

**Stale stats after panning**

The pan buttons use a D3 animated transition that emits ~31 `'navigated'` events (one per animation frame). `handleDatetimeLocationChange` was creating a new `_.debounce` instance on each call and storing it via `setState`. Because `setState` is async, `cancel()` always failed — so up to 31 independent debounced queries would fire, racing each other and blocking the final correct query at the destination date.

Mouse drag was unaffected because it fires a single `'navigated'` event on release.

**Stale stats after toggling BGM/CGM following a pan**

`updateChartPrefs` (called by the BGM/CGM toggle with `queryStats: true`) was using `chartEndpoints.current` as the query endpoint. This value is only updated during full data queries, not stats-only pan queries — so after panning, it remained at the pre-pan date.

## Fix

- Move the debounced `onUpdateChartDateRange` call to a stable instance property created once in the constructor, so all `'navigated'` events share the same debounced function and cancel correctly (`daily.js`, `bgLog.js`)
- Use `this.state.endpoints` instead of `chartEndpoints.current` in `updateChartPrefs`, as it is always current (`patientdata.js`)

[WEB-4476]: https://tidepool.atlassian.net/browse/WEB-4476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/chart/bgLog.js
+++ b/app/components/chart/bgLog.js
@@ -212,6 +212,8 @@ class BgLog extends Component {
     this.chartType = 'bgLog';
     this.log = bows('BgLog View');
     this.state = this.getInitialState()
+    // Created once so rapid-fire D3 'navigated' events during pan animation properly debounce
+    this.debouncedDateRangeUpdate = _.debounce((...args) => this.props.onUpdateChartDateRange(...args), 250);
   }
 
   getInitialState = () => {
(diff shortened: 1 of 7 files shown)


henry-tp (approved):
Looks good, great work 🚢

PR #1909, opened 2026-04-21
[WEB-4545] editable fields in SMART-on-FHIR

krystophv:
for [WEB-4545], adjusts editable fields in SMART-on-FHIR

Summary: 

This pull request updates the behavior of the `PatientForm` and `EditPatientDialog` components to support a new "smartOnFhirMode". When this mode is enabled (based on the presence of a `smartCorrelationId`), only certain fields are editable, and the "Save Changes" button remains enabled if the form is valid. The changes also update related unit tests to reflect this new logic.

The most important changes are:

**PatientForm behavior updates:**

* Added a new `smartOnFhirMode` prop to `PatientForm`, which, when true, disables editing of identity fields (Full Name, Birthdate, MRN, Email), while leaving clinic-related fields (Tags, Sites, Target, DiabetesType) enabled. [[1]](diffhunk://#diff-56647acbddc0cf6df8f6df5618ecb52cac31772e37449d9e3fd71846bd80f04dR65) [[2]](diffhunk://#diff-56647acbddc0cf6df8f6df5618ecb52cac31772e37449d9e3fd71846bd80f04dL267-R268) [[3]](diffhunk://#diff-56647acbddc0cf6df8f6df5618ecb52cac31772e37449d9e3fd71846bd80f04dL284-R285) [[4]](diffhunk://#diff-56647acbddc0cf6df8f6df5618ecb52cac31772e37449d9e3fd71846bd80f04dL293-R294) [[5]](diffhunk://#diff-56647acbddc0cf6df8f6df5618ecb52cac31772e37449d9e3fd71846bd80f04dL315-R316) [[6]](diffhunk://#diff-56647acbddc0cf6df8f6df5618ecb52cac31772e37449d9e3fd71846bd80f04dL329-R330) [[7]](diffhunk://#diff-56647acbddc0cf6df8f6df5618ecb52cac31772e37449d9e3fd71846bd80f04dR410)

* Updated the prop usage in `EditPatientDialog` to pass `smartOnFhirMode` instead of `isReadOnly`, and changed the logic so that the "Save Changes" button is only disabled if the form is invalid, not when in smartOnFhirMode. [[1]](diffhunk://#diff-223ec81a91c21216fb2165c05aba994a562b85b513336e843048f8b6b8e18ba1L110-R110) [[2]](diffhunk://#diff-223ec81a91c21216fb2165c05aba994a562b85b513336e843048f8b6b8e18ba1L127-R127)

**Unit test updates:**

* Added and updated tests in `PatientForm.test.js` to verify that identity fields are disabled and clinic fields are enabled when `smartOnFhirMode` is true.

* Updated the `EditPatientDialog.test.js` mock and tests to check for the new `data-smart-on-fhir-mode` attribute, and to ensure the "Save Changes" button is enabled when `smartCorrelationId` is present. [[1]](diffhunk://#diff-412be25dd9c93d0954acd51860e4e238c69c9aa1fc1ab9d5d1c09aeaf3d56ec2L13-R13) [[2]](diffhunk://#diff-412be25dd9c93d0954acd51860e4e238c69c9aa1fc1ab9d5d1c09aeaf3d56ec2L68-R68) [[3]](diffhunk://#diff-412be25dd9c93d0954acd51860e4e238c69c9aa1fc1ab9d5d1c09aeaf3d56ec2L79-R91)

[WEB-4545]: https://tidepool.atlassian.net/browse/WEB-4545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/components/clinic/PatientForm/PatientForm.js
```diff
@@ -62,6 +62,7 @@ export const PatientForm = (props) => {
     searchDebounceMs,
     initialFocusedInput = 'fullName',
     isReadOnly = false,
+    smartOnFhirMode = false,
```
henry-tp: Hmm, I don't think I'm a fan of this prop as a way to gate these fields. In the app we currently have 7 uses of `<PatientForm />` but only `<EditPatientDialog />` passes down the prop as a parent. It kind of begs the question - why don't the other instances get passed this prop as well, or why can't `<PatientForm />` internally access the value of `isSmartOnFhir` from redux (which it could). The `<PatientForm />` component itself also needs implicit knowledge that `isSmartOnFhir` means that it is being used in an "editing a patient" context.

IMO it makes more sense within the `EditPatientDialog` component to tie together the concepts of "If editing a patient, and the app is in smart-on-fhir mode, the certain fields should be locked". It would help keep `<PatientForm />` generic and require less implicit knowledge to be utilized. 

That being said, I struggle to come up with an interface that is not overly complex. Maybe something like:

```jsx
// In EditPatientDialog.js
<PatientForm
  // ...
  disabledFields={{
    fullName: true,
    birthDate: true
  }}
/>
```

```jsx
// In PatientForm.js
const { disabledFields = {} } = props; 

return (
<TextInput
  name="fullName"
  disabled={disabledFields.fullName}
/>);
```

If you don't agree or think this is too complicated, I won't push back and we can move forward as-is, but I did want share my opinion on this

app/components/clinic/PatientForm/PatientForm.js
henry-tp: Given that there are no longer any uses of `isReadOnly`, can we remove it from this component entirely?

henry-tp (changes requested):
I had some thoughts on the interface

app/components/clinic/PatientForm/PatientForm.js
krystophv: I actually like this a lot. It'd keep the implicit knowledge of what "smart-on-fhir" mode (or whatever else situation comes up) is out of the PatientForm and make it easier to reuse/customize elsewhere.

app/components/clinic/PatientForm/PatientForm.js
krystophv: Yup. 👍🏼

henry-tp (approved):
Looks great, thank you!

PR #1910, opened 2026-04-21
WEB-4526 - Print from Clinic Patient List

henry-tp:
WEB-4526

Viz: https://github.com/tidepool-org/viz/pull/650

-----

- [x] Fix loading modal transition
- [x] Metrics
- [x] Optimize initial data fetch
- [x] DRY up re-used logic
- [x] PDF race condition !!!!!!!!!!!
- [x] Loading Modal animation
- [x] AGP functions as expected
- [x] Resolve patient-side issue
- [x] Test cache invalidation / race conditions
- [x] `STATUS.NO_PATIENT_DATA state`
- [x] Test Report Type mixes / data sufficiency mixes
- [x] Test excludedDevices handles 'Healthkit twiist' exclusion
- [x] Test 1-min/5-min data
- [x] Bug on `printWindow.focus()` call (QA5 / E2E3 / Deepwater)
- [x] Replace timePrefs (3x)
- [x] Resolve print pop-up blocking
- [x] Resolve some bgStats not showing?
- [x] Resolve 1-min / 5-min CGM toggle
- [x] Resolve bgSource (BGM / CGM) toggle
- [x] Resolve PDF Demographic headers missing
- [x] Resolve siteChangeSource (??)
- [x] Double check excluded Devices
- [x] Resolve `state.bgPrefs` vs inferred `bgPrefs` issue
- [x] cgmSampleIntervalRange in getQueries
- [x] timePrefs derived instead of stored?
- [x] isAutomatedBasalDevice & isAutomatedBasalDevice
- [x] Glycemic Ranges
- [x] Call cleanup useEffect in an outer wrapper to reduce risk of stale data
- [x] Replace PatientData.js with abstracted version
- [x] Tests


test/unit/pages/patientdata.test.js
```diff
@@ -1726,17 +1723,17 @@ describe('PatientData', function () {
 
       // queryData set to false and queryStats set undefined
       instance.updateChartPrefs({ trends: 'bar'}, false);
-      sinon.assert.notCalled(instance.getStatsByChartType);
+      jestExpect(getStatsByChartType).not.toHaveBeenCalled();
```
henry-tp, line 1726: Need a mix of jest, chai, and sinon assertions in here. It will be weird for a while

app/pages/patientdata/patientdata.js
```diff
@@ -1932,86 +1591,12 @@ export const PatientDataClass = createReactClass({
...
-    const pdfGenerationFailed = _.get(nextProps, 'generatingPDF.notification.type') === 'error';
-
-    if (nextState.printDialogProcessing && pdfGenerated) {
-      this.setState({ printDialogProcessing: false });
-
-      if (nextState.printDialogOpen && !pdfGenerationFailed) {
-        this.closePrintDialog();
-      }
-
-      if (nextProps.pdf.combined.url) {
```
henry-tp, line 1976: This functionality has all been moved (mostly copy-pasted as-is) to `usePrintWindow.js`

app/core/usePrintPDF/usePrintPDF.js
```diff
@@ -0,0 +1,239 @@
...
+
+const usePrintPDF = (
+  api,
+  patientId,
+  onPrintTriggered = noop,
+) => {
+  const dispatch = useDispatch();
+  const generateAGPImages = useGenerateAGPImages();
+  const { openPrintWindow, triggerPrint } = usePrintWindow();
+  const [requestId] = useState(crypto.randomUUID());
```
henry-tp, line 78: New: I made this `requestId` as another internal safeguard against race conditions, just in case. It gets continuously passed through the opts as the PDF is generated, and at the end if there is some other PDF in redux state, `triggerPrint()` will not fire.

app/redux/reducers/pdf.js
```diff
@@ -28,7 +28,7 @@ const pdf = (state = {}, action) => {
       return update(state, { $merge: action.payload });
 
     case actionTypes.GENERATE_AGP_IMAGES_SUCCESS: {
-      let newOpts = { ...state.opts, svgDataURLS: action.payload.images };
+      let newOpts = { ...action.payload.opts, svgDataURLS: action.payload.images };
```
henry-tp: I found that using `...state.opts` was vulnerable to race conditions. If the modal closed (and the state was wiped), then `state.opts` would be blank and would not carry over the previous opts. Using `...action.payload.opts` ensures that the `opts` from the previous step is carried over, even if `state.opts` has been cleared

app/core/usePrintPDF/usePrintPDF.js
```diff
@@ -0,0 +1,239 @@
...
+      case STATUS.GENERATING_PDF:
+        const queries = getQueries(data, patient, clinicPatient, clinic, getTimePrefs(), getPrintOpts());
+        const pdfOpts = getPdfOpts(getPrintOpts(), user, patient, clinicPatient);
+        dispatch(actions.worker.generatePDFRequest('combined', queries, pdfOpts, patientId));
+        break;
+
+      case STATUS.GENERATING_AGP:
+        const hasAgpBGM = pdf?.opts?.agpBGM?.disabled === false;
+        const hasAgpCGM = pdf?.opts?.agpCGM?.disabled === false;
+        const reportTypes = [(hasAgpBGM && 'agpBGM'), (hasAgpCGM && 'agpCGM')].filter(s => s);
```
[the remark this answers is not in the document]
henry-tp: We never hit this step if agp is disabled

app/core/usePrintPDF/helpers.js
```diff
@@ -1,13 +1,85 @@
...
+  forceDataWorkerAddDataRequest: true,
+  returnData: false,
+  useCache: false,
+});
+
+export const getMainFetchOpts = (timePrefs, opts, fetchedUntil) => {
+  const enabledOpts = filter(opts, { disabled: false });
+  const earliestPrintDate = min(at(enabledOpts, map(keys(enabledOpts), key => `${key}.endpoints.0`)));
+
+  const startDate = moment.utc(earliestPrintDate).tz(getTimezoneFromTimePrefs(timePrefs)).toISOString();
```
clintonium-119, line 34: This should likely  set `startDate` to `getEarliestPrintDate(opts, timePrefs)` via the utillity below.  It's identical to the opening 3 lines here.

app/core/usePrintPDF/usePrintPDF.js
```diff
@@ -206,7 +149,7 @@ const usePrintPDF = (
 
       case STATUS.TRIGGERING_PRINT:
         if (pdf?.opts?.requestId === requestId) {
-          triggerPrint(pdf);
+          openPDF(pdf);
         }
         setTimeout(() => {
```
clintonium-119: Perhaps a comment here for why this setTimeout is needed would be helpful.

Also, not critical, but it would be a good practice to clear the timeout on unmount.

app/pages/clinicworkspace/ClinicPatientsPrintModal.js
```diff
@@ -32,15 +29,17 @@ const LoadingModal = ({ onClose = noop }) => {
...
 
   const { status, modalData, canPrint, print } = usePrintPDF(api, patientId, onClose);
 
   useEffect(() => {
     if (status === STATUS.NO_PATIENT_DATA) {
-      setToast({ message: t('This patient does not have any data'), variant: 'danger' });
       onClose();
+      setToast({
+        message: t('Insufficient data for patient to generate any report.'),
+        variant: 'danger'
```
clintonium-119: Not a big deal - I'd likely only use 'danger' for actual errors, and use an 'info' or 'warning' variant for this sort of expected result for a patient with insufficient data.

clintonium-119 (changes requested):
Just a couple small nits, but otherwise very good.

app/core/usePrintPDF/helpers.js
henry-tp: Thanks, good catch, I'm usually better at catching WET logic than this 02e812034

app/core/usePrintPDF/usePrintPDF.js
henry-tp: Added

709b88b1f

f8f593fac

app/pages/clinicworkspace/ClinicPatientsPrintModal.js
henry-tp: Sounds good. I tried both and `warning` seems more appropriate visually

d2dc47288

henry-tp:
Hey @clintonium-119, I had to make significant changes to fix a discrepancy Bill found. 

The issue was that the Print Modal's report options come up with the wrong default date values.

**Root cause**

Viz's data worker (dataUtil) normalizes each datum's display-time fields based on the `timePrefs` currently set on the data worker. If `timePrefs` is never set, it falls back to a default, which produces incorrect times - and then the  Print Modal derives its default date ranges from those times.

The problem is that `timePrefs` only gets set on the worker when a `dataUtil.query()` passes them in. The Patient Data View happens to do this as part of its normal first load:

patientdata `queryData()` → `props.dataWorkerQueryDataRequest()` → `dataUtil.query()`

That first `queryData()` call's real purpose is to figure out the default `chartType`, but it also passes `timePrefs`, which triggers viz's `setTimePrefs` [here](https://github.com/tidepool-org/viz/blob/48190a24955485bb183fb233b22540930c0b4617/src/utils/DataUtil.js#L2160). So the chart view ends up timezone-correct purely as a side effect of an unrelated step.

The Print Modal (and the AGP Drawer) never make that query, so their worker keeps the naive default — hence the wrong defaults.

**Why I didn't just add a `dataUtil.query()` to the print flow**

The print/AGP hooks (usePrintPDF, useAgpCGM) have no reason to call `dataUtil.query()` — that call exists on the Patient Data View to drive `chartType` selection, which PDF generation doesn't need. Adding one purely to set `timePrefs` would be illogical and risky: the worker and store are shared with the page underneath, so a stray query could disturb the data currently visible on screen.

**The fix**

I set `timePrefs` at the end of `addData` instead. This felt like the right place: when the data worker ingests data, it should "set itself up" to display that data correctly, rather than depending on an incidental later `dataUtil.query()` call. I gated it behind a param so it only runs where we want it (the print fetch), to keep the blast radius small and avoid regressions. But I actually don't see a strong reason we can't call `setTimePrefs` to the latest data's timezone  *every* time we add data.

Relevant Blip change: db346ccb41c959983f16cc265c5f1c1ff20aebf5
Relevant Viz change: https://github.com/tidepool-org/viz/pull/650/changes/02dee27c6efc8479dc134da0b829a2a2eaba6b45

This resolves the issue on my end. 

Let me know if this all makes sense.

clintonium-119 (dismissed):
LGTM 💯

PR #1917, opened 2026-05-07
WEB-4333 - Resolve Vanta issues for `glob` pkg

henry-tp:
[WEB-4333]

Very low risk issue, requires allowing attacker to use glob CLI to execute commands embedded in malicious files. We don't allow for this anywhere.

[WEB-4333]: https://tidepool.atlassian.net/browse/WEB-4333?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -202,6 +202,10 @@
     "redux": "4.x"
   },
   "resolutions": {
+    "glob@^10.0.0": "^10.5.0",
+    "glob@^10.2.2": "^10.5.0",
+    "glob@^10.3.7": "^10.5.0",
+    "glob@^10.3.10": "^10.5.0",
```
henry-tp, line 208: I dug heavily through the yarn `descriptors` but there doesn't seem to be a reliable way to target all `10.x.x` with one k:v pair. The reason I can't just target glob is because there are a number of packages using glob `7.x.x` and there are breaking changes.

There _is_ a way if I just change one and then run `yarn install && yarn dedupe glob`, but that causes us to be further removed from having the yarn lockfile being a "direct projection" of package.json which I don't like (relying on `--immutable` and `--frozen-lockfile`).

So this is the most straightforward way to do it for now, it seems.

PR #1922, opened 2026-05-12
WEB-4459 - Fix Settings Disappearing

henry-tp:
[WEB-4459]

So the bug is kind of weird.  The `<option>` elements still took up space (had DOM measurements) even though `overflow: hidden` was applied. Typically, the outer container would ignore the size of the grandchildren when determining its own height, but in this case for some weird reason the outer contiainer always took that space into account. Some weird CSS behaviour that is difficult to unset - I couldn't get it to actually ignore the space taken up by the grandchildren.

Here's a visual "demonstration" of the bug so you can see why it's going blank. It's actually just scrolled down, but the scrolling is locked due to `overflow: hidden`. You see that when I un-apply `overflow: hidden` I can scroll to the top and everything is visible.

https://github.com/user-attachments/assets/d30e9ae8-cddb-4716-9659-e0f9af6cde0d

Again, I can't seem to make the popover ignore the size of the grandchildren for some reason, I tried a number of approaches. Some weird CSS properties of the popover. So instead, I am just constraining the height of the entire popover container.

[WEB-4459]: https://tidepool.atlassian.net/browse/WEB-4459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/components/chart/settings.js
+++ b/app/components/chart/settings.js
@@ -13,6 +13,7 @@ import AddRoundedIcon from '@material-ui/icons/AddRounded';
 import InfoRoundedIcon from '@material-ui/icons/InfoRounded';
 import launchCustomProtocol from 'custom-protocol-detection';
 import { DesktopOnly } from '../mediaqueries';
+import { colors as vizColors } from '@tidepool/viz';
 
 import {
   bindPopover,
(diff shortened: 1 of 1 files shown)


clintonium-119 (approved):
LGTM 👍

PR #1929, opened 2026-05-19
[WEB-4592] Fix patient list stale-overwrite race under slow network

clintonium-119:
[WEB-4592] Fix patient list stale-overwrite race under slow network

When multiple fetchPatientsForClinic dispatches are in flight
simultaneously, an older response arriving last would overwrite the
newer one's correctly-filtered result, then flipping to the unfiltered
set.

[WEB-4592]: https://tidepool.atlassian.net/browse/WEB-4592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


henry-tp (approved):
Not a fan of the counter method but it resolves the issue we have for now and I can't come up with an immediate alternative. We should make an effort to move to the fetching away from platform-client as soon as we can.

PR #1931, opened 2026-05-25
[WEB-4566] Clinic admin mfa column

clintonium-119:
[WEB-4566]

[WEB-4566]: https://tidepool.atlassian.net/browse/WEB-4566?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/pages/clinicadmin/clinicadmin.js
+++ b/app/pages/clinicadmin/clinicadmin.js
@@ -99,9 +99,8 @@ export const ClinicAdmin = (props) => {
   const sortedClinicianArray = useMemo(() => {
     const { orderBy, order } = sortOptions;
 
-    // compareLabels delegates to String.prototype.localeCompare; coerce non-string
-    // sort keys (e.g. boolean mfaEnabled) so columns backed by non-string fields sort safely.
-    const toSortKey = (val) => (typeof val === 'string' || val === null || val === undefined ? val : String(val));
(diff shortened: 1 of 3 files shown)


app/pages/clinicadmin/clinicadmin.js
```diff
@@ -97,9 +99,13 @@ export const ClinicAdmin = (props) => {
   const sortedClinicianArray = useMemo(() => {
     const { orderBy, order } = sortOptions;
 
+    // compareLabels delegates to String.prototype.localeCompare; coerce non-string
+    // sort keys (e.g. boolean mfaEnabled) so columns backed by non-string fields sort safely.
+    const toSortKey = (val) => (typeof val === 'string' || val === null || val === undefined ? val : String(val));
```
henry-tp: I believe this will coerce bools to `'false'` and `'true'`, meaning that the ones with MFA disabled will be sorted ahead of those with enabled (when ascending). Not a huge problem

just for brevity though I think `String(val)` returns the string so it might be cleaner to omit that

henry-tp (approved):

app/pages/clinicadmin/clinicadmin.js
clintonium-119: Good call.  I've simplified this.

PR #1932, opened 2026-05-25
[WEB-4565] Account settings mfa page

clintonium-119:
[WEB-4565]

[WEB-4565]: https://tidepool.atlassian.net/browse/WEB-4565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/pages/userprofile/EditPersonalDetailsDialog.js
```diff
@@ -0,0 +1,189 @@
...
+    if (!inProgress && completed !== null && prevInProgress) {
+      formikContext.setSubmitting(false);
+      if (completed === true) {
+        setToast({ message: t('Personal details updated.'), variant: 'success' });
+        onClose();
+      } else if (notification) {
+        setToast({ message: notification.message, variant: 'danger' });
+      }
+    }
+  }, [working]);
```
henry-tp: If possible, I think there is a good opportunity here to switch us to the RTK Query for this PUT request, since it's quite isolated and doesn't have a complex chain of events, which would spare us from some of this reliance on detection of changes to `working` state. It's still a little complex to update the `user` value after the request succeeds, but I think we can just use the existing `sync.updateUserSuccess()` for that. What do you think?

app/pages/userprofile/EditPersonalDetailsDialog.js
```diff
@@ -0,0 +1,189 @@
...
+
+  const handleUpdateEmail = () => {
+    trackMetric('Clicked Update Email in Account');
+  };
+
```
henry-tp, line 70: A small UX rec here, is that because the Dialog is never unmounted, the form doesn't reset if the user clicks "Cancel", which IMO is an expected behaviour. It may be good to manually reset the form on close. 

```suggestion
const handleClose = () => {
  formikContext.resetForm();
  onClose();
}


return (
  <Dialog onClose={handleClose}>
    <DialogActions>
      <Button variant="secondary" onClick={handleClose}>
        {t('Cancel')}
      </Button>
    </DialogActions>
   </Dialog>
);
```

Or we can unmount the modal, but that loses us the fade-out which is uglier.

app/core/selectors.js
```diff
@@ -159,3 +159,32 @@ export const selectIsSmartOnFhirMode = createSelector(
...
+    return {
+      enabled: !!mfa?.enabled,
+      enabledTime: mfa?.enabledTime ?? null,
+      device: {
+        name: mfa?.device?.name ?? null,
+        registeredTime: mfa?.device?.registeredTime ?? null,
+      },
+      recoveryCodes: {
+        used: mfa?.recoveryCodes?.used ?? 0,
+        total: mfa?.recoveryCodes?.total ?? 12,
```
henry-tp: I feel like the fallback being slipped in the selector function here seems kind of hidden and may be hard to track down if there is an error, and it also seems kinda magic number-ey. Can we use the fallback in <RecoveryCodesRow /> for the fallback instead? 

In general, are there occassions where the availability of these values is known to be unreliable?

henry-tp (changes requested):

app/pages/userprofile/userprofile.js
```diff
@@ -15,325 +14,617 @@
...
-      _.set(result, 'preferences.displayLanguageCode', formValues.lang);
-    }
+export function UserProfile({ trackMetric, history, api }) {
+  const { t } = useTranslation();
+  const user = useSelector((state) => {
+    const allUsersMap = state.blip?.allUsersMap;
+    const loggedInUserId = state.blip?.loggedInUserId;
+    return allUsersMap && loggedInUserId ? allUsersMap[loggedInUserId] : null;
+  });
+  const mfaStatus = useSelector(selectMfaStatus);
```
henry-tp: I think we can use the `selectUser` selector for this

clintonium-119: Yep.  Good call.  Will fix this.

app/core/selectors.js
clintonium-119: Yeah, I agree on moving it out of the selector.  The numbers themselves will be reliable - they're coming directly from keycloak.  Will set these to null in the selector as a fallback.

app/pages/userprofile/EditPersonalDetailsDialog.js
clintonium-119: Good catch.  I'll implement this in a handleClose.

app/pages/userprofile/EditPersonalDetailsDialog.js
clintonium-119: Yeah, I had considered it during development, but decided against it in the moment.  I ended up later needing to implement a new API route for enabling 2FA, so I used RTK there, and I agree it's a good time to migrate all updateUser calls to RTK here.

I'll make the changes.  It will also affect a couple of other pages, but it's a good time to migrate away from the old pattern.

clintonium-119:
Thanks for the thoughtful review.  Changes made.  Ping for re-review

app/pages/userprofile/EditPersonalDetailsDialog.js
```diff
@@ -0,0 +1,182 @@
...
+      role: _.get(user, 'profile.clinic.role', ''),
+    },
+    validationSchema: schema,
+    onSubmit: (values) => {
+      const profileUpdates = { profile: { fullName: values.fullName.trim() } };
+      if (isClinician) {
+        profileUpdates.profile.clinic = { role: values.role || '' };
+      }
+      updateUserProfile(profileUpdates);
+    },
```
henry-tp: I hope you don't feel like I'm strongarm-ing here. 

The next thing I think would be a _huge_ improvement to our code quality is to move away from using useEffect "event listeners" because they can be hard to trace and understand the re-rendering implications of when we start having a lot of possible events on the page. With RTK query making the response available, it becomes much easier to tie the callback directly to the response.

For example, here, I think we can do:

```suggestion
onSubmit: async (values, { setSubmitting }) => {
      const profileUpdates = { profile: { fullName: values.fullName.trim() } };
      if (isClinician) {
        profileUpdates.profile.clinic = { role: values.role || '' };
      }
      try {
        await updateUserProfile(profileUpdates).unwrap();
        setToast({ message: t('Personal details updated.'), variant: 'success' });
        onClose();
      } catch (err) {
        setToast({ message: err?.data ?? t('An error occurred. Please try again.'), variant: 'danger' });
      } finally {
        setSubmitting(false);
      }
```

and I think this should be do-able for all 3 cases of `useUpdateUserProfileMutation()`. 

What do you think?

app/redux/features/userProfile/userProfileApi.js
```diff
@@ -0,0 +1,56 @@
...
+        if (userUpdates.username === loggedInUser.username) {
+          userUpdates = omit(userUpdates, 'username', 'emails');
+        }
+
+        try {
+          const updatedUser = await new Promise((resolve, reject) => {
+            tidepoolApi.user.put(userUpdates, (err, user) => {
+              if (err) reject(err);
+              else resolve(user);
+            });
```
henry-tp, line 35: (Discussion, no changes required)

I was a little confused here at why we would continue to use`tidepoolApi.user.put` (a blip wrapper around `platform-client` methods) instead of using RTK Query to make the request directly. I had missed that `tidepoolApi.user.put` is actually fires 3 parallel requests, I thought it was just one. So I apologize for missing that, and I probably wouldn't have made the request if I realized that.

That being said, I personally do still feel like this is an improvement. We could do the work now to switch these 3 parallel requests to using RTK Query, but it could be quite invasive and I understand that this is not the goal of your ticket. Something we should aim to do in the future though.

henry-tp (changes requested):

app/redux/features/userProfile/userProfileApi.js
clintonium-119: Yeah, ti's always been a bit of a mess due to some early-product-life arch decisions, but I still think it's worth migratring away from the old redux thunk boilerplate at this point in time.  It dropped a lot of it, even if we still have some api-level cruft to bear with for a while longer.

app/pages/userprofile/EditPersonalDetailsDialog.js
clintonium-119: Yeah, I think you're right to push back here.  I'll get rid of those.  Old habits....

clintonium-119: Updated.

henry-tp (approved):
🎉  thanks for bearing with me

PR #1949, opened 2026-06-16
[WEB-4615] Fix infinite loading icons on data view after updating patients with no data.

clintonium-119:
[WEB-4615]

Summary — Fixes the patient-data view sticking on an infinite loading spinner (until
manual refresh) after a clinician edits — or sends a connection invite for — a patient
with no data.

Fix — EditPatientDialog cleared the data-worker cache on every successful clinic-patient
update (it reacts to the global updatingClinicPatient state), which strands a no-data
view on the loader. It now clears only when the dialog drove the update (isOpen) and
Target Range actually changed and the patient has chart data — comparing against the
patient's saved range, not the form's frozen initialValues. Demographic-only edits,
no-data patients, and foreign updates (e.g. adding a data source) no longer clear.

Tests — EditPatientDialog.test.js covers each path (foreign update, demographic-only,
Target Range change with/without data, repeat change); added a DataConnections
chained-dispatch test. Manually verified in a clinic workspace.

[WEB-4615]: https://tidepool.atlassian.net/browse/WEB-4615?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/__tests__/unit/components/navpatientheader/EditPatientDialog.test.js
+++ b/__tests__/unit/components/navpatientheader/EditPatientDialog.test.js
@@ -2,18 +2,30 @@ import React from 'react';
 import { createStore, applyMiddleware } from 'redux';
 import { thunk } from 'redux-thunk';
 import { Provider } from 'react-redux';
-import { render, screen } from '@testing-library/react';
+import { render, screen, fireEvent } from '@testing-library/react';
 import { ThemeProvider } from 'theme-ui';
+import { utils as vizUtils } from '@tidepool/viz';
 import theme from '../../../../app/themes/baseTheme';
 import EditPatientDialog from '../../../../app/components/navpatientheader/EditPatientDialog';
 import { ToastProvider } from '../../../../app/providers/ToastProvider';
+import { buildGlycemicRangesFromPreset } from '../../../../app/core/glycemicRangesUtils';
+
+// When set by a test, the PatientForm mock feeds this Formik bag back through onFormChange so the
+// dialog's patientFormContext (values / initialValues) is populated.
+let mockFormBag;
 
 jest.mock('../../../../app/components/clinic/PatientForm', () => {
-  return (props) => (
-    <div data-testid="PatientForm" data-is-read-only={props.isReadOnly ? 'true' : 'false'}>
-      PatientForm Mock
-    </div>
-  );
+  const React = require('react');
+  return (props) => {
+    React.useEffect(() => {
+      if (mockFormBag) props.onFormChange(mockFormBag);
+    }, []);
+    return React.createElement(
+      'div',
+      { 'data-testid': 'PatientForm', 'data-is-read-only': props.isReadOnly ? 'true' : 'false' },
+      'PatientForm Mock'
+    );
+  };
 });
 
 jest.mock('../../../../app/core/hooks', () => ({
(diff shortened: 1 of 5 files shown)


test/unit/components/datasources/DataConnections.test.js
```diff
@@ -681,6 +681,64 @@ describe('DataConnections', () => {
       });
     });
 
+    // --- Test C: probes the colleague's theory (activeHandler / button spinner never resets) ---
```
henry-tp: I feel like a lot of these comments (both in the test files and in the application files) are a bit too verbose or perhaps unnecessary, e.g here it talks about "probing the colleagues theory". Can we  cut down on comments or remove if unnecessary?

__tests__/unit/components/navpatientheader/EditPatientDialog.test.js
```diff
@@ -2,18 +2,30 @@ import React from 'react';
...
 import { ThemeProvider } from 'theme-ui';
+import { utils as vizUtils } from '@tidepool/viz';
 import theme from '../../../../app/themes/baseTheme';
 import EditPatientDialog from '../../../../app/components/navpatientheader/EditPatientDialog';
 import { ToastProvider } from '../../../../app/providers/ToastProvider';
+import { buildGlycemicRangesFromPreset } from '../../../../app/core/glycemicRangesUtils';
+
+// When set by a test, the PatientForm mock feeds this Formik bag back through onFormChange so the
+// dialog's patientFormContext (values / initialValues) is populated.
+let mockFormBag;
```
henry-tp: With this formik `FormBag`, I'm concerned that things will start getting overly complex if we start needing more features and tests for `<EditPatientDialog />`, and we will start coding ourselves into a corner. I'm unsure of the reason we mocked up `<PatientForm />` in the first place (perhaps for isolation?). I think it would be better if we used the real thing so we don't need to mock up the useEffect behaviour and such.

Also when I review code, I like to actually try to make the  changes myself to make sure that I'm not recommending something that wouldn't work, so I don't end up looking dumb 😉  I put the code up on git at #1950 so it may be easier for you to copy over if you agree with my recommendations. Forgive me if this is heavy-handed or anything.

__tests__/unit/components/navpatientheader/EditPatientDialog.test.js
```diff
@@ -89,4 +101,154 @@ describe('EditPatientDialog', () => {
...
+            data: { metaData: { size: chartDataSize } },
+            clinics: {
+              clinic123: {
+                ...initialState.blip.clinics.clinic123,
+                patients: { patient123: { id: 'patient123', glycemicRanges: savedRange } },
+              },
+            },
+          },
+        };
+        return createStore((s = state, action) => { if (actions) actions.push(action); return s; }, applyMiddleware(thunk));
```
henry-tp: I think claude missed here that we have access to `redux-mock-store` to capture actions, so it make an ad-hoc utility to do so. I think we can use `redux-mock-store` instead. I also added this to the changes to #1950  so you can see what I mean

app/components/navpatientheader/EditPatientDialog.js
```diff
@@ -65,17 +67,42 @@ const EditPatientDialog = ({
...
+    // while the header is mounted. Only react when this dialog drove the update — otherwise a
+    // foreign update (e.g. adding a data source) would clear the data worker cache and leave the
+    // patient-data view stuck loading.
+    if (!isOpen) return;
 
-    dispatch(actions.worker.dataWorkerRemoveDataRequest(null, currentPatientInViewId));
+    onClose();
+
+    if (shouldClearDataRef.current) {
+      dispatch(actions.worker.dataWorkerRemoveDataRequest(null, currentPatientInViewId));
```
henry-tp, line 84: Because `<EditPatientDialog />` is permanently mounted on the header, it may be a good idea just to reset it to false for a clean slate

```suggestion
      dispatch(actions.worker.dataWorkerRemoveDataRequest(null, currentPatientInViewId));
        shouldClearDataRef.current = false;
```

henry-tp (approved):
Nothing blocking merge, but there are some improvements I think we can make to the test files

__tests__/unit/components/navpatientheader/EditPatientDialog.test.js
clintonium-119: Thanks.  I ended up pulling your changes in (will commit shortly)

It's not at all heavy-handed.  I appreciate your thoroughness, and these were all good suggestions.

app/components/navpatientheader/EditPatientDialog.js
clintonium-119: Done

__tests__/unit/components/navpatientheader/EditPatientDialog.test.js
clintonium-119: Applied

test/unit/components/datasources/DataConnections.test.js
clintonium-119: Yeah, I usually try to catch any of this cruft, and obviously didn't here.  Thanks for calling it out.  Removed a bunch of the cruft while trying to leave in the meat.

PR #1961, opened 2026-06-23
WEB-4611 - Round CGM use before display

henry-tp:
WEB-4611

The issue is that the minimum threshold of `70%` is being compared to the unrounded value (e.g. `69.6734%`) instead of the rounded value. The fix is that `69.xyz%` should be treated like `70%` and we should be using the rounded value for all intents and purposes.

The reason we don't need to do it for the _rest_ of the numbers in the tooltip is because the adjustment of the stats is handled by a call within `<BgSummaryCell />` to `reconcileTIRPercentages()`

image


clintonium-119 (approved):
LGTM 👍

PR #1965, opened 2026-06-25
[WEB-55] Daily view site change events

clintonium-119:
[WEB-55] 

Related PRs:
https://github.com/tidepool-org/viz/pull/651
https://github.com/tidepool-org/tideline/pull/553

[WEB-55]: https://tidepool.atlassian.net/browse/WEB-55?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


krystophv:
@coderabbitai review

app/components/chart/daily.js
```diff
@@ -307,15 +313,25 @@ class Daily extends Component {
...
         _.filter(nextProps.data.data.combined, d => !!d.tags?.alarm),
-        d => d.normalTime >= nextProps.data.data.current.endpoints.range[0] && d.normalTime <= nextProps.data.data.current.endpoints.range[1]
+        isInView
       );
 
-      if (hasAlarmEventsInView !== this.state.hasAlarmEventsInView) {
-        this.setState({ hasAlarmEventsInView });
-      }
+      const hasSiteChangeEventsInView = _.some(
+        _.filter(nextProps.data.data.combined, d => !!d.tags?.siteChange),
```
krystophv, line 327: `tags.siteChange` is set for any prime or reservoirChange (`viz DataUtil.js:847`), but the plot only draws the subtype the clinician selected, so these two can disagree.

Built a day of Tandem tubing primes with `siteChangeSource` resolved from a Tandem upload and no patient setting, which defaults to `cannulaPrime`:

```
hasSiteChangeEventsInView: true
icons drawn by tideline:   none
```

So the info icon appears and `EventsInfoTooltip` says site changes are displayed, over an events row with no site-change icons in it. Same for an unrecognized manufacturer, where `getSiteChangeSource` returns `'undeclared'`.

Matching the plot's test would mean filtering on the selected subtype rather than the tag. The three tag keys are the subtype names (`DataUtil.js:848-850`), so this reads directly:

```js
d => !!d.tags?.[nextProps.siteChangeSource]
```

`nextProps` rather than `this.props`, to match the rest of the block. That also handles `'undeclared'` for free, since there's no tag by that name.

app/components/chart/daily.js
```diff
@@ -307,15 +313,25 @@ class Daily extends Component {
       if (!_.isEmpty(updates)) this.chartRef.current?.rerenderChart(updates);
     }
 
-    if (nextProps.data?.data?.combined && (this.state.hasAlarmEventsInView === null || newEndpointsReceived)) {
+    const eventFlagsUnset = this.state.hasAlarmEventsInView === null || this.state.hasSiteChangeEventsInView === null;
+
+    if (nextProps.data?.data?.combined && (eventFlagsUnset || newEndpointsReceived)) {
+      const isInView = d => d.normalTime >= nextProps.data.data.current.endpoints.range[0] && d.normalTime <= nextProps.data.data.current.endpoints.range[1];
```
krystophv, line 319: Good extraction. Hoisting `isInView` out of the alarm branch makes the second predicate read as a parallel of the first rather than a copy.

app/components/chart/eventsInfoLabel.js
```diff
@@ -10,7 +10,7 @@ import { Body1 } from '../elements/FontStyles';
 const { EventsInfoTooltip } = vizComponents;
 
 const EventsInfoLabel = props => {
-  const { hasAlarmEventsInView } = props;
+  const { hasAlarmEventsInView, hasSiteChangeEventsInView } = props;
```
krystophv, line 13: nit: no `propTypes` on this component, and it's now reading two booleans instead of one. Since the label silently renders nothing when both are undefined, declaring them would make a wiring mistake upstream visible.

app/pages/patientdata/patientdata.js
```diff
@@ -625,6 +625,7 @@ export const PatientDataClass = createReactClass({
             onUpdateChartDateRange={this.handleChartDateRangeUpdate}
             onClickChartDates={this.handleClickChartDates}
             patient={this.props.patient}
+            siteChangeSource={vizUtils.aggregation.getSiteChangeSource(this.props.patient, this.getMetaData('latestPumpUpload.manufacturer', ''))}
```
krystophv, line 628: nit: resolving from `latestPumpUpload.manufacturer` is what pins the whole view to the current pump, so a previous pump's site changes are filtered out on every day. That matches Basics and the PDF, so no change needed here. Noting it because the ticket still carries an older answer saying previous-pump events should show, which is worth striking or confirming and implementing (raised on tideline#553 too).

krystophv (changes requested):
One thing worth settling before this lands: the Events label and the icons use different tests for "has site changes", so the label can appear over an empty events row. Detail on line 327.

The rest is small. Pulling `isInView` out of the alarm branch is a nice tidy.

app/pages/patientdata/patientdata.js
clintonium-119: I'd consider the 2017 Q&A answer in the migrated Trello comment ("for the previous pump, show iconography for all events") is superseded by the 2026-05-28 Done Criteria: Icon follows Basics selection, No Basics selection = no icons.  Agree that no change is needed.

PR #1972, opened 2026-07-03
[WEB-4629] Show 2FA banner on first workspace load after accepting invite

clintonium-119:
Small fix for an issue that was noticed for [WEB-4629]

After accepting a clinician invite, getClinicsForClinician seeds clinic.clinicians with only the viewer's own record. An admin viewer therefore looked like the sole admin, so the banner's sole-admin exemption suppressed the "Set Up 2FA" banner until a navigation round-trip reloaded the full roster.

To fix, we fetch the full clinician roster via a skip-gated RTK Query hook when the partial data makes the viewer appear to be the sole admin, and count admins from the fetched roster. The exemption is then decided on complete data and the banner renders on first load.

[WEB-4629]: https://tidepool.atlassian.net/browse/WEB-4629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/redux/features/clinicians/__mocks__/cliniciansApi.js
```diff
@@ -0,0 +1,5 @@
+/* global jest */
+
+// Stubs the RTK Query hook for suites on redux-mock-store (no api middleware). Defaults to no
+// data; override per-test with useGetCliniciansForClinicQuery.mockReturnValue({ data }).
+export const useGetCliniciansForClinicQuery = jest.fn(() => ({ data: undefined }));
```
henry-tp, line 5: I don't think this is used anywhere; should be able to delete the file?

test/unit/app/providers/AppBanner/AppBannerProvider.test.js
```diff
@@ -31,6 +31,8 @@ jest.mock('../../../../../app/components/datasources/DataConnections', () => ({
 
 jest.mock('../../../../../app/redux/features/mfaStatus/mfaStatusApi');
 import { useGetMfaStatusQuery } from '../../../../../app/redux/features/mfaStatus/mfaStatusApi';
+jest.mock('../../../../../app/redux/features/clinicians/cliniciansApi');
+import { useGetCliniciansForClinicQuery } from '../../../../../app/redux/features/clinicians/cliniciansApi';
```
henry-tp, line 35: We can do this for now since this is just for the hotfix, but in the future we should start switching to `msw` instead of mocking

henry-tp (approved):

test/unit/app/providers/AppBanner/AppBannerProvider.test.js
clintonium-119: Agree, but it depends what's under test.

baseApi/mfaStatusApi test the query layer itself (URLs, response mapping, retries), making msw a good fit. AppBannerProvider just consumes the data; it's not fetching. Going msw here means a real store + the legacy blip reducer and making ~30 non-network tests async on waitFor, for no obvious gain.

Suggested follow-up: keep this on hook-mocking and add a cliniciansApi endpoint test like mfaStatusApi.test.js when we want real roster-fetch coverage.

Thoughs?

app/redux/features/clinicians/__mocks__/cliniciansApi.js
clintonium-119: It's a Jest manual mock, not dead code.  The suites that `jest.mock('.../cliniciansApi')` with no factory (AppBannerProvider, useProviderConnectionPopup, app, ClinicianPatients, OAuthConnection, TideDashboard) pick it up by convention. 

It gives the hook a `{ data: undefined }` default; without it Jest auto-mocks the hook to `undefined` and `const { data } = useGetCliniciansForClinicQuery(...)` throws. Deleting it fails 5 tests in useProviderConnectionPopup.

Alternative is to drop this file and inline it everywhere:

```
jest.mock('.../cliniciansApi', () => ({
  useGetCliniciansForClinicQuery: jest.fn(() => ({ data: undefined })),
}));
```
More explicit (and a  precedent in AppBanner.test.js) but  it gets duplicated across 6 suites.   And if we go this route, we'd also want to do the same for the mfaStatusApi mocks, so it ends up being 2 file deletions and 12 inline mocks.

The jest docs don't seem ot recommend one approach over the other, so it's a personal preference issue.

I can see arguments both ways on this one.  Maybe this is a good topic for our next frontend huddle.  I'm ok with either approach, but I'm curious if both you and Chris heavily lean in one direction or the other.

PR #1989, opened 2026-07-15
WEB-4654 - Summary Period

henry-tp:
[WEB-4654]

[WEB-4654]: https://tidepool.atlassian.net/browse/WEB-4654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


__tests__/unit/app/pages/clinicworkspace/clinicPatientsFilters/FilterBySummaryPeriod.test.js
```diff
@@ -0,0 +1,79 @@
...
+    it('reflects the active summary period in the trigger label and the pre-selected radio', async () => {
+      wrapper = renderComponent({ activeSummaryPeriod: '7d' });
+
+      // Trigger label reflects the active period
+      expect(screen.getByRole('button', { name: /Summarizing 7 days of data/ })).toBeInTheDocument();
+
+      await userEvent.click(screen.getByRole('button', { name: /Filter by summary period duration/ }));
+
+      expect(screen.getByRole('radio', { name: /7 days/ })).toBeChecked();
+      expect(screen.getByRole('radio', { name: /14 days/ })).not.toBeChecked();
```
clintonium-119, line 74: For completeness, should probably verify that the 1 day option is unchecked as well.

app/pages/clinicworkspace/components/SummaryPeriodFilterDropdown.js
```diff
@@ -0,0 +1,174 @@
...
+        </Button>
+      </Grid>
+    </Box>
+  );
+};
+
+const SummaryPeriodFilterDropdown = ({
+  onChange = noop,
+  activeSummaryPeriod = null,
+}) => {
```
clintonium-119, line 106: `activeSummaryPeriod` is declared `.isRequired` at line 171 but also defaulted to `null` here. ES6 default params aren't applied before propTypes validation runs, so this default is unreachable — omitting the prop fires *both* a required-prop warning and a `oneOf` failure on `null`.

Separately, `FilterBySummaryPeriod.js:24` declares the same prop **optional** and passes it straight through, so the parent and child contracts disagree. The value always originates from `useLocalStorage('activePatientSummaryPeriod', '14d')` in `ClinicPatients.js:716`, so required-in-both seems right - I'd suggest that we just drop the default here.

```suggestion
const SummaryPeriodFilterDropdown = ({
  onChange = noop,
  activeSummaryPeriod,
}) => {
```

app/pages/clinicworkspace/clinicPatientsFilters/FilterBySummaryPeriod.js
```diff
@@ -0,0 +1,28 @@
...
+      onChange={handleChange}
+      activeSummaryPeriod={activeSummaryPeriod}
+    />
+  );
+};
+
+FilterBySummaryPeriod.propTypes = {
+  activeSummaryPeriod: PropTypes.oneOf(['1d', '7d', '14d', '30d']),
+  setActiveSummaryPeriod: PropTypes.func,
+};
```
clintonium-119, line 27: Counterpart to the `SummaryPeriodFilterDropdown` propTypes comment: this declares `activeSummaryPeriod` optional while the child declares it `.isRequired`. Since it's passed straight through, an undefined value here becomes a propTypes warning there. Suggest aligning both on required.

```suggestion
FilterBySummaryPeriod.propTypes = {
  activeSummaryPeriod: PropTypes.oneOf(['1d', '7d', '14d', '30d']).isRequired,
  setActiveSummaryPeriod: PropTypes.func,
};
```

app/pages/clinicworkspace/components/SummaryPeriodFilterDropdown.js
```diff
@@ -0,0 +1,174 @@
...
+
+  return null;
+};
+
+const getSummaryPeriodOptions = (t) => ([
+  { value: '1d', label: t('24 hours') },
+  { value: '7d', label: t('7 days') },
+  { value: '14d', label: t('14 days') },
+  { value: '30d', label: t('30 days') },
+]);
```
clintonium-119: `core/clinicUtils.js:83-88` already exports `summaryPeriodOptions` with these exact four entries, and the sibling filter in this same stack (`FilterByDataRecency.js`) imports `lastDataFilterOptions` from that module rather than re-declaring it. As written the period list now lives in four places: `clinicUtils`, this function, the `getSummaryPeriodSelectLabel` switch above, and the two `PropTypes.oneOf` arrays.

Can we import the shared export for consistency with `FilterByDataRecency`? That'd be `import { summaryPeriodOptions } from '../../../core/clinicUtils';` up top, deleting this block, and `options={summaryPeriodOptions}` at line 63.

You could also consider generating the oneOf propTypes list from the imported `summaryPeriodOptions` so that changing the options list only requires changing the options themselves, and the PropTypes will automatically update to suit.

__tests__/unit/app/pages/clinicworkspace/clinicPatientsFilters/FilterBySummaryPeriod.test.js
```diff
@@ -0,0 +1,79 @@
...
+      // "Summarizing ..." string that changes with the active period).
+      await userEvent.click(screen.getByRole('button', { name: /Filter by summary period duration/ }));
+      await screen.findByRole('radio', { name: /30 days/ });
+
+      await userEvent.click(screen.getByRole('radio', { name: /30 days/ }));
+      await userEvent.click(screen.getByRole('button', { name: 'Apply' }));
+
+      // The period value is passed straight through, not merged into an activeFilters object
+      expect(setActiveSummaryPeriod).toHaveBeenCalledTimes(1);
+      expect(setActiveSummaryPeriod).toHaveBeenCalledWith('30d');
```
clintonium-119, line 59: This spec already imports `trackMetric as mockTrackMetric` (line 9) and clears it in `beforeEach` (line 42), but never asserts on it — so the metric rename to `'Clinic - Summary period apply filter'` + `pageName` has no coverage after the legacy test was removed. Since the mock plumbing is already here, it's a cheap addition:

```suggestion
      expect(setActiveSummaryPeriod).toHaveBeenCalledWith('30d');

      expect(mockTrackMetric).toHaveBeenCalledWith('Clinic - Summary period apply filter', {
        clinicId: selectedClinicId,
        summaryPeriod: '30d',
        pageName: 'Population Health',
      });
```

app/pages/clinicworkspace/ClinicPatients.js
```diff
@@ -1962,99 +1945,19 @@ export const ClinicPatients = (props) => {
...
-                      }}
-                    >
-                      {t('Apply')}
-                    </Button>
-                  </DialogActions>
-                </Popover>
+                <FilterBySummaryPeriod
+                  activeSummaryPeriod={activeSummaryPeriod}
+                  setActiveSummaryPeriod={setActiveSummaryPeriod}
+                />
```
clintonium-119, line 1961: Nit: leftover formatting from removing the inline popover — `<FilterBySummaryPeriod>` and the `showRpmReportUI` block below sit at the same indent level as their `<Flex>` parent instead of one level deeper, which makes the nesting hard to follow against the closing tags at lines 1985-1986. There's also a stray blank line after the opening tag (and another at 1949).

```suggestion
                >
                  <FilterBySummaryPeriod
                    activeSummaryPeriod={activeSummaryPeriod}
                    setActiveSummaryPeriod={setActiveSummaryPeriod}
                  />

```

The `showRpmReportUI` block at 1962-1984 needs the same +2 shift.

app/pages/clinicworkspace/components/SummaryPeriodFilterDropdown.js
```diff
@@ -0,0 +1,174 @@
...
+  const { t } = useTranslation();
+  const pageName = useClinicMetricsPageName();
+  const selectedClinicId = useSelector((state) => state.blip.selectedClinicId);
+
+  const [pendingSummaryPeriod, setPendingSummaryPeriod] = useState(activeSummaryPeriod);
+
+  const handleChange = (summaryPeriod) => onChange(summaryPeriod);
+
+  return (
+    <Box data-testid='summary-period-filter-dropdown' mt={5} mx={2} sx={{ width: 300 }}>
```
clintonium-119: Nit: every sibling dropdown uses double quotes for this attribute (`DataRecencyFilterDropdown.js:46`, `SiteFilterDropdown.js`, `TagFilterDropdown.js`).

```suggestion
    <Box data-testid="summary-period-filter-dropdown" mt={5} mx={2} sx={{ width: 300 }}>
```

app/pages/clinicworkspace/components/SummaryPeriodFilterDropdown.js
```diff
@@ -0,0 +1,174 @@
...
+      <Grid sx={{ gridTemplateColumns: '1fr 1fr' }} mt={3} mb={2}>
+        <Button
+          id="cancel-summary-period-filter"
+          sx={{ fontSize: 1 }}
+          variant="secondary"
+          onClick={() => {
+            trackMetric('Clinic - Summary period filter cancel', { clinicId: selectedClinicId, pageName });
+            setPendingSummaryPeriod(activeSummaryPeriod);
+            onClose();
+          }}
```
clintonium-119: Nit: `DropdownContent` is conditionally rendered behind `summaryPeriodPopupFilterState.isOpen` (line 157), so it unmounts on close and `pendingSummaryPeriod` is re-initialised from `activeSummaryPeriod` on the next open regardless. This reset is a no-op. The "cancel resets the pending selection" test still passes either way, since it's really asserting the remount behaviour.

```suggestion
          onClick={() => {
            trackMetric('Clinic - Summary period filter cancel', { clinicId: selectedClinicId, pageName });
            onClose();
          }}
```

clintonium-119 (changes requested):
Great job refactoring all of this.  I love seeing ClinicPatients get leaned down/chopped away at.

Just some small bits of feedback in the comments to address.  Nothing major.

app/pages/clinicworkspace/components/SummaryPeriodFilterDropdown.js
henry-tp: Good catch, not sure how I missed that it was already defined

a0143d522969efe631b80f9a97dd1647ffa795a4
35f09a0f9b7b420db85f15993bdd9ea901cacd51

app/pages/clinicworkspace/components/SummaryPeriodFilterDropdown.js
henry-tp: 639fe1b27812bb4aeb90b5002957299ba312e3b4

app/pages/clinicworkspace/ClinicPatients.js
henry-tp: 639fe1b27812bb4aeb90b5002957299ba312e3b4

__tests__/unit/app/pages/clinicworkspace/clinicPatientsFilters/FilterBySummaryPeriod.test.js
henry-tp: The tests for metrics are missing, but IMO the tracking of metrics is not the intended responsibility of the `FilterBySummaryPeriod`, but instead the responsibility of `SummaryPeriodFilterDropdown`. I added the tests there and removed the imports here: 639fe1b27812bb4aeb90b5002957299ba312e3b4

app/pages/clinicworkspace/components/SummaryPeriodFilterDropdown.js
henry-tp: 639fe1b27812bb4aeb90b5002957299ba312e3b4

__tests__/unit/app/pages/clinicworkspace/clinicPatientsFilters/FilterBySummaryPeriod.test.js
henry-tp: 639fe1b27812bb4aeb90b5002957299ba312e3b4

app/pages/clinicworkspace/clinicPatientsFilters/FilterBySummaryPeriod.js
henry-tp: 35f09a0f9b7b420db85f15993bdd9ea901cacd51

clintonium-119 (approved):
LGTM!

PR #1990, opened 2026-07-15
WEB-4654 - Time in Range

henry-tp:
[WEB-4654]

[WEB-4654]: https://tidepool.atlassian.net/browse/WEB-4654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/pages/clinicworkspace/components/TimeInRangeFilterDropdown.js
```diff
@@ -0,0 +1,322 @@
...
+import Checkbox from '../../../components/elements/Checkbox';
+
+import { colors } from '../../../themes/baseTheme';
+import { MGDL_UNITS } from '../../../core/constants';
+
+const { reshapeBgClassesToBgBounds, generateBgRangeLabels } = vizUtils.bg;
+import useClinicMetricsPageName from '../useClinicMetricsPageName';
+import { timeInRangeFilterThresholds } from '../../../core/clinicUtils';
+
+const getTimeInRangeFilterOptions = (showExtremeHigh = false, t) => [
```
clintonium-119, line 43: This definition is assembled from translated fragments (`prefix`, `t('Time')`, `t(bgPrefix)`), so English word order is baked in. `t(bgPrefix)` is also a dynamic key — i18next-parser only extracts literals, so it can never be emitted by `update-translations`.

Whole-sentence keys instead:

```suggestion
const getRangeDefinition = (t, { comparator, threshold, bgRange, isBounded }) => {
  if (isBounded) {
    return comparator === '<'
      ? t('Less than {{threshold}}% Time between {{bgRange}}', { threshold, bgRange })
      : t('Greater than {{threshold}}% Time between {{bgRange}}', { threshold, bgRange });
  }

  return comparator === '<'
    ? t('Less than {{threshold}}% Time {{bgRange}}', { threshold, bgRange })
    : t('Greater than {{threshold}}% Time {{bgRange}}', { threshold, bgRange });
};

const getTimeInRangeFilterOptions = (showExtremeHigh = false, t) => [
```

Then drop `prefix: t('Greater than')` / `t('Less than')` from all six options below — the comparator already lives in `timeInRangeFilterThresholds`, so those are a second encoding of the same fact.

Two of the four keys are unreachable today (`generateBgRangeLabels` only sets a `prefix` on `target`), but they keep the helper total if a range changes shape.

---

Same pattern in `ClearFilterButtons.js`, `FILTER_AND_SEARCH` case — `[Reset All Filters] or [Clear Search]` joins two buttons with a floating `t('or')`, which a translator can't reorder. Wants `<Trans>` so it extracts as `<0>Reset All Filters</0> or <1>Clear Search</1>`:

```jsx
<Trans t={t}>
  <ClearButton className="reset-filters-button" onClick={onResetFilters}>Reset All Filters</ClearButton>
  {' '}or{' '}
  <ClearButton className="clear-search-button" onClick={onClearSearch}>Clear Search</ClearButton>
</Trans>
```

Noting it here rather than inline since that file lands in #1984, not this PR.

app/pages/clinicworkspace/components/TimeInRangeFilterDropdown.js
```diff
@@ -0,0 +1,322 @@
...
+  return (
+    <Box data-testid="time-in-range-filter-dropdown" mt={5} mx={2}>
+      <Box sx={{ padding: 1, color: vizColors.gray50, lineHeight: 1 }} mb={2}>
+        <Box sx={{ fontWeight: 'medium', fontSize: 1 }}>{t('% Time in Range')}</Box>
+        <Box mt={1} sx={{ fontSize: 0 }}>{t('Only patients using the standard target range will be included.')}</Box>
+      </Box>
+
+      <Box sx={{ border: `1px solid ${vizColors.gray10}`, borderRadius: 6, padding: 3 }}>
+        {map(filterOptions, ({ value, title, rangeName, threshold, prefix }, i) => {
+          const { prefix: bgPrefix, suffix, value: glucoseTargetValue } = bgLabels[rangeName];
```
clintonium-119: Derive the definition once, up front:

```suggestion
        {map(filterOptions, ({ value, title, rangeName, threshold }, i) => {
          const { prefix: bgPrefix, suffix, value: bgValue } = bgLabels[rangeName];
          const { comparator } = timeInRangeFilterThresholds[value];

          const definition = getRangeDefinition(t, {
            comparator,
            threshold,
            bgRange: `${bgValue} ${suffix}`,
            isBounded: !!bgPrefix,
          });
```

`bgRange` stays a plain template literal on purpose — it's `"<54"` + `"mg/dL"`, data with no words, so it needs no key.

app/pages/clinicworkspace/components/TimeInRangeFilterDropdown.js
```diff
@@ -0,0 +1,322 @@
...
+                    {' '}
+                    {threshold}%
+                    {' '}
+                    {t('Time')}
+                    {' '}
+                    {bgPrefix && `${t(bgPrefix)} `}
+                    {' '}
+                    {glucoseTargetValue}
+                    {' '}
+                    {suffix}
```
clintonium-119:
```suggestion
                    {definition}
```

Also fixes a latent double space: `{' '}` sits on both sides of the optional `bgPrefix`, and the truthy branch bakes in its own trailing space too. Every row is affected. HTML whitespace collapsing and RTL's `getByText` normalization both hide it, which is why nothing caught it.

clintonium-119 (changes requested):
This is functionally complete as far as I can tell, but I did have some concerns with how the filter definition was being stringed together.

Please see the comments on `TimeInRangeFilterDropdown.js`.  I think it'll end up being a cleaner way to render the string, and, more importantly, will allow us more flexibility when we begin supporting other translations.

app/pages/clinicworkspace/components/TimeInRangeFilterDropdown.js
henry-tp: Makes sense 
fd71f21b9
f9aa7f6ed

app/pages/clinicworkspace/components/TimeInRangeFilterDropdown.js
henry-tp: fd71f21b9

app/pages/clinicworkspace/components/TimeInRangeFilterDropdown.js
henry-tp: fd71f21b9

clintonium-119 (approved):
Great job!  LGTM +1

PR #2005, opened 2026-08-18
WEB-4460 - Dashboard Setup

henry-tp:
WEB-4460

Sorry, had to kick it off with a big one. I felt it was the minimum amount of code changes needed for something testable that isn't just "I fetched something and displayed Hello World on the page". But yeah, this is just to be able to show the dashboard, flip through pages and categories.


app/components/clinic/WorkspaceSwitcher.js
```diff
@@ -85,6 +86,7 @@ export const WorkspaceSwitcher = props => {
     dispatch(actions.sync.setIsPatientListVisible(false));
     dispatch(actions.async.selectClinic(api, option.id));
     dispatch(push(option.id ? '/clinic-workspace' : '/patients', { selectedClinicId: option.id }));
+    dispatch(resetTideDashboardState());
```
henry-tp, line 89: So this is a little tricky, and I don't like doing it this way.

In most situations, I like to reset the state of the view in Redux then the page dismounts, e.g. `useEffect(() => { return resetPage() }, []);`. This is for stuff like which page number you're on, so that you're back to Page 1 if you come back later. I mean, you should basically _always_ reset to Page 1 if you leave the page, right.

The issue is that Product has identified that users want to keep the view state particularly in this scenario:

- User clicks patient 
- Redirected to Patient Data View
- User clicks "Back" on Patient Data View
- Redirected back to TIDE Dashboard

In this case, the user wants to maintain their current position in the list of patients (which is effectively a combination of Category + Page Number) so they can see the _next_ patient in their checklist.

I think that resetting the state when (1) the user clicks "TIDE Dashboard" on the button and (2) when they switch workspaces is good enough and covers all of the routes to accessing the TIDE Dashboard organically, but it is a little more brittle and "correct"  than I like it to be. In the future this will kinda break if there are new ways of accessing TIDE Dashboard, but that seems like small potatoes.

app/core/navutils.js
```diff
@@ -44,11 +44,11 @@ export const getPatientListLink = (clinicFlowActive, selectedClinicId, query, pa
   const drawerTab = query?.drawerTab;
 
   if (dashboard && drawerTab !== undefined && patientId) {
-    return `/dashboard/${dashboard}?drawerPatientId=${patientId}&drawerTab=${drawerTab}`;
+    return `/clinic-workspace/tide-dashboard?drawerPatientId=${patientId}&drawerTab=${drawerTab}`;
   }
 
   if (dashboard) {
-    return `/dashboard/${dashboard}`;
+    return '/clinic-workspace/tide-dashboard';
```
henry-tp, line 51: This is a bit of a hierarchical shift, but I'm going to frame the TIDE Dashboard as a subunit of the `clinic-workspace`, as will be the Device Issues Dashboard. This is reflected in the directory structure as well

app/pages/clinicworkspace/TideDashboardV2/CGMExclusionQuery.js
```diff
@@ -0,0 +1,43 @@
+// Builds mutually exclusive query params for CGM categories.
+
+// Rules are registered in priority order. Each rule's query params include negations
+// of all higher-priority rules, so patients match at most one category.
+
+export default class CGMExclusionQuery {
```
henry-tp, line 6: See, I use objects sometimes ;) 

I'm going to avoid commenting further on the purpose/function of this utility; I'd like to know your thoughts on the comprehensibility at first glance. If you have trouble deciphering it quickly, it might be a sign that I should rework it. Then again we live in a very AI time so maybe first-glance comprehensibility has lower value 🤷

test/unit/pages/ClinicPatients.test.js
```diff
@@ -2013,202 +2013,13 @@ describe('ClinicPatients', () => {
             defaultProps.trackMetric.resetHistory();
           });
 
-          it('should render the TIDE Dashboard CTA', () => {
```
henry-tp, line 2016: The responsibility of rendering these buttons has shifted to the `clinicworkspace` component, so it will just be deletions in this file

app/pages/clinicworkspace/clinicworkspace.js
```diff
@@ -96,9 +115,16 @@ export const ClinicWorkspace = (props) => {
...
-    trackMetric(tabs[newValue]?.metric, { clinicId: selectedClinicId, source: 'Workspace table' });
+    const newTab = tabs[newValue];
+
+    trackMetric(newTab?.metric, { clinicId: selectedClinicId, source: 'Workspace table' });
     setSelectedTab(newValue);
-    dispatch(push(`/clinic-workspace/${tabs[newValue].name}`));
+
+    if (newTab?.name === 'tide-dashboard') {
+      dispatch(resetTideDashboardState());
+    }
```
henry-tp, line 151: See https://github.com/tidepool-org/blip/pull/2005/changes#r3826133458

app/pages/clinicworkspace/TideDashboardV2/useTideDashboardPatients.js
```diff
@@ -0,0 +1,31 @@
...
+
+const LIMIT = 12;
+
+// TEMPORARY, will be set in redux
+const tideDashboardFilters = {
+  lastData: 7,
+  patientTags: [],
+  clinicSites: [],
+  summaryPeriod: '14d',
+};
```
henry-tp, line 13: Sorry about this. I don't like having "will do" code in a PR that actually touches business logic. I thought about actually setting up the Redux slice in this PR, but then those slice values won't be modifiable, so you can't actually test them. I think it's better to  define the slice in a PR where we actually connect it to setters & getters, so you can test it. But then that leaves stuff like this.

May make sense to discuss this workflow detail in a FE huddle or smth

__tests__/unit/app/pages/clinicworkspace/TideDashboardV2/TideDashboardV2.test.js
```diff
@@ -0,0 +1,254 @@
...
+    store = setupStore({ blip: { selectedClinicId: 'clinic123' } }, { blip: blipReducer });
+  });
+
+  afterEach(() => {
+    server.resetHandlers();
+  });
+
+  afterAll(() => server.close());
+
+  it('fetches and renders each category of patients', async () => {
```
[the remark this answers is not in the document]
henry-tp: Ignore this practice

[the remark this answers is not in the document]
henry-tp: Ignore this practice

app/pages/clinicworkspace/TideDashboardV2/useTableColumns.js
```diff
@@ -0,0 +1,40 @@
...
+
+  return [
+    columnTypes.patientDetails,
+    columnTypes.placeholder,
+    columnTypes.placeholder,
+    columnTypes.placeholder,
+    columnTypes.placeholder,
+    columnTypes.placeholder,
+    columnTypes.placeholder,
+    columnTypes.placeholder,
```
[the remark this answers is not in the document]
henry-tp: this goes away in the next branch

app/pages/clinicworkspace/TideDashboardV2/TideDashboardV2.js
```diff
@@ -0,0 +1,57 @@
...
+  const { data } = useTideDashboardPatients();
+
+  // Sync category to data fetching resolution; prevents visual glitch due to
+  // category updating view before the API call resolves and updates it again
+  const resolvedCategory = data?.category || category;
+
+  const tableColumns = useTableColumns(resolvedCategory);
+  const emptyContentNode = useMemo(() => <EmptyContentNode />, []);
+
+  if (!data) return null;
```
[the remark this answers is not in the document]
henry-tp: There is potential UX detriment to having loading states, due to very visible (and jarring) pop-in. I'm going to leave this up to see if you think there is a problem Chris, but I have learned over time that loading spinners for relatively fast point-to-point API calls are potentially detrimental, and this one is pretty fast.

app/pages/clinicworkspace/TideDashboardV2/Cells.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { Box } from 'theme-ui';
```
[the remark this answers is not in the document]
henry-tp: Always ignore import order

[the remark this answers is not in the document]
henry-tp: I'll add the PropTypes at a later branch where I create all the cells, as this is currently just a stub.

app/pages/clinicworkspace/clinicworkspace.js
```diff
@@ -14,57 +13,106 @@ import { useFlags } from 'launchdarkly-react-client-sdk';
 import TabGroup from '../../components/elements/TabGroup';
 import ClinicWorkspaceHeader from '../../components/clinic/ClinicWorkspaceHeader';
 import ClinicPatients from './ClinicPatients';
+import TideDashboardV2 from './TideDashboardV2';
 import Prescriptions from '../prescription/Prescriptions';
 import { PatientInvites } from '../share';
 import * as actions from '../../redux/actions';
-import config from '../../config';
+import { resetTideDashboardState } from './TideDashboardV2/tideDashboardSlice';
+
```
henry-tp, line 21: Due to the complexity of how the TIDE flag loads asynchronously and the new TIDE tab being placed between the "core" tabs, I had to re-architect all of how the tabs work. It is now more directly derived state based on the URL, rather than having the state intermediary with the `useEffect`.

I had an interesting challenge with the gating. 

I first tried a hard gating, where ClinicWorkspace doesn't show the optional tabs until LD fetch returns, then we show the tabs based on the flags, then we show the TIDE content. This was okay, but it creates a sluggish experience when someone hard navigates to `/clinic-workspace/tide-dashboard` because there is a waterfall loading .. at first, it looks like you've landed on the `/patients` tab, then after a second or two you get thrown to `/tide-dashboard`, etc. This is a compounding problem with our already-kinda-sluggish-at-times platform.

NOT IMPLEMENTED: This is what I tried to avoid for authorized user doing a hard navigation:
https://github.com/user-attachments/assets/6f80e5d0-9100-4672-ab4b-06f7910c7168

The strategy I chose was "Optimistic Authorization" of page access - to assume that if you land on `/tide-dashboard`, you _probably_ got there legitimately and it should be an option. I then leave the responsibility with the child to gate access to the details within the page or redirect if unauthorized. So page access authorization is optimistic, but authorization to render within the page is not optimistic. Then, we wait for LD to finish fetching, so we know what other tabs to show (e.g. prescriptions).

This is the result:

Hard Navigation, Authorized User
https://github.com/user-attachments/assets/e9912825-9d64-449f-9f77-6cdfbe91b429

Hard Navigation, Unauthorized User
https://github.com/user-attachments/assets/400d7da0-8db3-4148-ae8b-14c41ff43e60

Navigation to an unknown path
https://github.com/user-attachments/assets/f757d765-06c3-4cd1-9d77-e86ff60437a8

app/pages/clinicworkspace/TideDashboardV2/TideDashboardV2.js
```diff
@@ -0,0 +1,64 @@
...
+  const resolvedCategory = data?.category || category;
+
+  const tableColumns = useTableColumns(resolvedCategory);
+  const emptyContentNode = useMemo(() => <EmptyContentNode />, []);
+
+  if (isUnauthorized) {
+    return <Redirect to={{ pathname: '/clinic-workspace/patients', search }} />;
+  }
+
+  if (!isAuthorized || !data) return null;
```
henry-tp, line 37: Here, I still bar access at the data view layer until we can verify the user is authorized (but I still let the data load, just not show). I didn't want to have it show by default.

Importantly, the table data can still load while LD is fetching in parallel, so there is no waterfall effect.

app/bootstrap.js
```diff
@@ -107,8 +107,6 @@ const LAUNCHDARKLY_FLAG_DEFAULTS = {
   'showSummaryDashboard': false,
   'showSummaryDashboardLastReviewed': false,
   'showTideDashboard': false,
-  'showTideDashboardLastReviewed': false,
-  'showTideDashboardPatientDrawer': false,
```
henry-tp, line 111: not relevant anymore, everyone who has access to TIDE Dashboard has access to both of these

PR #2008, opened 2026-08-18
WEB-4460 - Patient Last Reviewed

henry-tp:
WEB-4460

We are intentionally getting rid of some flags now, so all clinics that have access to TIDE will have access to Patient Last Reviewed. I tried to create a sensible abstraction for the toggle.


__tests__/unit/app/components/clinic/PatientLastReviewed.test.js
```diff
@@ -0,0 +1,142 @@
+/* global jest, beforeEach, describe, it, expect */
```
henry-tp, line 1: So there are a few things happening here. PatientLastReviewed uses a similar Adapter pattern to the filters.

Firstly, I migrated these tests out of the older `ClinicPatients.test.js` file to translate them to RTL best practices. 

Next, while we have typically held all of the tests at the top level (ClinicPatients.js) and that _is_ better for full "integration" testing, I feel like it's getting so large as to be difficult to manage and parse. Now that I've chunked up the component into 3 different layers, I further divided up tests and allocated them according to what I felt is each component's responsibility. 

These are:

- ClinicPatients.js (Container Level) - "Does it show the column and the button under the correct conditions?"

- PatientLastReviewed.js (Adapter Level) - "Does it fire the correct API calls, handle the responses, and set the state  correctly?"

- ReviewPatientToggle.js (Element Level) - "Does it show the right data and visual elements in different patient states?"

Hopefully as you read through the tests, these will seem sensible and kind of self-evident for where the tests "belong"

app/pages/clinicworkspace/components/ReviewPatientToggle/PatientLastReviewedGenericAdapter.js
```diff
@@ -0,0 +1,65 @@
...
+}) => {
+  const { set: setToast } = useToasts();
+  const selectedClinicId = useSelector((state) => state.blip.selectedClinicId);
+  const patientId = patient?.id;
+
+  const [reviews, setReviews] = useState(patient?.reviews || []);
+
+  useEffect(() => { // Sync up if prop updates
+    setReviews(patient?.reviews || []);
+  }, [patientId, patient?.reviews]);
```
henry-tp, line 32: Hate this, but the UX is much smoother with it on.

If I remove it, what happens is:

```
POST -> 200 -> GET -> 200 -> Update UI
         |                      |
         |                      -> We don't show it until here
         |
         -> We know it's good here, but keep stale UI
```

So there is noticable lag due to waiting for 2 HTTP calls to happen in series.

With the useEffect, what happens is

```
POST -> 200 -> Update UI -> GET -> 200 -> Update UI (from props)
                  |                          |
                  |                          -> Should be a no-op
                  |
                  -> Show it as soon as it's good
```

The secondary sync GET call is still valuable to sync with BE state, IMO

app/pages/clinicworkspace/components/ReviewPatientToggle/ReviewPatientToggle.js
```diff
@@ -0,0 +1,120 @@
...
+  onUndo = noop,
+  processing = false,
+  recentlyReviewedThresholdDate = moment().startOf('isoWeek').toISOString(),
+}) => {
+  const { t } = useTranslation();
+  const pageName = useClinicMetricsPageName();
+  const selectedClinicId = useSelector((state) => state.blip.selectedClinicId);
+  const loggedInUserId = useSelector((state) => state.blip.loggedInUserId);
+  const timePrefs = useSelector((state) => state.blip.timePrefs);
+
```
henry-tp, line 31: This logic is mostly just 1:1 copypasta from the existing PatientLastReviewed component. I just wanted to genericize it for use in other places (like TIDE) where we don't have access to the pre-assembled patient list, for example

app/pages/clinicworkspace/TideDashboardV2/PatientLastReviewed.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import PatientLastReviewedGenericAdapter from '../components/ReviewPatientToggle/PatientLastReviewedGenericAdapter';
+
+const PatientLastReviewed = (props) => {
+  return <PatientLastReviewedGenericAdapter {...props} />;
+};
```
henry-tp: Let me know if you hate this kind of indirection. For me, there is value in keeping things organized like this, but you may not feel the same. For me, it's kind of explicitly stating "Some components implement their own adapter, this one uses the Generic one as-is with no extension".

__tests__/unit/pages/dashboard/PatientDrawer/MenuBar/MenuBar.test.js
```diff
@@ -163,20 +156,13 @@ describe('MenuBar Component', () => {
...
 
   describe('Last Reviewed Component', () => {
-    it('should display last reviewed section when feature flag is enabled', () => {
-      renderMenuBar({}, defaultState, { showTideDashboardLastReviewed: true });
+    it('should display last reviewed section', () => {
+      renderMenuBar({}, defaultState);
 
       expect(screen.getByTestId('last-reviewed-section')).toBeInTheDocument();
-      expect(screen.getByTestId('patient-last-reviewed')).toBeInTheDocument();
+      expect(screen.getByTestId('patient-review-toggle')).toBeInTheDocument();
```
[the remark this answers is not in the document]
henry-tp: No, I am specifically asserting the grandchild component be rendered

PR #2009, opened 2026-08-18
WEB-4460 - More Menu and Modals

henry-tp:
WEB-4460

As part of this branch I needed to abstract the EditPatientDialog into it's own generic component so it's a little messy. My apologies


app/pages/clinicworkspace/TideDashboardV2/TideDashboardV2.js
```diff
@@ -19,11 +19,14 @@ import usePruneInvalidFilters from './usePruneInvalidFilters';
...
 
+import EditPatientDialogController from './modals/EditPatientDialogController';
+import DataConnectionsModalController from './modals/DataConnectionsModalController';
+
 const Gap = () => <Box sx={{ marginLeft: 'auto' }}></Box>;
 
 const tableContainerProps = { sx: { containerType: 'inline-size' } };
 
-const TideDashboardV2 = () => {
+const TideDashboardV2 = ({ api }) => {
```
henry-tp, line 31: 😠

app/components/clinic/EditPatientDialog.js
```diff
@@ -1,119 +1,88 @@
-import React, { useState, useEffect, useMemo, useRef } from 'react';
-import { useSelector, useDispatch } from 'react-redux';
+import React, { useState, useEffect, useMemo } from 'react';
+import { useSelector } from 'react-redux';
```
henry-tp, line 2: Before there was only one `<EditPatientDialog />` within `navPatientHeader`.

When you look at the past files, it has some funky page-specific concerns like "should clear data if there is a chart data" which is kind of unrelated to the EditPatientDialog. 

I've moved it here into the generic clinic namespace and made the `navPatientHeader` "version" of the modal to be an abstraction over top of the generic modal

By abstracting it into its own component we can ensure it's only concerned with what it should be concerned about, and leave those page-specific concerns to components that are dedicated to the page.

I could have just made another duplicate, rather than tying everything together here, but it just seemed like the right thing to do here. There is one more "version" of the modal in ClinicPatients.js but touching that could get really messy, so I was afraid to dive into that in this ticket.

app/components/navpatientheader/EditPatientDialogController.js
```diff
@@ -0,0 +1,84 @@
...
+import { useSelector, useDispatch } from 'react-redux';
+import { useTranslation } from 'react-i18next';
+import noop from 'lodash/noop';
+import get from 'lodash/get';
+import isEqual from 'lodash/isEqual';
+import { DEFAULT_GLYCEMIC_RANGES } from '../../core/glycemicRangesUtils';
+import { useToasts } from '../../providers/ToastProvider';
+import * as actions from '../../redux/actions';
+import EditPatientDialog from '../clinic/EditPatientDialog';
+
```
henry-tp, line 11: This is a new file but it's basically copypasta'd from the existing `EditPatientDialog` and then had all of the dialog-specific concerns ripped out. Like if you copy the file, compare the diff with the old `EditPatientDialog`, you'll see it's the same with a lot of lines missing

krystophv:
@coderabbitai review

app/pages/clinicworkspace/TideDashboardV2/TideDashboardV2.js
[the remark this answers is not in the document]
henry-tp: Test doesn't need it, and calling `api` is outside of the scope of responsibilities of the component.

In the future, always evaluate whether use of a function or utility is part of the scope of responsibilities before asserting that a more intricate fixture should be used

app/components/navpatientheader/EditPatientDialogController.js
```diff
@@ -0,0 +1,84 @@
...
+      dispatch(actions.worker.dataWorkerRemoveDataRequest(null, currentPatientInViewId));
+      shouldClearDataRef.current = false;
+    }
+  };
+
+  const handleEditFailure = () => {
+    setToast({
+      message: get(notification, 'message'),
+      variant: 'danger',
+    });
```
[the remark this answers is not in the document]
henry-tp: I'll leave this comment up, I don't really want to touch this because I basically just abstracted the logic out (which was preferable to duplicating the JSX into another file). So this logic is all from the existing component.

app/redux/reducers/misc.js
```diff
@@ -763,6 +763,11 @@ export const clinics = (state = initialState.clinics, action) => {
       const patient = _.get(action.payload, 'patient');
       const patientId = _.get(action.payload, 'patientId');
       const clinicId = _.get(action.payload, 'clinicId');
+
+      if (!state[clinicId].patients) {
+        return state;
+      }
```
henry-tp, line 769: Edge case; without this fix, a crash occurs  when the `patients` map in redux state has not been populated. Tidepool tries to set a nested field `patients.[patientId]` but `patients` is undefined.

This only happens if the user hard navigates to `/clinic-workspace/tide-dashboard`, not if they load Tidepool and then click the TIDE Dashboard tab to get there.

PR #2016, opened 2026-08-21
[WEB-4710] Ensure 2FA banner is shown correctly when switching between clinics

clintonium-119:
[WEB-4710]
• Replace `data` with `currentData` from RTK Query hook to
  prevent using cached results from previous clinic queries
  when switching between clinics.

• RTK Query's `data` field persists values from previous
  requests on the hook; `currentData` only contains data
  from the active query, ensuring correct behavior when
  skip conditions change.

• Update mock return values in tests to include both `data`
  and `currentData` fields for RTK Query hook parity.

• Add test case for skipped queries: verifies that stale
  roster data is ignored when the query is skipped for the
  selected clinic.

• Add test case for in-flight queries: verifies that stale
  roster data is ignored while the selected clinic's roster
  is being fetched, preserving the 2FA exemption until the
  current request resolves.

[WEB-4710]: https://tidepool.atlassian.net/browse/WEB-4710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


henry-tp (approved):
I've fallen into this trap a bit before

PR #2027, opened 2026-09-03
[WEB-4722] Allow users to share with clinics using legacy share codes from merged clinics

clintonium-119:
[WEB-4722](https://tidepool.atlassian.net/browse/WEB-4722)
Allow users to share with clinics using legacy share codes from merged clinics

We don't get the full list of supported share codes returned in clinic fetches, so I added a new RTK endpoint that will track clinic responses by requested share codes, and then continue to use the canonical share code for the actual invite action. 

[WEB-4722]: https://tidepool.atlassian.net/browse/WEB-4722?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


__tests__/utils/setupStore.js
```diff
@@ -1,9 +1,10 @@
 import { configureStore } from '@reduxjs/toolkit';
 import { RTKQueryApi } from '@app/redux/api/baseApi';
 
-export const setupStore = (preloadedState = {}) => {
+export const setupStore = (preloadedState = {}, extraReducers = {}) => {
   return configureStore({
     reducer: {
+      ...extraReducers,
```
henry-tp, line 7: Well, well, well: https://github.com/tidepool-org/blip/pull/2005/changes#diff-3813d3c1b565b2c49e0823914f7f893fc9af0e992146651d07983cbce8db2964R4-R8

=P

app/pages/share/ShareInvite.js
```diff
@@ -221,7 +226,8 @@ const ShareInvite = (props) => {
         } else {
           trackMetric('fetched clinic details with share code');
 
-          dispatch(actions.async.fetchClinicByShareCode(api, values.shareCode));
+          setSubmittedShareCode(values.shareCode);
+          setSubmitting(false);
```
henry-tp, line 230: Would you be open to using the `Lazy` version of the query for this? I prefer it because it reads Click -> Request -> Response -> Handler. This removes two instances where you need to "look around" for what a listener is doing, including the `useEffect` for the error. I know I have become a bit of a `useEffect` cop so I ask that you forgive me for always harping on it =P

So for me that looks like:

```jsx
const [fetchClinic, {
    currentData: clinic,
    isFetching: fetchingClinic,
    reset: resetClinicByShareCodeQuery,
  }] = useLazyGetClinicByShareCodeQuery();
```

```jsx
onSubmit: async (values, { setSubmitting }) => {
  // ...
  } else {
    trackMetric('fetched clinic details with share code');

    fetchClinic(values.shareCode)
    .unwrap()
    .catch(() => {
      setToast();
    })
    .finally(() => {
      setSubmitting(false)
    });
  }
}
```

```jsx
  const handleBack = () => {
    // ...
    resetClinicByShareCodeQuery();
  }
```

henry-tp (changes requested):

henry-tp:
I'll just ask you to double-check the release version for this before merge; this branch is pointed at `release-1.100.1` but the Fix Label in JIRA is for `1.101.2`

app/pages/share/ShareInvite.js
clintonium-119: I like this.  That's nicer.  Will do.

app/pages/share/ShareInvite.js
```diff
@@ -226,8 +224,15 @@ const ShareInvite = (props) => {
         } else {
           trackMetric('fetched clinic details with share code');
 
-          setSubmittedShareCode(values.shareCode);
-          setSubmitting(false);
+          // Returned so Formik holds isSubmitting until the lookup settles.
```
henry-tp, line 227: I learned something new today

henry-tp (approved):

PR #2029, opened 2026-09-04
WEB-4676 - Rename 'Bring Data into Tidepool' and remove link

henry-tp:
WEB-4676

### DIFF
--- a/app/components/elements/Button.js
+++ b/app/components/elements/Button.js
@@ -60,7 +60,7 @@ export function Button(props) {
   };
 
   let justifyContent = 'center';
-  if (icon) justifyContent = isLeftIcon ? 'flex-end' : 'flex-start';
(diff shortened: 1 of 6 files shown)


app/components/elements/Button.js
```diff
@@ -60,7 +60,7 @@ export function Button(props) {
   };
 
   let justifyContent = 'center';
-  if (icon) justifyContent = isLeftIcon ? 'flex-end' : 'flex-start';
+  if (icon || iconSrc) justifyContent = isLeftIcon ? 'flex-end' : 'flex-start';
```
henry-tp, line 63: just a little existing bug that's affecting the other button, it must also acknowledge isLeftIcon when `iconSrc` is used instead of `icon`

clintonium-119 (approved):
LGTM :)

PR #2047, opened 2026-09-17
[WEB-4720] Enable Zendesk Messaging

krystophv:
to address [WEB-4720]

The Zendesk AI agent uses Zendesk Messaging, while our `connect-src` allowlist
only covered Web Widget Classic. The widget loaded its static assets, then
stopped at "Offline. You won't receive messages."

Messaging connects to per-pod Sunshine Conversations hosts such as
`pod-20-sunco-ws.zendesk.com` and `*.smooch.io`. Neither was allowed.

We set `connect-src` explicitly, so it does not use `default-src` as a fallback.
The messaging hosts must appear in `connect-src` itself.

This change uses Zendesk's domain allowlist. 

Classic remains in use during the migration. Its Zopim websocket and image
hosts remain allowed. 

### `server.js`

- `connect-src` now allows Zendesk, Smooch, Sentry, and Twilio over the protocols
  each service uses. The Zendesk wildcard replaces two account-specific hosts.
- `img-src` now allows Zendesk message attachments and avatars.
- `media-src` now allows Zendesk voice audio. It previously fell back to
  `default-src 'none'`, which blocked all audio.
- `script-src` no longer contains two WSS sources. Scripts cannot load over WSS.

### `app/pages/smartonfhir/smartonfhir.js`

The SMART-on-FHIR page must hide the support widget inside the EHR embed. It
previously called only the Classic API, which does not hide Messaging.

The page now calls both hide APIs. Each call catches the error raised when that
widget implementation is unavailable.

The unit tests cover both configurations. One test makes the Classic API fail;
the other makes the Messaging API fail.

[WEB-4720]: https://tidepool.atlassian.net/browse/WEB-4720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


clintonium-119 (approved):
Approved :)

